PR #2, opened 2022-01-04
MXDGA-3724: Added the MVP for the fast grid scan motion

DominicOram:
See https://jira.diamond.ac.uk/browse/MXDGA-3724

To review:
* Confirm that the Fast Grid Scan device is doing everything required for https://jira.diamond.ac.uk/browse/MXGDA-3733
* Confirm unit tests run
* Note that the system tests will not work against s03 currently as the fast grid scan IOC is not running on it

### DIFF
--- a/src/artemis/devices/fast_grid_scan.py
+++ b/src/artemis/devices/fast_grid_scan.py
@@ -1,0 +1,141 @@
+import threading
+import time
+from typing import List
+from ophyd import Component, Device, EpicsSignal, EpicsSignalRO, EpicsSignalWithRBV
+from ophyd.status import DeviceStatus, StatusBase, SubscriptionStatus
+
+
+class GridScanCompleteStatus(DeviceStatus):
+    """
+    A Status for the grid scan completion
+    A special status object that notifies watches (progress bars)
+    based on comparing device.expected_images to device.position_counter.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.start_ts = time.time()
+
+        # Notify watchers (things like progress bars) of new values
+        self.device.position_counter.subscribe(self._notify_watchers)
+        self.device.status.subscribe(self._running_changed)
+        # some state needed only by self._notify_watchers
+        self._name = self.device.name
+        self._target_count = self.device.expected_images
+
+    def _notify_watchers(self, value, *args, **kwargs):
+        if not self._watchers:
+            return
+        time_elapsed = time.time() - self.start_ts
+        try:
+            fraction = value / self._target_count
+        except ZeroDivisionError:
+            fraction = 1
+        except Exception:
+            fraction = None
+            time_remaining = None
+        else:
+            time_remaining = time_elapsed / fraction
+        for watcher in self._watchers:
+            watcher(
+                name=self._name,
+                current=value,
+                initial=0,
+                target=self._target_count,
+                unit="images",
+                precision=0,
+                fraction=fraction,
+                time_elapsed=time_elapsed,
+                time_remaining=time_remaining,
+            )
+
+    def _running_changed(self, value=None, old_value=None, **kwargs):
+        if (old_value == 1) and (value == 0):
+            # Stopped running
+            number_of_images = self.device.position_counter.get()
+            if number_of_images != self._target_count:
+                self.set_exception(
+                    Exception(
+                        f"Grid scan finished without collecting expected number of images. Expected {self._target_count} got {number_of_images}."
+                    )
+                )
+            else:
+                self.set_finished()
+            self.clean_up()
+
+    def clean_up(self):
+        self.device.position_counter.clear_sub(self._notify_watchers)
+        self.device.status.clear_sub(self._running_changed)
+
+
+class FastGridScan(Device):
+
+    x_steps: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "X_NUM_STEPS")
+    y_steps: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Y_NUM_STEPS")
+    z_steps: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Z_NUM_STEPS")
+
+    x_step_size: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "X_STEP_SIZE")
+    y_step_size: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Y_STEP_SIZE")
+    z_step_size: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Z_STEP_SIZE")
+
+    dwell_time: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "DWELL_TIME")
+
+    x_start: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "X_START")
+    y1_start: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Y_START")
+    y2_start: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Y2_START")
+    z1_start: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Z_START")
+    z2_start: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Z2_START")
+
+    position_counter: EpicsSignal = Component(
+        EpicsSignal, "POS_COUNTER", write_pv="POS_COUNTER_WRITE"
+    )
+    x_counter: EpicsSignalRO = Component(EpicsSignalRO, "X_COUNTER")
+    y_counter: EpicsSignalRO = Component(EpicsSignalRO, "Y_COUNTER")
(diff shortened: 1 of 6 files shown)


src/artemis/devices/unit_tests/test_gridscan.py
```diff
@@ -0,0 +1,134 @@
+from ophyd.sim import make_fake_device
+from src.artemis.devices.fast_grid_scan import FastGridScan, time
+
+from mockito import *
```
graeme-winter, line 4: Personally I prefer to import what you use rather than `from thing import *` - so for example when I wonder what `when()` is below, I can see where it comes from

src/artemis/devices/fast_grid_scan.py
```diff
@@ -0,0 +1,144 @@
...
+    scan_invalid: EpicsSignalRO = Component(EpicsSignalRO, "SCAN_INVALID")
+
+    run_cmd: EpicsSignal = Component(EpicsSignal, "RUN.PROC")
+    stop_cmd: EpicsSignal = Component(EpicsSignal, "STOP.PROC")
+    status: EpicsSignalRO = Component(EpicsSignalRO, "SCAN_STATUS")
+
+    # Kickoff timeout in seconds
+    KICKOFF_TIMEOUT: float = 5.0
+
+    def set_program_data(self, nx, ny, width, height, exptime, startx, starty, startz):
```
graeme-winter, line 103: Kinda feel like more structured API here would help? `set_program_data` call in the test just looks like pushing a bunch of random numbers

graeme-winter (commented):
Made a couple of comments, but these are just opinions on scanning the code. 

Do we use `black` for formatting?

I don't know what the "house style" is for blue sky so I don't know if `set_params(1, 2, 3, 4, 5)` is good practice or no...

src/artemis/devices/unit_tests/test_gridscan.py
DominicOram: Agreed, I wouldn't use `import *` in the actual production code so I shouldn't in the test code.

DominicOram:
> Made a couple of comments, but these are just opinions on scanning the code.
> 
> Do we use `black` for formatting?
> 
> I don't know what the "house style" is for blue sky so I don't know if `set_params(1, 2, 3, 4, 5)` is good practice or no...

I've been using `black`, have a mild preference for it as it avoids arguments over specific formatting settings. Not sure if ophyd/bluesky use it. Looks like they use flake8 (https://github.com/bluesky/ophyd/blob/master/.flake8) but they're not necessarily mutually exclusive. Now that this repo is starting to grow I'll add a ticket in for some CI, including format checking against a similar flake8 profile to them.

src/artemis/devices/fast_grid_scan.py
DominicOram: Yh, I left this in as a bit of a placeholder until I had a better feel for how we would actually be using the Device upstream but I'll think about improving it now.

DominicOram: I borrowed some of the work of @callumforrester for this, which also included some checking against motor limits.

graeme-winter:
> > Made a couple of comments, but these are just opinions on scanning the code.
> > Do we use `black` for formatting?
> > I don't know what the "house style" is for blue sky so I don't know if `set_params(1, 2, 3, 4, 5)` is good practice or no...
> 
> I've been using `black`, have a mild preference for it as it avoids arguments over specific formatting settings. Not sure if ophyd/bluesky use it. Looks like they use flake8 (https://github.com/bluesky/ophyd/blob/master/.flake8) but they're not necessarily mutually exclusive. Now that this repo is starting to grow I'll add a ticket in for some CI, including format checking against a similar flake8 profile to them.

We already have `black` and `flake8` set up as pre-commit hooks in most of our analysis code, so we can probably just copypasta the configuration over. Prefer pre-commit to working it out in a CI system as keeps the history cleaner... and makes for easier to read diffs.

src/artemis/devices/fast_grid_scan.py
```diff
@@ -0,0 +1,218 @@
...
+
+        self._name = self.device.name
+        self._target_count = self.device.expected_images.get()
+
+    def _notify_watchers(self, value, *args, **kwargs):
+        if not self._watchers:
+            return
+        time_elapsed = time.time() - self.start_ts
+        try:
+            fraction = value / self._target_count
```
callumforrester, line 87: It turns out there have been some issues with the `fraction` parameter in Ophyd. For some classes it goes from 0 -> 1, for other classes it goes the other way around. The correct behaviour is now documented here: https://github.com/bluesky/bluesky/pull/1496/files. I think this code needs to be reversed, unless I've read it wrong.

src/artemis/devices/fast_grid_scan.py
```diff
@@ -0,0 +1,218 @@
...
+        except ZeroDivisionError:
+            fraction = 1
+            time_remaining = 0
+        except Exception:
```
callumforrester, line 91: If there is some unknown exception I think we should pass it upwards with `self.set_exception()`.

src/artemis/devices/system_tests/test_gridscan_system.py
```diff
@@ -0,0 +1,16 @@
...
+
+@pytest.fixture()
+def fast_grid_scan():
+    fast_grid_scan = FastGridScan(name="fast_grid_scan", prefix="BL03S-MO-SGON-01:FGS:")
+    yield fast_grid_scan
+
+
+@pytest.mark.s03
+def test_set_program_data_and_kickoff(fast_grid_scan: FastGridScan):
+    fast_grid_scan.set_program_data(2, 2, 0.1, 0.1, 1, 0, 0, 0)
```
callumforrester, line 14: I'm probably missing something obvious but I can't find this method, has it been renamed?

callumforrester (changes requested):
Looks great, Dom. Code is very clear and well-documented. I particularly like the studious use of type annotations! Have added a few minor comments for things I spotted here and there.

callumforrester:
@graeme-winter and @DominicOram 

> Prefer pre-commit to working it out in a CI system as keeps the history cleaner... and makes for easier to read diffs.

I prefer both because unknown agents can get around pre-commit, but pre-commit covers the 90% case so probably okay. However I suggest making a CI job for the tests regardless.

src/artemis/devices/fast_grid_scan.py
callumforrester: Didn't spot this chain, I think this needs updating to reflect this thread: https://github.com/DiamondLightSource/python-artemis/pull/2#discussion_r780119290

thomascobb:
I'm going to put a plug in for how we do things in Controls, feel free to adopt whatever is useful:
- Pipenv for version management
- Pre-commit with black, flake8, isort and mypy for static analysis
- Pytest for code and coverage
- Sphinx for tutorials, how-to guides, explanations and reference documentation
- GitHub Actions for code and docs CI and deployment to PyPI and GitHub Pages
- VSCode settings using black, flake8, isort and mypy on save

None of this depends on internal DLS infrastructure so will still work if you unset the PYPI mirror.

We've captured this setup in a skeleton module that you can either copy/paste from or git pull changes from:
https://github.com/dls-controls/dls-python3-skeleton

Here's an example of a module that uses it:
https://github.com/dls-controls/sphinx_rtd_theme_github_versions

src/artemis/devices/motors.py
```diff
@@ -0,0 +1,52 @@
...
+from ophyd.epics_motor import MotorBundle
+
+
+class GridScanLimit:
+    """
+    Represents motor limit(s)
+    """
+
+    def __init__(self, motor: EpicsMotor):
+        motor.wait_for_connection()
```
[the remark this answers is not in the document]
DominicOram: I agree that where I had it before is incorrect.

I've changed it to the responsibility of waiting lying with where the `MotorBundle` is created (e.g. see the tests). What do you think?

src/artemis/devices/fast_grid_scan.py
DominicOram: I used `ADTriggerStatus` (https://github.com/bluesky/ophyd/blob/master/ophyd/areadetector/trigger_mixins.py) as the basis for this, which I guess is also incorrect, I'll add a PR into `ophyd` to fix it.

DominicOram:
Done the rework from both @callumforrester and @graeme-winter. CI issues will be addressed in https://jira.diamond.ac.uk/browse/MXGDA-3761

src/artemis/devices/fast_grid_scan.py
DominicOram: https://github.com/bluesky/ophyd/pull/1025 if you're interested

src/artemis/devices/motors.py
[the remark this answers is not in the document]
callumforrester: Looks good to me

DominicOram:
After discussing with @graeme-winter today I think this is just waiting on review from @bentom08, if you have any thoughts?

bentom08 (approved):
Everything looks fine to me, happy for this to be merged in as is.

PR #1, opened 2021-12-15
MXGDA-3723: Added the MVP implementation of a Zebra

DominicOram:
See https://jira.diamond.ac.uk/browse/MXGDA-3723

To review:
* Confirm that the stage/unstage of the Zebra is doing everything required for https://jira.diamond.ac.uk/browse/MXGDA-3733 (Not including fluorescence https://jira.diamond.ac.uk/browse/MXGDA-3759)
* Confirm unit tests run
* If on the controls network confirm system tests run against s03

PR #3, opened 2022-01-06
MXGDA-3722 Ophyd implementation of an Eiger detector

bentom08:


DominicOram (changes requested):
Great work! I've gone through the whole process in GDA and compared it to what you're doing and I think other than the minor things I've pointed out we're all there. Some additional general things:

* I think we need some unit and system tests. Unit test particularly around any logic, (like the interprolation, where you could add the tests first, then refactor to use numpy and so confirm the behaviour is the same). And system tests running against s03 touching as much of the code as possible, this will help us confirm that all the PV names are correct. Let me know if you need help onadding these :)
* Filenames should be snake case (https://www.python.org/dev/peps/pep-0008/#id38)
* In most places in GDA we do a put and wait for CA callback, ophyd doesn't by default. We should probably have a think in the future about whether we want to stick with the wait or not but at this stage seems safer to err on the side of robustness. I added a utility at https://github.com/DiamondLightSource/python-artemis/pull/1/files#diff-bcfb33410fc63e07b5cc48160a519ae435d07cf2ea5232c1b2682e1027783416 that will wait for the callback. Careful though, I think we don't want to do that when we have a RBV as in that case we want to wait on the RBV being correct as I have commented in the code.
* The diagram the Helen wrote includes a step of clearing odin errors and checking it's initialised before doing the staging. You have the check for initialised but I don't see the clearing errors part
* For setting the parameters what we could do as a bit of a placeholder for now is create a `dataclass` for them all then just have a method on the device which takes it. A bit like what I've done with `GridScanParams` at https://github.com/DiamondLightSource/python-artemis/pull/2/files

DominicOram (approved):
Great work, thanks @bentom08!

Issue #13, opened 2022-01-20
Fix FGS system tests


DominicOram:
Following the work done in https://gitlab.diamond.ac.uk/controls/python3/s03_utils/-/merge_requests/4 S03 is now reporting that twice the number of images than expected are being taken.

# Acceptance Criteria
* The code in the FGS system test that forces the expected images to be twice the expected is removed
* System tests still work

DominicOram:
System tests inside gridscan may need updating to use latest versions of classes

abbiemery:
The Zebra and eiger system tests need updating and fixing too. The others pass when run with s03 working.

dperl-dls:
this is described in more detail by other open issues

Issue #11, opened 2022-01-19
Fix Eiger system tests


DominicOram:
Currently the eiger system tests (https://github.com/DiamondLightSource/python-artemis/blob/main/src/artemis/devices/system_tests/test_eiger_system.py) have a number of workarounds to make them pass. Specifically:
1. Odin's meta does not get initialised
1. The `StaleParameters_RBV` PV does not exist on s03 (it does on i03)
1. After unstaging the eiger some filewriters are still in error

This issues should be compared to real tests on i03 and if there is a descrepency between s03 and i03 then s03 should be updated

PR #9, opened 2022-01-17
8: Expand fgs system tests

DominicOram:
Fixes #8 

Note, is branched from https://github.com/DiamondLightSource/python-artemis/pull/7 so merge that PR first.

To test:
* Ensure you have https://gitlab.diamond.ac.uk/controls/python3/s03_utils/-/merge_requests/4 running for the fast grid scan (this is not yet on ws019
* Run the system tests and confirm they all pass

PR #15, opened 2022-01-20
Mxgda 3733: Minimal gridscan plan

DominicOram:
Fixes https://jira.diamond.ac.uk/browse/MXGDA-3733 and https://jira.diamond.ac.uk/browse/MXGDA-3765.

Note, is branched from https://github.com/DiamondLightSource/python-artemis/pull/9 so merge that first.

To test:
* Cherrypick the code in https://github.com/DiamondLightSource/python-artemis/tree/test_minimal_scan (this is required due to https://github.com/DiamondLightSource/python-artemis/issues/13 and https://github.com/DiamondLightSource/python-artemis/issues/11)
* Run this code i.e. `pipenv run python src/artemis/fast_grid_scan_plan.py`
* Confirm that the coe runs without any errors and gives a progress bar of what it's doing
* Confirm logs are produced in `/tmp/bluesky.log` and `/tmp/ophyd.log`


DominicOram:
Once this code is merged delete https://github.com/DiamondLightSource/python-artemis/tree/test_minimal_scan

src/artemis/fast_grid_scan_plan.py
```diff
@@ -0,0 +1,109 @@
...
+    x_start=0.0,
+    y1_start=0.0,
+    z1_start=0.0,
+)
+DETECTOR_PARAMS = DetectorParams(
+    current_energy=100,
+    exposure_time=0.1,
+    acquisition_id="test",
+    directory="/tmp",
+    prefix="file.h5",
```
bentom08, line 59: If, as I assume, these are intended as example parameters the extension should be removed, as it's incorrect and is added on by odin anyway

bentom08 (changes requested):
Looks good, just a minor, slightly picky comment on the parameters

src/artemis/fast_grid_scan_plan.py
DominicOram: Agreed, fixed

PR #21, opened 2022-01-26
MXGDA 3762: Created a dummy flask server for basic comms with GDA

DominicOram:
See https://jira.diamond.ac.uk/browse/MXGDA-3762

To test:
* Make sure to re-install dependencies
* Run the new system tests and confirm they pass
* Run `pipenv run python src/artemis/main.py`
* Using another terminal confirm that you can communicate with the running flask server:
    * `curl -X PUT http://127.0.0.1:5000/fast_grid_scan/start -d '{"name": "Dom"}' -H "Content-Type: application/json"` to start the "scan"
    * `curl -X PUT http://127.0.0.1:5000/fast_grid_scan/stop` to stop the "scan"
    * `curl http://127.0.0.1:5000/fast_grid_scan/status` to get the server status

### DIFF
--- a/src/artemis/main.py
+++ b/src/artemis/main.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 import json
 import bluesky
 from click import Parameter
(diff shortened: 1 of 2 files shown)


src/artemis/main.py
```diff
@@ -0,0 +1,86 @@
...
+        status = Status.FAILED
+        if action == Actions.START.value:
+            parameters = json.loads(request.data)
+            status = Status.SUCCESS if self.runner.start(parameters) else Status.FAILED
+        elif action == Actions.STOP.value:
+            status = Status.SUCCESS if self.runner.stop() else Status.FAILED
+        return jsonify({"status": status.value})
+
+    def get(self, action):
+        status = Status.BUSY if self.runner.bluesky_running else Status.IDLE
```
bentom08, line 68: Should also specify if there is an error

bentom08 (changes requested):
The server works fine for receiving and returning commands/statuses, but could use some improvements to functionality before we start expanding on it

bentom08 (approved):

PR #22, opened 2022-01-26
added more comprehensive error message for node error check

### LINKED ISSUE #12: Eiger: Show error message if a filewriter is failing
Currently if a filewriter is in error Artemis reports "One or more filewriters is in an error state". A more descriptive error message should be reported with:
* Which filewriter has the error
* What the error is i.e. the `error_message` on the node

The errors should be concated together if they're coming from multiple nodes.

bentom08:
Fixes #12
Also added extra unit test for the odin

### DIFF
--- a/src/artemis/devices/eiger_odin.py
+++ b/src/artemis/devices/eiger_odin.py
@@ -79,11 +79,14 @@ class OdinNodesStatus(Device):
             lambda node: node.frames_dropped.get(), "dropped"
         )
 
-    def get_error_state(self) -> bool:
+    def get_error_state(self) -> Tuple[bool, str]:
         is_error = []
+        error_messages = []
         for node_number, node_pv in enumerate(self.nodes):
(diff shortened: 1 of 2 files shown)


DominicOram (commented):
Looks great, I like the extra test. 

I tidied up the formatting using black (I really need to finish https://github.com/DiamondLightSource/python-artemis/pull/5 so this will be automatic) and I added an extra test, just to double check my understanding of what you'd done.

In adding the test I realised that the error message will report the filewriter number as 0 index. As a programmer I like this but it doesn't match the PV naming convention (which is 1 indexed). I'm not sure which is clearer, what do you think?

DominicOram:
Agreed offline to make it 0-indexed and to change the `node_x` variable names to match

DominicOram (approved):
Have made the suggested indexing change, all looks good now.

PR #23, opened 2022-01-26
Fixes #18 add wait for meta filewriter to be ready

### LINKED ISSUE #18: Gridscan reported failed and no _meta file produced
On the last test done at i03 see logs etc. [here](https://github.com/DiamondLightSource/python-artemis/tree/test_on_i03) it was found that no `_meta` file was produced and the gridscan reported failed despite data looking correct.

## Acceptance Criteria
* The logs are examined for what the issue might be
* Fix is put in place if discovered, otherwise it's written up for any follow up tsting

bentom08:
Fixes #18

### DIFF
--- a/src/artemis/devices/eiger.py
+++ b/src/artemis/devices/eiger.py
@@ -172,7 +172,7 @@ class EigerDetector(Device):
         self.wait_for_stale_parameters()
 
         bit_depth = self.bit_depth.get()
-        self.odin.file_writer.data_type.put(bit_depth)
+        self.odin.file_writer.data_type.put(f"UInt{bit_depth}")
 
         odin_status = self.odin.file_writer.capture.set(1)
         odin_status &= await_value(self.odin.meta.ready, 1)
--- a/src/artemis/devices/eiger.py
+++ b/src/artemis/devices/eiger.py
@@ -174,7 +174,7 @@ class EigerDetector(Device):
         bit_depth = self.bit_depth.get()
         self.odin.file_writer.data_type.put(bit_depth)
 
-        odin_status = set(self.odin.file_writer.capture, 1)
+        odin_status = self.odin.file_writer.capture.set(1)
         odin_status &= await_value(self.odin.meta.ready, 1)
         odin_status.wait(10)
 
--- a/src/artemis/devices/eiger.py
+++ b/src/artemis/devices/eiger.py
@@ -174,7 +174,9 @@ class EigerDetector(Device):
         bit_depth = self.bit_depth.get()
         self.odin.file_writer.data_type.put(bit_depth)
 
-        set_and_wait(self.odin.file_writer.capture, 1, timeout=10)
+        odin_status = set(self.odin.file_writer.capture, 1)
+        odin_status &= await_value(self.odin.meta.ready, 1)
+        odin_status.wait(10)
 
         set_and_wait(self.cam.acquire, 1, timeout=10)
 
--- a/src/artemis/devices/eiger_odin.py
+++ b/src/artemis/devices/eiger_odin.py
@@ -21,6 +21,7 @@ class EigerFan(Device):
 class OdinMetaListener(Device):
     file_name: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "FileName")
     initialised: EpicsSignalRO = Component(EpicsSignalRO, "ProcessConnected_RBV")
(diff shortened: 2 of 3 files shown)


src/artemis/devices/eiger.py
```diff
@@ -174,7 +174,9 @@ def arm_detector(self):
         bit_depth = self.bit_depth.get()
         self.odin.file_writer.data_type.put(bit_depth)
 
-        set_and_wait(self.odin.file_writer.capture, 1, timeout=10)
+        odin_status = self.odin.file_writer.capture.set(1)
+        odin_status &= await_value(self.odin.meta.ready, 1)
+        odin_status.wait(10)
```
graeme-winter, line 179: 😱

src/artemis/devices/eiger.py
```diff
@@ -174,7 +174,9 @@ def arm_detector(self):
         bit_depth = self.bit_depth.get()
         self.odin.file_writer.data_type.put(bit_depth)
 
-        set_and_wait(self.odin.file_writer.capture, 1, timeout=10)
+        odin_status = self.odin.file_writer.capture.set(1)
+        odin_status &= await_value(self.odin.meta.ready, 1)
```
GDYendell, line 178: You should wait for OD:Capture_RBV here as well, which I guess `set` does not do.

bentom08: both the `set` method and the `await_value` methods here return status objects (futures), and the `wait` method below then waits for both futures to complete

GDYendell: Ah. Put with callback on Capture returns when the file is closed, not when it has finished opening, so that may not do what you want.

bentom08: my bad, looks like I was looking at the wrong method in ophyd when I wrote that edit. The `set` method calls `set_and_wait` in a separate thread, and then returns a future which completes once the read PV is correct. So using `set` here should work.

DominicOram:
So whilst I'm sure this is required it's currently failing against S03 and getting stuck waiting for the the meta writer to be ready:

```
(python3.7) [ffv81422@ws455 python-artemis]$ camonitor BL03S-EA-EIGER-01:OD:META:Writing_RBV BL03S-EA-EIGER-01:OD:Capture_RBV BL03S-EA-EIGER-01:OD:META:ProcessConnected_RBV -S BL03S-EA-EIGER-01:OD:AcquisitionID
BL03S-EA-EIGER-01:OD:META:Writing_RBV 2022-02-03 13:13:09.757993 Done  
BL03S-EA-EIGER-01:OD:Capture_RBV 2022-02-03 13:16:40.342380 Done  
BL03S-EA-EIGER-01:OD:META:ProcessConnected_RBV 2022-02-03 13:13:09.758119 1  
BL03S-EA-EIGER-01:OD:AcquisitionID 2022-02-03 13:15:37.829977 file.h5  
BL03S-EA-EIGER-01:OD:AcquisitionID 2022-02-03 13:17:08.007233 001  
BL03S-EA-EIGER-01:OD:AcquisitionID 2022-02-03 13:17:08.053712 file.h5  
BL03S-EA-EIGER-01:OD:Capture_RBV 2022-02-03 13:17:09.338266 Capturing  
```
I've spent some time trying to work out why (and fixed an issue with bit depth being set incorrectly in the process) but it's still failing.

GDYendell:
This seems wrong
```
BL03S-EA-EIGER-01:OD:AcquisitionID 2022-02-03 13:15:37.829977 file.h5  
BL03S-EA-EIGER-01:OD:AcquisitionID 2022-02-03 13:17:08.007233 001  
BL03S-EA-EIGER-01:OD:AcquisitionID 2022-02-03 13:17:08.053712 file.h5
```
Firstly, nothing (e.g. file name, acquisition ID) should include `.h5` on the end as this is just one section of the full file name (for the data writers and the meta writers).

Secondly, OD:FileName is synced to the more specific records to provide a simple interface. If you want more fine grained control you should set only the lower level PVs and not OD:FileName. I.e. it shouldn't be set OD:FileName and FP:FileName / OD:AcquisitionID. Assuming you want the acquisition id to be different to the file prefix, you can set OD:Acquisition and then FP:FileName. Setting META:FileName is also redundant if it is not different, but I don't think that should cause any harm.

DominicOram (approved):
With the latest version of s03 this passes the system tests and looks sensible, happy to merge.

PR #24, opened 2022-02-04
fix zebra config to work on i03

### LINKED ISSUE #17: Zebra: Use the correct settings
After initial testing on i03 it was found that the settings used for the zebra are incorrect. They should in fact use the TTL settings from `multizebra.py` and do not require any position control logic at all. See [here](https://github.com/DiamondLightSource/python-artemis/tree/test_on_i03) for what was changed to work on the instrument

## Acceptance Criteria
* For fast grid scans the Zebra should be configured correctly
* Unit/system tests are updated accordingly
* Check if this fixes issue 3 in https://github.com/DiamondLightSource/python-artemis/issues/11

bentom08:
Fixes #17

### DIFF
--- a/src/artemis/devices/system_tests/test_zebra_system.py
+++ b/src/artemis/devices/system_tests/test_zebra_system.py
@@ -1,6 +1,14 @@
 import pytest
 
-from src.artemis.devices.zebra import Zebra
+from src.artemis.devices.zebra import (
+    Zebra,
+    TTL_DETECTOR,
+    TTL_SHUTTER,
+    PC_PULSE,
+    OR1,
+    IN3_TTL,
+    IN4_TTL,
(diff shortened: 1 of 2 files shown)


DominicOram (changes requested):
Looks good, I think we can go a bit further though:
* We should remove `LogicGateConfigurer.setup_fast_grid_scan` and `PositionCompare.setup_fast_grid_scan`, they'll only cause future confusion as to why they're there. Arguably we can probably remove the whole classes but they might be useful in the future.
* Now we're not using the position compare we can remove the arming/disarming in the stage/unstage. This will break the system tests so we'll have to fix them too

DominicOram (approved):
Looks good, thanks

PR #28, opened 2022-02-15
Mxgda 3764: Run plan from flask server

DominicOram:
See https://jira.diamond.ac.uk/browse/MXGDA-3764

To test:
* Update dependencies
* Run modified tests and confirm they pass
* Start up a runner as in README instructions and start/stop some scans
* Confirm runner acts as expected and starts a scan based on the ophyd/bluesky logs in /tmp/

### DIFF
--- a/src/artemis/fast_grid_scan_plan.py
+++ b/src/artemis/fast_grid_scan_plan.py
@@ -4,7 +4,7 @@ import sys
 
 sys.path.append(os.path.join(os.path.dirname(__file__), "..", ".."))
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from src.artemis.devices.eiger import DetectorParams, EigerDetector
 from src.artemis.devices.fast_grid_scan import (
     FastGridScan,
@@ -20,6 +20,7 @@ from src.artemis.devices.zebra import Zebra
 from src.artemis.devices.det_dim_constants import (
     EIGER2_X_16M_SIZE,
     DetectorSizeConstants,
+    constants_from_type,
 )
 import argparse
 from src.artemis.devices.det_dist_to_beam_converter import (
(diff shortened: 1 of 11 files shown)


src/artemis/main.py
```diff
@@ -17,88 +25,131 @@
...
 
     def get(self, action):
-        return self._craft_return_message(self.runner.get_status())
-
-
-def create_app(test_config=None) -> Tuple[Flask, BlueskyRunner]:
-    runner = BlueskyRunner()
+        if action == Actions.SHUTDOWN.value:
+            self.runner.shutdown()
+            shutdown_func = request.environ.get("werkzeug.server.shutdown")
```
bentom08, line 140: Not entirely clear what this is doing and why the shutdown action is in a GET request rather than PUT request like the other actions

src/artemis/main.py
```diff
@@ -17,88 +25,131 @@
...
 
+@dataclass
+class Command:
+    action: Actions
+    parameters: Optional[FullParameters] = None
+
+
+@dataclass_json
 @dataclass
 class StatusAndError:
```
bentom08, line 47: Maybe rename this class to StatusAndMessage or something similar? Putting Error in the class name is a little misleading at a glance, because if I see an object being created with "Error" in it I assume an error/exception is being raised rather than a status being reported.

bentom08 (changes requested):
Mostly fine, the beam converter and shutdown comments can wait until a later ticket if they're not going to be possible before next testing.

src/artemis/fast_grid_scan_plan.py
```diff
@@ -38,46 +41,62 @@
...
+        detector_distance=100.0,
+        omega_start=0.0,
+        omega_increment=0.1,
+        num_images=10,
+    )
+    det_to_distance = DetectorDistanceToBeamXYConverter(
+        os.path.join(
+            os.path.dirname(__file__),
+            "devices",
+            "det_dist_to_beam_XY_converter.txt",
```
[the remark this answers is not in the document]
DominicOram: Thanks for pointing this out, will do in the linked ticket.

bentom08 (approved):

PR #38, opened 2022-02-21
add put for the meta filename back in

### LINKED ISSUE #36: Metafile name is not changing
When starting a new scan `META:FileName` does not get correctly set, meaning that the meta file will be saved over the previous one.

This was probably introduced in https://github.com/DiamondLightSource/python-artemis/pull/26/

bentom08:
Fixes #36


DominicOram (changes requested):
This doesn't work as `meta.file_name` no longer exists. Sorry, I did too thorough a job removing it in the last ticket!

DominicOram (approved):
Great, thank you!

PR #41, opened 2022-03-03
40: Added code to trigger analysis and associated tests

### LINKED ISSUE #40: Interact with data analysis pipeline
Once we have put the data into ISpyB we will need to tell the analysis pipeline to start running on it. Currently this is done by running:
`/dls_sw/apps/mx-scripts/bin/RunAtStartOfCollect-i03.sh` once we are finished we will then need to tell the pipeline we're finished, currently by running `/dls_sw/apps/mx-scripts/bin/RunAtEndOfCollect-i03.sh` (only on success).

# Acceptance Criteria
* The data analysis pipeline is informed at start/end of collection

# Notes 
* As part of this look into adding the logic from these scripts into Artemis itself

DominicOram:
Fixes #40

To test:
* Confirm unit tests pass
* Confirm that start/end functions perform the same job as `RunAtStartOfCollect-i03.sh` and `RunAtEndOfCollect-i03.sh` (both in `dls_sw\apps\mx=scripts\bin`) Note that they both call https://github.com/DiamondLightSource/python-zocalo/blob/main/src/zocalo/cli/go.py but with different args

### DIFF
--- a/src/artemis/zocalo_interaction.py
+++ b/src/artemis/zocalo_interaction.py
@@ -1,0 +1,47 @@
+import zocalo.configuration
+from workflows.transport import lookup, default_transport
+import getpass
+import socket
+
+
+def _send_to_zocalo(parameters: dict):
+    zc = zocalo.configuration.from_file()
+    zc.activate()
+
+    transport = lookup(default_transport)()
+    transport.connect()
+    message = {
+        "recipes": "mimas",
+        "parameters": parameters,
+    }
+    header = {
+        "zocalo.go.user": getpass.getuser(),
+        "zocalo.go.host": socket.gethostname(),
+    }
+    transport.send("processing_recipe", message, headers=header)
+
+
+def run_start(data_collection_id: int):
+    """Tells the data analysis pipeline we have started a grid scan.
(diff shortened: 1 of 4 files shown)


src/artemis/zocalo_interaction.py
```diff
@@ -0,0 +1,47 @@
...
+    transport.connect()
+    message = {
+        "recipes": "mimas",
+        "parameters": parameters,
+    }
+    header = {
+        "zocalo.go.user": getpass.getuser(),
+        "zocalo.go.host": socket.gethostname(),
+    }
+    transport.send("processing_recipe", message, headers=header)
```
bentom08, line 21: should this have transport.disconnect() at the end? (also feels like a disconnect should go in a finally block, although zocalo has it in the try...)

bentom08 (changes requested):
Tests ran fine, seems to replicate what the zocalo script is doing. Few small changes.

DominicOram:
> Tests ran fine, seems to replicate what the zocalo script is doing. Couple small changes.

Agree with comments, fixed both and added tests.

bentom08 (approved):
Looks fine, tests pass for me.

PR #46, opened 2022-03-15
Fix unit test for checking fast grid scan kickoff

DominicOram:
To test:
* Run unit tests and confirm they still pass


DominicOram:
Scratch this, I have a better fix as part of https://github.com/DiamondLightSource/python-artemis/pull/48

PR #51, opened 2022-03-31
50: Fixed passing detector parameters to eiger

DominicOram:
Fixes #50

To test:
* Confirm new test passes and detector parameters now correctly passed to eiger

PR #52, opened 2022-04-01
Mxgda 3734: Create nexus file

DominicOram:
See https://jira.diamond.ac.uk/browse/MXGDA-3734

Note that this includes https://github.com/DiamondLightSource/python-artemis/pull/51 so review that first.

This also requires new dependencies so you will need to reinstall them.

To test:
* Confirm unit tests pass
* Copy some existing `*_00001.h5` and `*_meta.h5` data from e.g. `/dls/i03/data/2022/cm31105-1/xraycentring/TestProteinaseK/protk_1` to a local folder
* Run Artemis and send it parameters with:
```
"detector_params": {"acquisition_id": "test",
                     "current_energy": 100,
                     "detector_distance": 100.0,
                     "directory": path_to_local_data_folder,
                     "exposure_time": 0.1,
                     "num_images": 10,
                     "omega_increment": 0.1,
                     "omega_start": 0.0,
                     "prefix": prefix_of_file_eg_protk_1_1,
                     "use_roi_mode": false},
```
* Confirm a nexus file is written in the local folder
* Compare with the nexus file from the original folder and see that they follow a broadly similar structure


src/artemis/nexus_writing/write_nexus.py
```diff
@@ -0,0 +1,240 @@
...
+                    self.detector["image_size"][1],
+                    self.detector["image_size"][0],
+                ),
+            )
+
+    def __exit__(self, *_):
+        with h5py.File(self.nexus_file, "r+") as nxsfile:
+            nxsfile["entry"].create_dataset(
+                "end_time", data=np.string_(self._get_current_time())
+            )
```
bentom08, line 240: If we do go with an external timestamp for start/end time as per my earlier comment I'm not sure the context manager will be necessary, as its main function at the moment seems to be to make sure the end time is written.

bentom08 (changes requested):
The merge conflicts here also need to be sorted out

src/artemis/nexus_writing/write_nexus.py
DominicOram: We'll still need to actually write the `end_time` into the nexus file, even if the actual timestamp comes from elsewhere

src/artemis/nexus_writing/write_nexus.py
```diff
@@ -0,0 +1,240 @@
...
+        self.scan_spec = create_scan_spec(parameters.grid_scan_params)
+        self.directory, self.filename = (
+            Path(parameters.detector_params.directory),
+            parameters.detector_params.prefix,
+        )
+        self.num_of_images = parameters.detector_params.num_images
+        self.nexus_file = self.directory / f"{self.filename}.nxs"
+
+    def _get_current_time(self):
+        return datetime.utcfromtimestamp(time.time()).strftime(r"%Y-%m-%dT%H:%M:%SZ")
```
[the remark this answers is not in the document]
DominicOram: Agreed but seems better done in a separate ticket, created https://github.com/DiamondLightSource/python-artemis/issues/72

PR #54, opened 2022-04-06
change formatter max line length to 120

bentom08:


.vscode/settings.json
```diff
@@ -4,5 +4,6 @@
     ],
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
-    "python.formatting.provider": "black"
-}
\ No newline at end of file
+    "python.formatting.provider": "black",
+    "python.formatting.blackArgs": ["--line-length", "120"]
```
graeme-winter, line 8: Since when did `black` have arguments? 🤔 

(in related news, I would humbly suggest `black` is correct here with its defaults)

PR #57, opened 2022-04-12
Call zocalo and ispyb

### LINKED ISSUE #64: Actually call zocalo and ispyb
During testing we had to add in the lines to call zocalo and ispyb, these should be in the code in general.

Notes
* Ensure they don't cause test failures
* Note that `update_grid_scan_with_end_time_and_status` also needs to set `parentid` see https://github.com/DiamondLightSource/python-artemis/tree/local_i03_testing_changes

DominicOram:
Fixes #64


DominicOram:
Closing in favour of https://github.com/DiamondLightSource/python-artemis/pull/78, which has a better history

PR #75, opened 2022-04-19
34 grid scan complete fails

### LINKED ISSUE #34: GridScanCompleteStatus fails
During testing it was found that the GridScanCompleteStatus will fail at the end of the scan, despite the scan being successful. This looks to be because the `running` PV can go low before the monitor is posted for the final image to have been created.

## Acceptance Criteria
* The GridScanCompleteStatus does not fail
* There is a test for the above

DominicOram:
Fixes #34

To test:
* Confirm unit tests still pass

### DIFF
--- a/src/artemis/devices/fast_grid_scan.py
+++ b/src/artemis/devices/fast_grid_scan.py
@@ -90,6 +90,7 @@ class GridScanCompleteStatus(DeviceStatus):
             fraction = None
             time_remaining = None
             self.set_exception(e)
+            self.clean_up()
         else:
             time_remaining = time_elapsed / fraction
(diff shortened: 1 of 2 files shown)


bentom08 (approved):
codecov.yml addition should ideally go in a separate change, but I don't think its of paramount importance so I'll leave it up to you whether you want to separate them or not.

DominicOram:
> codecov.yml addition should ideally go in a separate change, but I don't think its of paramount importance so I'll leave it up to you whether you want to separate them or not.

It was added in https://github.com/DiamondLightSource/python-artemis/pull/89, I just merged it on to this branch to test that it fixed the failing check.

PR #85, opened 2022-04-20
83: Account for no sample id

### LINKED ISSUE #83: Account for no sample
As a user I may run a gridscan without any sample in the beam. Currently doing this causes GDA to send a `sample_id` of 0 and a `sample_barcode` of `None`. This fails when entering values into ISPyB with an unknown foreign key.

DominicOram:
Fixes #83 

To test:
* Confirm new unit tests pass


bentom08 (commented):
I'm not sure why GDA uses a sample id of 0 for no sample (and why it doesn't run into the same error when trying to store it). But if we're going to store a null value for the sampleid I'd prefer to pass a null value to artemis in the JSON rather than adding conditionals to our ispyb methods.

DominicOram:
> I'm not sure why GDA uses a sample id of 0 for no sample (and why it doesn't run into the same error when trying to store it). But if we're going to store a null value for the sampleid I'd prefer to pass a null value to artemis in the JSON rather than adding conditionals to our ispyb methods.

Hmm... You're right, I should have thought a bit more about this. I think you're right that a `None`is better, we should also accept no sample_id being specified in the JSON

DominicOram:
See also https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/35058/5..6

bentom08 (changes requested):
Default value of all dictionary fields in the ispyb-api is None, so these if statements are unnecessary (value of the sampleid field will be None with or without them).

The artemis change should be sufficient for this ticket, although we still need to make sure the barcode is also None before sending the JSON.

DominicOram:
I've left the unit tests in as they're still valid, also allowed the interface to not have the sample info specified as I think that's cleaner.

Added the barcode more explicitly at https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/35058/6..7/configurations/mx-config/scripts/gridscan/artemis_grid_scan.py

src/artemis/ispyb/store_in_ispyb.py
```diff
@@ -148,7 +148,6 @@ def _store_data_collection_group_table(self) -> int:
 
         params = self.mx_acquisition.get_data_collection_group_params()
         params["parentid"] = session_id
-        params["experimenttype"] = "mesh"
```
bentom08, line 151: Is there a reason for removing this? (although it has reminded me I need to update the 3d scan ispyb change to include changing this)

bentom08 (changes requested):

src/artemis/ispyb/store_in_ispyb.py
DominicOram: Oops, typo, well spotted

bentom08 (approved):

PR #105, opened 2022-05-23
61: Correctly cast data to get in ispyb

### LINKED ISSUE #61: Incorrect types being passed to ISPyB
During testing on i03 it was found that we were passing incorrect types to ISPyB in some cases:
* When converting json using `dataclasses_json` numbers come back as numpy floats, ISPyB wants them as primitive floats
* `beam_position`, `xtal_snapshots` and `snapshotOffsetXPixel` were being set as lists rather than single items

Note these have been fixed in https://github.com/DiamondLightSource/python-artemis/tree/local_i03_testing_changes but need tidying up

DominicOram:
Fixes #61 

To test:
* Confirm unit tests pass 
* Changes from https://github.com/DiamondLightSource/python-artemis/compare/testing_round_5 have been incorporated


src/artemis/ispyb/store_in_ispyb.py
```diff
@@ -119,7 +119,7 @@ def _store_data_collection_table(self, data_collection_group_id: int) -> int:
         params["xbeam"], params["ybeam"] = beam_position
         params["xtal_snapshot1"], params["xtal_snapshot2"], params["xtal_snapshot3"] = [
             self.ispyb_params.xtal_snapshots
-        ] * 3
+        ]
```
bentom08, line 122: I thought we passed this in as a list of length 3 so we didn't need the square brackets?

bentom08 (changes requested):

bentom08 (approved):

Issue #93, opened 2022-05-05
Nexus file updates should be atomic


bentom08:
https://jira.diamond.ac.uk/browse/MXGDA-3620 this jira ticket outlines a previous issue in GDA where the nexus file was unable to be read for data analysis because the nexus writing wasn't atmoic.

Solution there was to make the update atomic by creating a temporary copy of the nexus file, updating the copy, then overwriting the original file with the copy.

We should update our nexus writing to do something similar so we don't run into the same problems down the line.

PR #106, opened 2022-05-23
102: Added code to handle datasets spread over multiple h5 files

### LINKED ISSUE #102: Handle when two h5 files exist for the same scan
When the Eiger takes over 1000 images it creates a new file for the other images. We need to make sure that Artemis.nexgen are handling this correctly

DominicOram:
Fixes #102 

To test:
* Confirm unit/system tests pass


DominicOram:
This did not work after testing on i03 (https://github.com/DiamondLightSource/python-artemis/issues/111). Need to think of a better way of testing it offline

DominicOram:
I couldn't work out what the issue was from #111. The new test works and covers the real behavior very closely, if we see something go wrong in testing again we should document the issue better.

DominicOram:
@MattPrit and @bentom08, can you review this together?

src/artemis/nexus_writing/tests/test_write_nexus.py
```diff
@@ -88,6 +89,46 @@ def test_given_parameters_when_nexus_writer_used_as_context_manager_then_all_dat
...
+    params = FullParameters()
+    params.detector_params.num_images = 2540
+    nexus_writer = NexusWriter(params)
+    assert len(nexus_writer.get_image_datafiles()) == 3
+    paths = [str(filename) for filename in nexus_writer.get_image_datafiles()]
+    assert paths == [
+        "/tmp/file_name_0_000001.h5",
+        "/tmp/file_name_0_000002.h5",
+        "/tmp/file_name_0_000003.h5",
+    ]
```
bentom08, line 102: nit - this test could use parameterize to test more use cases (1, 2 and 3 files for example)

bentom08 (approved):
looks fine, one minor comment

MattPrit (approved):
Looks good to me. Once merged, I think I can make use of `dummy_nexus_writer` in my tests for #93.

DominicOram:
In updating the test I found a bug on the edge case. @bentom08, can you re-review? @MattPrit, assuming it's approved feel free to merge this whilst I'm away if it helps you move forward

bentom08 (approved):

PR #110, opened 2022-05-24
97: Added function to wait for analysis results

### LINKED ISSUE #97: Get xray centre results from analysis
Now that we are correctly providing 3D grid scan data to analysis we should look at getting the optimum data collection position back.

## Acceptance Criteria
* The position is retrieved in Artemis and logged
* We have a system test where we can test receiving a message from anaylsis
* The code to get the position is removed from GDA

## Note
* GDA does this by watching a results file, given we can communicate through Zocalo that's probably a better way to do it
* It is undecided what co-ordinate system the positions will be returned in, this is fine for now

DominicOram:
Fixes #97 

To test:
* Confirm unit tests pass

### DIFF
--- a/src/artemis/zocalo_interaction.py
+++ b/src/artemis/zocalo_interaction.py
@@ -1,4 +1,5 @@
 import getpass
+import queue
 import socket
 from time import sleep
 
@@ -73,7 +74,7 @@ def wait_for_result(data_collection_id: int, timeout: int = TIMEOUT) -> Point3D:
         Point in grid co-ordinates that is the centre point to move to
     """
     transport = _get_zocalo_connection()
-    result_received = None
+    result_received: queue.Queue = queue.Queue()
 
     def receive_result(
         rw: workflows.recipe.RecipeWrapper, header: dict, message: dict
@@ -83,8 +84,7 @@ def wait_for_result(data_collection_id: int, timeout: int = TIMEOUT) -> Point3D:
         print(f"Recipe step parameters: {recipe_parameters}")
         transport.ack(header)
         if recipe_parameters["dcid"] == str(data_collection_id):
-            nonlocal result_received
-            result_received = Point3D(*message["max_voxel"])
+            result_received.put(Point3D(*message["max_voxel"]))
         else:
             print(
                 f"Warning: results for {recipe_parameters['dcid']} received but expected {data_collection_id}"
@@ -99,10 +99,6 @@ def wait_for_result(data_collection_id: int, timeout: int = TIMEOUT) -> Point3D:
     )
 
     try:
-        for _ in range(timeout):
-            sleep(1)
-            if result_received:
-                return result_received
-        raise TimeoutError(f"Result not received in {timeout} seconds")
+        return result_received.get(timeout=timeout)
     finally:
         transport.disconnect()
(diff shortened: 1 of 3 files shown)


src/artemis/zocalo_interaction.py
```diff
@@ -49,3 +62,47 @@ def run_end(data_collection_id: int):
...
+        recipe_parameters = rw.recipe_step["parameters"]
+        print(f"Recipe step parameters: {recipe_parameters}")
+        if recipe_parameters["dcid"] == str(data_collection_id):
+            transport.ack(header)
+            nonlocal result_recieved
+            result_recieved = Point3D(*message["max_voxel"])
+        else:
+            print(
+                f"Warning: results for {recipe_parameters['dcid']} recieved but expected {data_collection_id}"
+            )
```
rjgildea, line 91: With `acknowledgment=True` you need to explicitly `ack` or `nack` each message otherwise you won't receive any subsequent messages until you have done so:
```suggestion
            )
            transport.ack(header)
```

DominicOram:
Thanks for the review @rjgildea, I have addressed all your comments.

src/artemis/zocalo_interaction.py
```diff
@@ -49,3 +62,47 @@ def run_end(data_collection_id: int):
...
+    )
+
+    try:
+        for _ in range(timeout):
+            sleep(1)
+            if result_received:
+                return result_received
+        raise TimeoutError(f"Result not received in {timeout} seconds")
+    finally:
+        transport.disconnect()
```
rjgildea, line 108: I don't think this use of `nonlocal` to communicate the result back to the main thread will work as the subscription callback will be called in another thread. Something along the lines of the following, using `queue.Queue` to communicate the result works in my test consumer:

```suggestion
    results_queue: queue.Queue = queue.Queue()

    def receive_result(
        rw: workflows.recipe.RecipeWrapper, header: dict, message: dict
    ) -> None:
        print(f"Received {message}")
        recipe_parameters = rw.recipe_step["parameters"]
        print(f"Recipe step parameters: {recipe_parameters}")
        transport.ack(header)
        if recipe_parameters["dcid"] == str(data_collection_id):
            results_queue.put(max_voxel)
        else:
            print(
                f"Warning: results for {recipe_parameters['dcid']} received but expected {data_collection_id}"
            )

    workflows.recipe.wrap_subscribe(
        transport,
        "xrc.i03",
        receive_result,
        acknowledgement=True,
        allow_non_recipe_messages=False,
    )

    max_voxel = None
    wait = 0.1  # seconds
    start_time = time.time()
    while True:
        try:
            max_voxel = Point3D(*results_queue.get(True, wait))
            print(f"{max_voxel=}")
            break
        except queue.Empty:
            if time.time() - start_time >= timeout:
                print(f"Timed out after {timeout} seconds")
                break
            time.sleep(wait)
            continue
    transport.disconnect()
    if max_voxel is None:
        raise TimeoutError(f"Result not received in {timeout} seconds")
    return max_voxel
```

DominicOram: Great, thanks. I've added a `queue, I think we can rely on the `queue` itself to do the timeout, rather than having to do it ourselves?

rjgildea: Yes, I think this should be fine if you're just expecting to receive a single result.

rjgildea (approved):

DominicOram:
After testing on beamline it was decided to wait based on the group id rather than data collection id as that's more stable. I will update this PR.

PR #112, opened 2022-05-26
49: move to position from analysis

### LINKED ISSUE #49: Move to centre as reported from analysis
As it's final task, once we have retrieved the position (https://github.com/DiamondLightSource/python-artemis/issues/97) and converted it to motor positions (https://github.com/DiamondLightSource/python-artemis/issues/98) Artemis should move to the center position as given by the analysis software. 

## Acceptance Criteria
* Do the final movement
* Tell GDA we're done and at center

DominicOram:
Fixes DiamondLightSource/hyperion#49

Note: branched from https://github.com/DiamondLightSource/python-artemis/pull/110 so merge that first


bentom08:
I noticed the grid position -> motor position method we have doesn't account for omega at all, is that already taken care of by epics/analysis or should we be adjusting the motor positions depending on our scan's omega angle? (as an example, we'll get different centring results taking scans at omega=0-90 vs omega=90-180 as the y/z axes will be flipped, but we should find the same x, y, z motor positions in each case)

DominicOram:
I think I do need to think about Ben's comment here. I will need to:
* Ensure the grid scan is done from omega = 0
* Move to omega = 0 before moving the other axes

src/artemis/fast_grid_scan_plan.py
```diff
@@ -58,6 +63,19 @@ def do_fgs():
...
+        parameters.grid_scan_params.grid_position_to_motor_position(xray_centre)
+    )
+    yield from bps.mv(
+        sample_motors.x,
+        xray_centre_motor_position.x,
+        sample_motors.y,
+        xray_centre_motor_position.y,
+        sample_motors.z,
+        xray_centre_motor_position.z,
+    )
```
bentom08, line 77: Must - Since #141 just got merged the test for this method will also need to be updated to mock this section

src/artemis/tests/test_zocalo_interaction.py
```diff
@@ -80,3 +82,41 @@ def test_run_start_and_end(
...
+@patch("zocalo.configuration.from_file")
+@patch("src.artemis.zocalo_interaction.lookup")
+def test_when_message_recieved_from_zocalo_then_point_returned(
+    mock_transport_lookup, mock_from_file, mock_wrap_subscribe
+):
+    message = {
+        "max_voxel": [3, 5, 5],
+        "centre_of_mass": [2.942925659754348, 7.142683401382778, 6.79110544979448],
+    }
+    step_params = {"dcid": "8183741", "dcgid": "7263143"}
```
bentom08, line 97: should - the values here which are also used further below (dcid and max_voxel) should be put into constants

bentom08 (changes requested):
Needs to be rebased on main after which run_gridscan test will break and need updating

DominicOram:
After discussing with @ndg63276 and @npat113 it's ok to assume `omega = 0` for now. We will fix this in https://github.com/DiamondLightSource/python-artemis/issues/154

DominicOram:
As seen on beamline the assert for the omega needs to be less severe and just check that they are close

bentom08 (approved):
Looks good

PR #120, opened 2022-06-14
#117: Created device for backlight and start of snapshot plan

### LINKED ISSUE #117: Snapshots: Create ophyd device for moving the backlight
To do https://github.com/DiamondLightSource/python-artemis/issues/91 we will need to be able to move the backlight in and out so we will need an ophyd device for it.

ndg63276:
Fixes #117 

To test:
* `pipenv run python artemis/snapshot_plan.py` moves the backlight


DominicOram (approved):
Tested on beamline and works. Happy to ignore coverage as I don't think there's much we can test here

PR #132, opened 2022-06-21
Udc default state

npat113:
moving stuff from my fork to main

### DIFF
--- a/src/artemis/UDC_move.to.default_state.py
+++ b/src/artemis/UDC_move.to.default_state.py
@@ -1,0 +1,106 @@
+from bluesky import RunEngine
+from devices.I03Smargon import I03Smargon
+from devices.backlight import Backlight
+from devices.BeamStop import BeamStop
+from devices.miniap import MiniAperture
+from devices.FluorescenceDetector import FluorescenceDetector
+from devices.Scintillator import Scintillator
+from devices.Scatterguard import Scatterguard
+from devices.GonioLowerStages import GonioLowerStages
+from devices.Cryostream import Cryo
+from devices.CTAB import CTAB
+from bluesky.plan_stubs import mv, null
+import bluesky.plan_stubs as bps
+import bluesky.plans as bp
+from datetime import datetime
+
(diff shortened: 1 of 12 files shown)


DominicOram:
See https://github.com/DiamondLightSource/python-artemis/issues/133

src/artemis/UDC_move.to.default_state.py
```diff
@@ -0,0 +1,106 @@
...
+		yield from bps.mv(bs.y, 45, bs.x, 1.6) 
+		
+
+def check_temp():
+	cryo_temp = cryo.temp.get()
+	if cryo_temp > 110:
+		print ("temperature too high - ",cryo_temp)  # TODO abort here
+	else:
+		pass
+	yield from null()
```
DominicOram, line 71: I suggested this line to make it work without much explanation so here is some. 

You can think of Bluesky plans as providing a list (actually a [generator](https://wiki.python.org/moin/Generators)) of small operations we're going to do e.g. a `rd` or a `mv`. It is useful to provide the list rather than doing the operations as then we can do things like ask a plan what operations it's going to perform without actually running them on real hardware, it also makes pausing and rewinding plans possible. When you're doing `yield from check_temp` in this context it's sort of like saying "Add the operations that check_temp returns to the list of operations to do". It was throwing an error because without the `yield from null()` here `check_temp` was returning no operations. `Null` is a sort of dummy operation.

**TLDR**: When using a function in Bluesky e.g. `yield from x` x must do at least one `yield from `

DominicOram (changes requested):
Some comments in code. The files should also follow python naming conventions and be in snake case e.g. `UDC_move_to_default_state.py`

DominicOram:
Happy for you address the comments I have, disagree with them or ask us to do them instead, whatever you prefer.

src/artemis/devices/I03Smargon.py
```diff
@@ -0,0 +1,11 @@
+
+from ophyd import Device, EpicsMotor
+from ophyd import Component as Cpt
+
+class I03Smargon(Device):
```
DominicOram, line 5: **should**: We actually already have the smargon in the codebase, it was just badly named. I've renamed it at https://github.com/DiamondLightSource/python-artemis/pull/112/commits/488143aacf30099d4f48c8354ea6767a284d1aa7

npat113:
As this was a pull in from a fork, I've already moved it across to a branch of main. Will incorporate the suggestions and submit from there.

DominicOram:
Ok, I will close this then in that case

PR #139, opened 2022-06-23
125 get undulator gap

### LINKED ISSUE #125: Undulator gap: Get using ophyd
Currently the undulator gap is passed from GDA to Artemis. There is no reason that Artemis could not read the undulator gap itself, this would be good as it will make Artemis more standalone and less dependent on GDA.

## Acceptance Criteria
* There is an Ophyd device for the undulator
* Artemis reads the gap from this device and GDA no longer sends it

MattPrit:
Fixes #125

Undulator device tested using:
```
from bluesky import RunEngine

undulator = Undulator(name="Undulator", prefix="SR03I-MO-SERVC-01:")

def get_value():
    gap = yield from bps.rd(undulator.gap)
    print(gap)

RE = Runengine({})
RE(get_value())
```
Output was compared with the relevant PV using `caget SR03I-MO-SERVC-01:BLGAPMTR.RBV` and found to be consistent.

Changes needed on the GDA side:
- No longer pass undulator_gap parameter to artemis
- Need to pass insertion_prefix parameter to artemis

### DIFF
--- a/src/artemis/fast_grid_scan_plan.py
+++ b/src/artemis/fast_grid_scan_plan.py
@@ -15,6 +15,7 @@ from ophyd.log import config_ophyd_logging
 from src.artemis.devices.eiger import EigerDetector
 from src.artemis.devices.fast_grid_scan import FastGridScan, set_fast_grid_scan_params
 from src.artemis.devices.zebra import Zebra
+from src.artemis.devices.undulator import Undulator
 from src.artemis.ispyb.store_in_ispyb import StoreInIspyb
 from src.artemis.nexus_writing.write_nexus import NexusWriter
 from src.artemis.parameters import SIM_BEAMLINE, FullParameters
@@ -31,10 +32,20 @@ config_ophyd_logging(file="/tmp/ophyd.log", level="DEBUG")
 # Start analysis run collection
 
 
+def update_params_from_epics(parameters: FullParameters):
+    undulator = Undulator(
+        name="Undulator",
+        prefix=f"{parameters.beamline}-MO-SERVC-01:"
+    )
+    undulator_gap = yield from bps.rd(undulator.gap)
+    parameters.ispyb_params.undulator_gap = undulator_gap
+
+
 @bpp.run_decorator()
 def run_gridscan(
     fgs: FastGridScan, zebra: Zebra, eiger: EigerDetector, parameters: FullParameters
 ):
+    yield from update_params_from_epics(parameters)
     ispyb = StoreInIspyb("config", parameters)
     _, datacollection_id, datacollection_group_id = ispyb.store_grid_scan()
     run_start(datacollection_id)
--- a/src/artemis/fast_grid_scan_plan.py
+++ b/src/artemis/fast_grid_scan_plan.py
@@ -15,8 +15,8 @@ from bluesky.utils import ProgressBarManager
 from ophyd.log import config_ophyd_logging
 from src.artemis.devices.eiger import EigerDetector
 from src.artemis.devices.fast_grid_scan import FastGridScan, set_fast_grid_scan_params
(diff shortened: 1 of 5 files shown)


src/artemis/fast_grid_scan_plan.py
```diff
@@ -45,7 +52,7 @@ def run_gridscan(
 
     for id in datacollection_ids:
         run_start(id)
-
+
```
DominicOram, line 55: **nit**: Can you remove the superfluous whitespace? I believe vscode should do this for you when you save the file though

abbiemery (changes requested):
Change the name of the test then should be good to go.

src/artemis/fast_grid_scan_plan.py
```diff
@@ -91,7 +98,9 @@ def get_plan(parameters: FullParameters):
     )
     zebra = Zebra(name="zebra", prefix=f"{parameters.beamline}-EA-ZEBRA-01:")
 
-    return run_gridscan(fast_grid_scan, zebra, eiger, parameters)
+    undulator = Undulator(name="undulator", prefix=f"{parameters.beamline}-MO-SERVC-01:")
```
DominicOram, line 101: **must**: This isn't quite going to work. The beamline prefix is `BL03I` but the prefix for the undulator is `SR03I`. I think to solve this we need a new parameter called something like `insertion_prefix` that GDA will pass in

src/artemis/tests/test_fast_grid_scan_plan.py
```diff
@@ -23,3 +25,14 @@ def test_given_full_parameters_dict_when_detector_name_used_and_converted_then_d
 def test_when_get_plan_called_then_generator_returned():
     plan = get_plan(FullParameters())
     assert isinstance(plan, types.GeneratorType)
+
+
+def test_parameters_updated_from_epics_devices_correctly():
```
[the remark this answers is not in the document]
MattPrit: Agreed. A more general test can be added later.

src/artemis/fast_grid_scan_plan.py
MattPrit: `beamline` replaced with new parameter `insertion_prefix`

src/artemis/fast_grid_scan_plan.py
MattPrit: yes

abbiemery:
@DominicOram Is it okay in this project to merge PRs that have codecov failing?

DominicOram:
> @DominicOram Is it okay in this project to merge PRs that have codecov failing?

Yes, you improved the code coverage with this PR so it's fine. We have some other tickets to get the general coverage up to the target.

src/artemis/fast_grid_scan_plan.py
```diff
@@ -91,7 +102,11 @@ def get_plan(parameters: FullParameters):
     )
     zebra = Zebra(name="zebra", prefix=f"{parameters.beamline}-EA-ZEBRA-01:")
 
-    return run_gridscan(fast_grid_scan, zebra, eiger, parameters)
+    undulator = Undulator(
+        name="undulator", prefix=f"{parameters.beamline}-MO-SERVC-01:"
```
DominicOram, line 106: **must**: This is still using the beamline prefix rather than the insertion prefix.

DominicOram (changes requested):
Thanks for adding the new prefix but we still need to use it for the Undulator...

src/artemis/fast_grid_scan_plan.py
MattPrit: Whoops. Fixed.

DominicOram (approved):
Looks good to me, thanks! Feel free to merge, I think you've fixed @abbiemery's comment.

PR #142, opened 2022-06-23
Store 2 sets of snapshots for 3d scans

### LINKED ISSUE #134: GDA: Take two snapshots
To expedite https://github.com/DiamondLightSource/python-artemis/issues/91 it would be easier to take the two snapshots in GDA for now and then work on adding more to Artemis later.

# Acceptance Criteria
* GDA takes two snapshots with a 90 deg rotation between them
* GDA passes the path for both snaphots to Artemis
* Artemis correctly puts these into ISpyB and they are correctly shown on synchweb

bentom08:
Fixes #134

To Test -

Artemis side should just need unit tests to pass, GDA side of this issue will need beamline testing


DominicOram:
See https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/35058/11..12 for GDA side

bentom08:
https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/35058/11..13 new GDA patchset for this change

src/artemis/ispyb/ispyb_dataclass.py
```diff
@@ -29,7 +29,8 @@ class IspybParams:
     )
 
     synchrotron_mode: str
-    xtal_snapshots: List[str]
+    xtal_snapshots_y: List[str]
+    xtal_snapshots_z: List[str]
```
DominicOram, line 33: **nit**: It's only y/z 'cos we're assuming 0/90. Maybe better to do something like omega start and omega end?

DominicOram (changes requested):
Looks good generally, need to fix the issue in GDA

DominicOram (approved):
Great, thank you!

PR #148, opened 2022-06-24
138 create example json

### LINKED ISSUE #138: Documentation: Create an example json file for testing
We have some documentation in https://github.com/DiamondLightSource/python-artemis#starting-a-scan that describes how to start a scan but does not have a full set of parameters. Equally these parameters are unwieldy to use from the command line. 

# Acceptance Criteria
* We should create an example json file and update the command in the README to use that
* This file should be run through Artemis in the unit tests

MattPrit:
Fixes #138

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,35 +37,8 @@ Starting a scan
 
 To start a scan you can do the following:
 ```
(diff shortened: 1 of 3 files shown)


DominicOram (approved):
Great, thank you!

PR #149, opened 2022-06-27
93 atomic nexus updates

### LINKED ISSUE #93: Nexus file updates should be atomic

MattPrit:
Fixes #93


src/artemis/nexus_writing/write_nexus.py
```diff
@@ -231,7 +232,10 @@ def __enter__(self):
 
     def __exit__(self, *_):
```
DominicOram, line 233: **should**: Can you add a docstring on the `__exit__` which says why we're creating the temp file etc?

DominicOram (changes requested):
Generally looks good, thanks! It would be good if we can create a test that will fail on main but not on here. Maybe assert that `open` is not called on the nexus file when you call `__exit__`?

src/artemis/nexus_writing/write_nexus.py
MattPrit: Done

DominicOram (approved):

PR #153, opened 2022-06-28
126 ophyd get synchrotron mode

### LINKED ISSUE #126: Synchrotron Mode: Get using Ophyd
Currently the synchrotron mode is passed from GDA to Artemis. There is no reason that Artemis could not read the synchrotron mode itself, this would be good as it will make Artemis more standalone and less dependent on GDA.

## Acceptance Criteria
* There is an Ophyd device for the synchrotron mode
* Artemis reads the synchrotron mode from this device and GDA no longer sends it

abbiemery:
fixes #126


abbiemery:
Ran the following as a sanity check to get the pv value via bluesky. returns same value as `caget CS-CS-MSTAT-01:MODE`.

```
from bluesky import RunEngine

from src.artemis.devices.synchrotron import Synchrotron

synchrotron = Synchrotron(name="Synchrotron")
synchrotron.wait_for_connection()


def get_value():
    mode = yield from bps.rd(synchrotron.machine_status.synchrotron_mode)
    print(mode)


RE = RunEngine({})
RE(get_value())
```


To do @DominicOram :
- Stop GDA passing in the value

src/artemis/devices/synchrotron.py
```diff
@@ -0,0 +1,28 @@
...
+    )
+    topup_end_countdown: EpicsSignal = Component(
+        EpicsSignal, "SR-CS-FILL-01:ENDCOUNTDN"
+    )
+
+
+class Synchrotron(Device):
+
+    machine_status: SynchrotoronMachineStatus = Component(SynchrotoronMachineStatus, "")
+    top_up: SynchrotronTopUp = Component(SynchrotronTopUp, "")
```
DominicOram, line 26: **should**: You can do the following to remove repetition elsewhere in the device.
```suggestion
    machine_status: SynchrotoronMachineStatus = Component(SynchrotoronMachineStatus, "CS-CS-MSTAT-01:")
    top_up: SynchrotronTopUp = Component(SynchrotronTopUp, "SR-CS-FILL-01:")
```

DominicOram (changes requested):
Looks good! Minor comment in code, also can you add the `Fixes` keyword to the PR description so that merging it will close the associated issue

DominicOram (approved):
Looks good, thank you!

PR #156, opened 2022-06-29
140 slit gaps using ophyd

### LINKED ISSUE #140: Slit gaps: Get using Ophyd
We should be able to get the slit gaps using Ophyd rather than passing from GDA, as we've done in https://github.com/DiamondLightSource/python-artemis/issues/126

The PVs are:
* BL03I-AL-SLITS-04:YGAP
* BL03I-AL-SLITS-04:XGAP

abbiemery:
Fixes #140


DominicOram (approved):
Great, thank you!

PR #157, opened 2022-07-01
Do Image recognition for both omega rotations in a 3D scan

### LINKED ISSUE #135: GDA: Do image recognition for both snapshots
Once https://github.com/DiamondLightSource/python-artemis/issues/134 is done we should use the image recognition in GDA to find the grid parameters to use and pass these on to Artemis.

bentom08:
Fixes #135 

This is a pull request for the artemis portion of this, the bulk of the change is in GDA, here - https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/35058/13..20


src/artemis/parameters.py
```diff
@@ -48,7 +48,7 @@ class FullParameters:
         undulator_gap=1.0,
         pixels_per_micron_x=None,
         pixels_per_micron_y=None,
-        upper_left=Point2D(x=None, y=None),
+        upper_left=Point3D(x=None, y=None, z=None),
```
DominicOram, line 51: **should** can you put a comment here about how we're sort of cheating here and it should actually be 2x 2D upper lefts. Also can you comment that it is in pixels

src/artemis/ispyb/store_in_ispyb.py
```diff
@@ -38,6 +42,11 @@ def store_grid_scan(self, full_params: FullParameters):
         self.run_number = self.detector_params.run_number
         self.omega_start = self.detector_params.omega_start
         self.xtal_snapshots = self.ispyb_params.xtal_snapshots_omega_start
+        self.upper_left = Point2D(
+            self.ispyb_params.upper_left.x, self.ispyb_params.upper_left.y
+        )
+        self.y_steps = full_params.grid_scan_params.y_steps
+        self.y_step_size = full_params.grid_scan_params.y_step_size
```
DominicOram, line 49: **must**: You're storing these in `self` but when you actually write them in `_store_grid_info_table` you use the `full_parameters`

DominicOram: Can you write a test that would have caught this too please?

DominicOram (changes requested):
Generally ok, see comments

DominicOram (approved):
Looks good to me, thanks

PR #159, opened 2022-07-07
Fix pre-commit

MattPrit:
With the current configuration, sometimes the changes made by isort are incompatible by the changes made by black. Also, the version of black currently used seems to be incompatible with the version of click used. The both result in pre-commit failing.

Here we configure isort to use the black profile and update black so that it works with click.


DominicOram (approved):
Great, thank you

PR #167, opened 2022-07-15
(#165) Created a composite device that we can wait on for connection

### LINKED ISSUE #165: Wait for connection on all devices before starting plan
On first starting Artemis it takes some time for the PVs to connect to the devices. We need to ensure that they are connected before continuing with plans.

DominicOram:
Fixes #165 

To test:
* In a python terminal run:
```python
from src.artemis.parameters import FullParameters
from src.artemis.devices.fast_grid_scan_composite import FGSComposite

params = FullParameters()
params.insertion_prefix = "SR03I"

fgs = FGSComposite(name="test", full_params=params, prefix="BL03I")
fgs.wait_for_connection()
```
* Confirm that this passes and so all devices are connected

### DIFF
--- a/src/artemis/devices/fast_grid_scan_composite.py
+++ b/src/artemis/devices/fast_grid_scan_composite.py
@@ -1,0 +1,24 @@
+from ophyd import Component, Device, FormattedComponent
+
+from src.artemis.devices.fast_grid_scan import FastGridScan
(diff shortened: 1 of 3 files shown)


MattPrit (approved):
Looks good to me

PR #176, opened 2022-07-18
Startup bash script for deploying, stopping and starting Artemis

### LINKED ISSUE #163: Deployment and start up script for Artemis
As a developer it's frustrating to have to remember how to deploy Artemis every time I go down to the beamline. We should:
* Tag a `0.1` release
* Document how to tag a release 
* Have a script that:
    * Pulls a specified version (or the latest if none specified)
    * Starts Artemis, making sure to:
         * module unload controls_dev
         * pipenv install
         * module load dials
         * set ispyb config

bentom08:
Fixes #163

Current behaviour -

Will fail if --deploy is run and the user running the script is not in dls_dasc group due to file permissions. Permissions will ned changing to resolve this (although it may not be such a bad thing for dls_dasc to be the only group with permission to deploy new versions)

By default the script at the moment will both start and deploy a new Artemis version. --deploy and --start flags are used to only perform those actions. The other option I could think of would be to have the default action be just starting the server and have --deploy only deploy and use both tags (--deploy --start) to perform both actions.

The --stop flag will currently override the --start and --deploy flags and only stop the currently running artemis.

If any of these should/could be changed let me know in the review,

### DIFF
--- a/run_artemis.sh
+++ b/run_artemis.sh
@@ -32,7 +32,7 @@ checkver () {
 }
 
 STOP=0
-START=0
+START=1
 DEPLOY=0
 for option in "$@"; do
     case $option in
@@ -47,8 +47,8 @@ for option in "$@"; do
         --stop)
             STOP=1
             ;;
-        --start)
-            START=1
+        --no-start)
+            START=0
             ;;
         --deploy)
             DEPLOY=1
@@ -63,10 +63,10 @@ for option in "$@"; do
             echo "Operations"
             echo "  --stop                  Used to stop a currently running instance of Artemis. Will override any other operations"
             echo "                          options"
-            echo "  --start                 Used to start the Artemis server with the currently installed version."
             echo "  --deploy                Used to update and install a new version of Artemis."
+            echo "  --no-start              Used to specify that the script should be run without starting the server."
             echo " "
-            echo "If no Operations options are specified both --deploy --start will be used by default."
+            echo "By default this script will start an Artemis server unless the --no-start flag is specified."
             exit 0
             ;;
         -*|--*)
@@ -75,11 +75,6 @@ for option in "$@"; do
             ;;
     esac
 done
-
-if [[ $START == 0 && $DEPLOY == 0]]; then
-    START=1
-    DEPLOY=1
-fi
 
 if [[ -z "${BEAMLINE}" ]]; then
     echo "BEAMLINE parameter not set. Use the option -b, --beamline=BEAMLNE to set it manually"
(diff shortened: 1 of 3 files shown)


DominicOram (changes requested):
Generally looks good to me. I think the default behavior should be just to start then specifying `--deploy` will also deploy. Just 'cos it's more dangerous to do an accidental deploy than an accidental restart.

Also, can you fix the test by adding a default ISPYB config, maybe default to the dev database?

src/artemis/fast_grid_scan_plan.py
```diff
@@ -62,11 +62,11 @@ def run_gridscan(
     yield from update_params_from_epics_devices(
         parameters, undulator, synchrotron, slit_gap
     )
-    config = "config"
+    ispyb_config = os.environ["ISPYB_CONFIG_PATH"]
```
DominicOram, line 65: **should**: We can just add the default here rather than in the test itself using `print(os.environ.get('KEY_THAT_MIGHT_EXIST', default_value))`

run_artemis.sh
```diff
@@ -0,0 +1,148 @@
...
+            echo "  -b, --beamline=BEAMLINE Overrides the BEAMLINE environment variable with the given beamline"
+            echo "  -v, --version=VERSION   Specifies the artemis version number to deploy. Option should be given in the form 0.0.0.0"
+            echo "                          Will check git tags and use the lastest version as a default if no version is specified."
+            echo "                          Unused outside of deploy operation."
+            echo " "
+            echo "Operations"
+            echo "  --stop                  Used to stop a currently running instance of Artemis. Will override any other operations"
+            echo "                          options"
+            echo "  --deploy                Used to update and install a new version of Artemis."
+            echo "  --no-start              Used to specify that the script should be run without starting the server."
```
DominicOram, line 67: **nit**: This is a bit weird cos running the script as `run_artemis --no-start` basically does nothing so I guess `--no-start` only really makes sense in conjunction with `--deploy` but I think it's fine.

DominicOram (changes requested):
Thanks, I think you can make it slightly tidier, otherwise fine

run_artemis.sh
bentom08: I could also use --deploy-no-start instead of the --no-start flag to specify a deployment without a server start, but neither is particularly nice if we want that option to be available. If you have any other ideas for it I'd be glad to do the implementation.

DominicOram: No, I can't really think of any cleaner way necessarily, just highlighting it as odd behavior. We'll leave it as and maybe come up with a better idea for it in the future.

DominicOram (approved):

Issue #164, opened 2022-07-14
Use a recommender pattern for external interactions in the plan


DominicOram:
After discussion with NSLS-II they suggested that rather than have the interactions with ISPyB etc. inside the plan they should be happening in a separate object outside the plan that is listening on the Bluesky event stream. This object can then recommend to the plan how to continue e.g in pseudo code something like:

```python
def fgs_plan(recommender)
    # Some hardware interactions here, emitting documents as we go that the recommender is picking up on
    results = recommender.wait_on_results()
    # Move to results

recommender = FGSRecommender(params) # This object does all interaction with Ispyb, nexus writing and zocalo
RE.subscribe(recommender)
RE(fgs_plan())
```

DominicOram:
@callumforrester, something like this sound about right?

callumforrester:
Looks correct, can also make the recommender use optional for flexibility

```python
if recommender.has_result():
    # Move to results
else:
    # Continue as normal
```

DominicOram:
It probably makes sense to split this ticket into moving each kind of interaction out of the plan.

Some more pseudo-code after discussion with @callumforrester:
```
def my_plan():
    cache = SingleRunCache()
    yield from subscribe(cache)

    yield from open_run()
    yield from bps.trigger_and_read(pvs)
    yield from close_run()

    data = cache[".../..."]

    yield from open_run()
    yield from bp.scan(data.thing)
    yield from close_run()

def artemis_plan():
    ispyb_conn = IspybConn()
    yield from subscribe(ispyb_conn)
    yield from record_start_state()

    if not ispyb_conn.validate():
        raise ...
    key = ispyb_conn.get_key()
    

    @run_decorator(md={"key": key})
    def do_actual_scan():
        zocalo = ZocaloConn(key)
        yield from subscribe(zocalo)

        # scan logic

    yield from do_actual_scan()

@run_decorator
def record_start_state():
    yield from, bps.trigger_and_read(thing1)
    yield from, bps.trigger_and_read(thing2)
    yield from, bps.trigger_and_read(thing3)
    yield from, bps.trigger_and_read(thing4)
```

PR #177, opened 2022-07-19
100: Write 2nd nexus file

### LINKED ISSUE #100: 3D Grid Scans: Create 2nd Nexus File
For 3D grid scans to work a second nexus file needs to be created. This was done in https://github.com/DiamondLightSource/python-artemis/tree/testing_round_5 but needs tidying up

## Notes
* The second file should have the same template for the filename but with the run number incremented
* Artemis should still be able to do 2D grid scans

DominicOram:
Fixes #100 

Writes multiple nexus files based on some helper functions.

To test:
* Confirm unit tests still pass and the difference between the files looks sensible


src/artemis/nexus_writing/write_nexus.py
```diff
@@ -48,6 +49,35 @@
...
 
 
+def create_parameters_for_first_file(parameters: FullParameters):
+    new_params = deepcopy(parameters)
+    new_params.grid_scan_params.z_axis = GridAxis(
+        parameters.grid_scan_params.z1_start, 0, 0
+    )
+    new_params.detector_params.num_images = (
+        parameters.grid_scan_params.x_steps * parameters.grid_scan_params.y_steps
+    )
```
bentom08, line 59: Does this need setting in the 2nd file params method as well? if y/z steps are different the 2 files should have different numbers of images

src/artemis/fast_grid_scan_plan.py
```diff
@@ -84,8 +88,9 @@ def do_fgs():
         yield from bps.kickoff(fgs_motors)
         yield from bps.complete(fgs_motors, wait=True)
 
-    with NexusWriter(parameters):
-        yield from do_fgs()
+    with NexusWriter(create_parameters_for_first_file(parameters)):
+        with NexusWriter(create_parameters_for_second_file(parameters)):
+            yield from do_fgs()
```
bentom08, line 93: This double indentation is a little awkward, and I'm not a big fan of the plan calling both file writers separately (it makes them harder to split up when we want to do 2D scans, and doesn't work well with the potential future implementation of the ispyb writing in this PR https://github.com/DiamondLightSource/python-artemis/pull/186 )

Maybe a wrapper/superclass that calls both filewriters, so we only need to call the nexuswriter from the plan once, and it can then write 1 or 2 files as appropriate?

bentom08 (changes requested):

src/artemis/fast_grid_scan_plan.py
rjgildea: FYI you _can_ [write multiple items in a single `with` statement](https://docs.python.org/3/reference/compound_stmts.html#the-with-statement), e.g.:
```
with A() as a, B() as b:
    # code
```
or
```
with (
    A() as a,
    B() as b,
):
    # code
```

DominicOram: I've changed to the multiple with statements, I don't like it but I think a larger refactor should be addressed in https://github.com/DiamondLightSource/python-artemis/issues/164

src/artemis/nexus_writing/write_nexus.py
DominicOram: No, the second nexus file needs to know the full number of images that have been taken, which is what will be in `new_params.detector_params.num_images` if we don't override it. Effectively this value stores the upper limit for the VDS slicing. It's messy, I agree, really needs a second refactor but think we'll do it in https://github.com/DiamondLightSource/python-artemis/issues/164

bentom08 (approved):

PR #180, opened 2022-07-19
#179: Index list returned from zocalo

### LINKED ISSUE #179: Zocalo returns a list of messages
During testing on the beamline we found that zocalo was returning a list of responses and so we were getting an error, we should make sure that we treat it like a list

DominicOram:
Fixes #179 

To test:
* Confirm modified unit test passes here but fails on main
* Confirm this matches the released code in `/dls_sw/i03/software/artemis`


DominicOram:
@MattPrit, you saw this issue with me on the beamline, can you review it?

MattPrit (approved):

PR #183, opened 2022-07-21
128 overlay grid on the oav

### LINKED ISSUE #128: Snapshots: Overlay grid on the OAV
After a snapshot image is taken in GDA the gridscan grid is overlayed on top of the picture. We should do this in Artemis. We can use how GDA does it as a guide but we can also look at leveraging other Python 3 libraries if they're helpful.

MattPrit:
Fixes #128


src/artemis/devices/unit_tests/test_grid_overlay.py
```diff
@@ -0,0 +1,149 @@
...
+    add_grid_overlay_to_image(
+        image, top_left_x, top_left_y, box_width, num_boxes_x, num_boxes_y
+    )
+    expected_calls = 2 * [call(image)] + [
+        call(image).line(line) for line in expected_lines
+    ]
+    actual_calls = mock_imagedraw.mock_calls
+    expected_calls.sort()
+    actual_calls.sort()
+    assert actual_calls == expected_calls
```
DominicOram, line 149: **nit**: Could you move the code shared between this and the above into a function they both call?

src/artemis/devices/oav/grid_overlay.py
```diff
@@ -0,0 +1,116 @@
...
+            (start_x + i * spacing, start_y + line_length),
+        )
+        for i in range(num_lines)
+    ]
+    draw = ImageDraw.Draw(image)
+    for line in lines:
+        draw.line(line)
+
+
+_add_vertical_parallel_lines_to_image = partial(
```
DominicOram, line 30: Nice use of partials!

DominicOram (changes requested):
Generally great, thanks! Some minor comments in code

DominicOram (approved):
Great, thank you!

DominicOram:
Happy to merge w/o coverage target check.

PR #186, opened 2022-07-21
#77: Inform ISpyB if scan fails

### LINKED ISSUE #77: Inform ISpyB if scan fails
Currently we're writing the scan successful state to ISPyB but not the unsuccessful state. We should write unsuccessful if the scan fails for whatever reason

MattPrit:
Fixes #77

### DIFF
--- a/src/artemis/ispyb/store_in_ispyb.py
+++ b/src/artemis/ispyb/store_in_ispyb.py
@@ -2,12 +2,12 @@ import datetime
 import re
 from abc import ABC, abstractmethod
 
+import ispyb
 from sqlalchemy.connectors import Connector
+
 from src.artemis.ispyb.ispyb_dataclass import Orientation
 from src.artemis.parameters import FullParameters
 from src.artemis.zocalo_interaction import run_end, run_start, wait_for_result
-
-import ispyb
 
 I03_EIGER_DETECTOR = 78
 EIGER_FILE_SUFFIX = "h5"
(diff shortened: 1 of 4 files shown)


MattPrit:
Note that by switching to StoreInIspyb using a context manager, several tests are now broken.

src/artemis/ispyb/store_in_ispyb.py
```diff
@@ -30,6 +31,38 @@ def __init__(self, ispyb_config):
...
+                current_time,
+                run_status,
+                id,
+                self.datacollection_group_id,
+            )
+
+        for id in self.datacollection_ids:
+            run_end(id)
+
+        wait_for_result(self.datacollection_group_id)
```
bentom08, line 64: *Must* - We don't assign the value `wait_for_result` returns on the main branch at the moment, but we will need the value when we want to move to the position analysis gives us. I would suggest returning the ids from the __enter__ method and assigning them to variables in the plan. So it would look something like:

```
with ispyb as ids:
        dc_ids = ids[0]
        dcg_id = id[2]
        run_start(dc_ids)
        do_fgs()

run_end(dc_ids)
wait_for_results(dcg_id)
```

bentom08 (changes requested):
Looking over this I think we'll need to separate the Zocalo and Ispyb stuff after all.

Sorry for not picking up on this earlier, unfortunately I didn't see it when we were going through it before.

bentom08 (approved):

PR #190, opened 2022-07-26
#181: Change setup to package format

### LINKED ISSUE #181: Move to using setup.cfg
We should move the build system to using setup.cfg and subsequently sort out all the imports so they don't need the `src`

abbiemery:
Fixes #181

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -3,28 +3,11 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 name = "pypi"
 
-[packages]
-bluesky = "*"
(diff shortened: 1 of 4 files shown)


abbiemery:
Just to reiterate the comment I made in the issue (mostly for future Abi).

I've done this for now and moved the build and setup into setup.cfg, however in the future we might need to change it again to instead use the pyproject.toml as raised by [pep 621](https://peps.python.org/pep-0621/) and defined [here](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/). Support for this in setuptools was introduced in 61.0.0 but not everything is quite on board yet. As such I did it the old way for convenience.

PR #194, opened 2022-07-27
Edit config files so that the correct coverage reports are produced

MattPrit:
Fixes an issue introduced in #190, whereby the coverage CI was slightly broken due to pytest not using setup.cfg as its configuration file.


abbiemery:
Seams to work all on my end, cheers for moving that to the toml

abbiemery (approved):
Seams to work all on my end, cheers for moving that to the toml

PR #195, opened 2022-07-28
Fix build system

MattPrit:
This adds a minimal `setup.py`, required by some versions of setuptools for an editable install of our package. It also introduces a change in how the bluesky runner is started.

### DIFF
--- a/run_artemis.sh
+++ b/run_artemis.sh
@@ -129,7 +129,7 @@ if [[ $START == 1 ]]; then
     fi
 
     if [[ -z $(pgrep -f artemis) ]]; then
-        pkill -f src/artemis.main.py
+        pkill -f artemis
     fi
 
     cd ${ARTEMIS_PATH}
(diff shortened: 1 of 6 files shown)


run_artemis.sh
```diff
@@ -128,7 +128,7 @@ if [[ $START == 1 ]]; then
         ssh -T -o BatchMode=yes -i ${SSH_KEY_FILE_LOC} gda2@${BEAMLINE}-control.diamond.ac.uk
     fi
 
-    if [[ -z $(pgrep -f src/artemis/main.py) ]]; then
+    if [[ -z $(pgrep -f artemis) ]]; then
         pkill -f src/artemis.main.py
```
DominicOram: **must**: Need to change this kill too.

DominicOram (approved):
Looks good, minor issue that I fixed myself. Otherwise thanks for fixing docs etc.

PR #196, opened 2022-07-29
Correct axes on det dist lookup table and update to latest values

### LINKED ISSUE #189: Beam Centres are wrong
In testing we found that the beam centre Artemis is calculating is different from the one that GDA gives. Part of this was due to the lookup table in Artemis being out of date, but after correcting it there was still a discrepency.

We need to investigate why this discrepancy exists and correct it.

bentom08:
Fixes #189


src/artemis/devices/det_dist_to_beam_XY_converter.txt
```diff
@@ -1,10 +1,10 @@
 #Table giving position of beam X and Y as a function of detector distance
 #Units mm mm mm
 # Eiger values
-# distance beamY beamX (values from mosflm)
+# distance beamX beamY (values from mosflm)
```
DominicOram, line 4: **should**: Can you make the same change in GDA?

src/artemis/devices/det_dist_to_beam_converter.py
```diff
@@ -4,8 +4,8 @@
 
 
 class Axis(Enum):
-    Y_AXIS = 1
-    X_AXIS = 2
+    Y_AXIS = 2
+    X_AXIS = 1
```
DominicOram, line 8: **nit** can you swap the order of these? It's a bit confusing as is.

DominicOram (changes requested):
Looks good, thanks. Minor comments in code

src/artemis/devices/det_dist_to_beam_XY_converter.txt
bentom08: I've made ticket MXGDA-3879 to address this.

bentom08:
GDA change addressed in https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/36541

DominicOram (approved):
Looks good to me!

PR #192, opened 2022-07-27
#94: Add tests for eiger.py

MattPrit:
Fixes #94

PR #198, opened 2022-07-29
Set correct Odin PVs and wait for filename to propagate to meta writer

### LINKED ISSUE #184: Wait needed after Odin filename PVs are set
When testing multiple subsequent collections, we found that a wait was needed after the Odin filename PVs are set in order to give Odin time to propagate them to the values to the appropriate places. We also found that some of the PVs we were setting were unnecessary. Specifically -

Currently we are setting:
OD:AcquisitionID
OD:FilePath
OD:FP:FileName
OD:META:FileName

However on the beamline we were able to produce what looks like the same results by setting only:
OD:FilePath
OD:FileName

We need to investigate exactly which PVs we need to set in Odin and how long we need to wait for them (and whether ts possible to monitor a PV instead of waiting a set amount of time)

bentom08:
Fixes #184 

To test:

Make sure the meta file is written correctly when multiple collections are queued on the beamline


DominicOram (changes requested):
Changes look good, I think I basically got `aquisition_id` completely confused. At the moment we're passing it in from GDA and now after this doing nothing with it. Can we remove it from the parameters and in GDA too?

DominicOram:
I'm not too fussed about reduced coverage in this case, I know that https://github.com/DiamondLightSource/python-artemis/pull/192 will fix it

DominicOram (approved):
Looks good, I have made the corresponding Artemis change in https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/36713/1..2

PR #199, opened 2022-07-29
reverse zocalo coordinates and use centre_of_mass rather than max_voxel

### LINKED ISSUE #188: Switch from using max_voxel to centre_of_mass
In testing on the beamline it seems that the "max_voxel" coordinate returned from analysis was too heavily influenced by rogue spots of high diffraction.

The "centre_of_mass" coordinate provided by them should give more consistent results

bentom08:
Fixes #188 

To test:

Make sure tests still pass


DominicOram (approved):
Looks good, thanks

PR #201, opened 2022-07-29
#146: Test that undulator can connect to simulation

### LINKED ISSUE #146: Simulation: Write tickit simulations for Undulator Gap
For us to be able to run system tests properly against S03 we will need a simple IOC that gives us the Undulator Gap

We can do this using the real `db` file and then the `EpicsAdapter` in `tickit`. See https://github.com/dls-controls/tickit/tree/master/tickit/devices/femto

MattPrit:
Fixes #146

### DIFF
--- a/src/artemis/devices/system_tests/test_undulator_system.py
+++ b/src/artemis/devices/system_tests/test_undulator_system.py
@@ -1,0 +1,15 @@
+import pytest
+
+from src.artemis.devices.undulator import Undulator
+from src.artemis.parameters import SIM_INSERTION_PREFIX
+
+
+@pytest.fixture
+def undulator():
+    undulator = Undulator(f"{SIM_INSERTION_PREFIX}-MO-SERVC-01:", name="undulator")
+    return undulator
+
+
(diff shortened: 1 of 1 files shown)


DominicOram:
I think we should hold off on merging this until we have the device in S03 properly. There is documentation on doing this at https://confluence.diamond.ac.uk/display/SSCC/Simulated+IOC+Builder. Can you look at adding it in?

MattPrit:
> I think we should hold off on merging this until we have the device in S03 properly. There is documentation on doing this at https://confluence.diamond.ac.uk/display/SSCC/Simulated+IOC+Builder. Can you look at adding it in?

See https://gitlab.diamond.ac.uk/controls/support/bl03s-builder/-/merge_requests/1 and https://gitlab.diamond.ac.uk/controls/python3/s03_utils/-/merge_requests/6.

abbiemery (approved):
This is fine, it only checks for connections and fails when S03 is not running and passes when it is, as we expect. 

Obviously it's not much of a test but, it does work.

PR #202, opened 2022-07-29
Change Underload to zero in nexus file

### LINKED ISSUE #187: Underload value should be zero
In write_nexus we're storing the underload in detector parameters as -1 when it should be 0

bentom08:
Fixes #187

### DIFF
--- a/src/artemis/nexus_writing/write_nexus.py
+++ b/src/artemis/nexus_writing/write_nexus.py
@@ -2,6 +2,7 @@
 Define beamline parameters for I03, Eiger detector and give an example of writing a gridscan.
 """
 import math
+import shutil
 import time
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


src/artemis/nexus_writing/write_nexus.py
```diff
@@ -103,7 +104,7 @@ def create_detector_parameters(detector_params: DetectorParams) -> Dict:
         "sensor_material": "Silicon",
         "sensor_thickness": "4.5E-4",
         "overload": 46051,
-        "underload": -1,  # Not sure of this
+        "underload": 0,  # sure of this
```
DominicOram, line 107: **should**: I think just remove the comment.

DominicOram (changes requested):
Thanks, see comment in code

DominicOram (approved):

Issue #27, opened 2022-02-11
Logging: Integrate with graylog


DominicOram:
Graylog is used for logging across DLS. It would be useful if the Artemis logs can also be sent here.

## Acceptance Criteria
* Artemis logs are sent to graylog
* The correct meta-data is sent with the logs e.g. beamline

DominicOram:
Note that during testing on i03 we were accidentally logging into the data analysis stream. This was caused by us loading in the zocalo config. As part of this ticket ensure this doesn't happen.

DominicOram:
Note, we should also ensure we're not slowed down by using graylog e.g. https://docs.python.org/3/howto/logging-cookbook.html#dealing-with-handlers-that-block use

DominicOram:
To get a local graylog up and running:
```
podman pod create --name graylog-pod
podman run -d --net host --pod=graylog-pod --name=mongo mongo:4.2
podman run -d --net host --pod=graylog-pod -e "http.host=0.0.0.0" -e "discovery.type=single-node" -e "ES_JAVA_OPTS=-Xms512m -Xmx512m" --name=elasticsearch docker.elastic.co/elasticsearch/elasticsearch-oss:7.10.0
podman run -d --net host --pod=graylog-pod -e GRAYLOG_HTTP_EXTERNAL_URI="http://localhost:9000/" -e GRAYLOG_MONGODB_URI="mongodb://localhost:27017/graylog" -e GRAYLOG_ELASTICSEARCH_HOSTS="http://localhost:9200/" --name=graylog graylog/graylog:4.0
```

DominicOram:
As part of this investigate the issue where we are currently sending logs to graylog as zocalo. I think this is a bug in zocalo that when we import the config it sets up the default logger with this behaviour. @rjgildea may be able to help more

abbiemery:
> Note that during testing on i03 we were accidentally logging into the data analysis stream. This was caused by us loading in the zocalo config. As part of this ticket ensure this doesn't happen.

https://github.com/DiamondLightSource/python-artemis/pull/203 Solved by Richard

PR #172, opened 2022-07-15
(#171) Added flake8 options and added to CI

DominicOram:
Fixes #171

PR #203, opened 2022-08-01
Use artemis-specific zocalo environment

rjgildea:
This avoids bluesky logs ending up in the Zocalo/data analysis graylog stream.

See also https://github.com/DiamondLightSource/python-artemis/issues/27

Add E501 to flake ignore: Since we are already relying on black to apply an appropriate max line length, and flake8 can sometimes disagree ignore this particular warning

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,6 +58,7 @@ max-line-length = 88
 extend-ignore =
     E203,  # See https://github.com/PyCQA/pycodestyle/issues/373
     F811,  # support typing.overload decorator
+    E501,  # line too long
 
 [coverage:run]
 # This is covered in the versiongit test suite so exclude it here
--- a/src/artemis/tests/test_zocalo_interaction.py
+++ b/src/artemis/tests/test_zocalo_interaction.py
@@ -34,7 +34,7 @@ def _test_zocalo(
 
     func_testing(mock_transport)
 
-    mock_zc.activate.assert_called_once()
+    mock_zc.activate_environment.assert_called_once_with("artemis")
     mock_transport.connect.assert_called_once()
     expected_message = {
         "recipes": ["mimas"],
--- a/src/artemis/zocalo_interaction.py
+++ b/src/artemis/zocalo_interaction.py
@@ -1,7 +1,6 @@
 import getpass
 import queue
 import socket
-from time import sleep
 
 import workflows.recipe
 import workflows.transport
@@ -15,7 +14,7 @@ TIMEOUT = 30
 
 def _get_zocalo_connection():
     zc = zocalo.configuration.from_file()
-    zc.activate()
+    zc.activate_environment("artemis")
 
     transport = lookup("PikaTransport")()
     transport.connect()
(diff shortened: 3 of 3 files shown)


DominicOram (changes requested):
Thank you, would you be able to fix the tests? I like having the check in both as we may well have contributions from people that are not regularly using `black`. I haven't seen them disagree as long as you have them set up with the same line lengths?

rjgildea:
> I like having the check in both as we may well have contributions from people that are not regularly using `black`. I haven't seen them disagree as long as you have them set up with the same line lengths?

Without ignoring E501 set, flake8 complained about the line lengths that I changed in https://github.com/DiamondLightSource/python-artemis/pull/203/commits/4bae73dfb8b96f217b9ebf65135b1e40d83dca88.  It also complained about [this line](https://github.com/DiamondLightSource/python-artemis/blob/4bae73dfb8b96f217b9ebf65135b1e40d83dca88/src/artemis/zocalo_interaction.py#L94) which black does not touch.

rjgildea:
> > I like having the check in both as we may well have contributions from people that are not regularly using `black`. I haven't seen them disagree as long as you have them set up with the same line lengths?
> 
> Without ignoring E501 set, flake8 complained about the line lengths that I changed in [4bae73d](https://github.com/DiamondLightSource/python-artemis/commit/4bae73dfb8b96f217b9ebf65135b1e40d83dca88). It also complained about [this line](https://github.com/DiamondLightSource/python-artemis/blob/4bae73dfb8b96f217b9ebf65135b1e40d83dca88/src/artemis/zocalo_interaction.py#L94) which black does not touch.

By way of standalone example:
```
$ cat long_lines.py 
a_really_long_string = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"


def function_with_really_long_docstring():
    """This doesn't really tell you anything but it is reaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaally long"""
    pass
$ black long_lines.py 
All done! ✨ 🍰 ✨
1 file left unchanged.
[hko55533@ws175 rjgildea]$ flake8 long_lines.py 
long_lines.py:1:80: E501 line too long (130 > 79 characters)
long_lines.py:5:80: E501 line too long (123 > 79 characters)
```

In my experience it is best to leave anything to do with formatting (such as line length) to black, since that makes any necessary changes automatically, and disable related checks in flake8 (since as demonstrated above, they are not necessarily consistent with black). I don't understand under what circumstances people would be running flake8 but not black.

DominicOram (approved):
We've had previous issues where people were running w/o the pre-commit hooks and so not using `black`/`flake8` and issues crept in. I think with https://github.com/DiamondLightSource/python-artemis/pull/172 and https://github.com/DiamondLightSource/python-artemis/pull/205 this is less likely to happen though so happy to blanket ignore E501

abbiemery (approved):

PR #213, opened 2022-08-08
206: Ensure directory from GDA is correct

### LINKED ISSUE #206: Folders coming from GDA do not always contain trailing /
When running a gridscan manually the folders being passed from gda contain a trailing `/`. When running from UDC they do not. Artemis should handle both

DominicOram:
Fixes #206


MattPrit (approved):
Looks good to me

PR #215, opened 2022-08-10
#145: synchrotron simulation for system tests

### LINKED ISSUE #145: Simulation: Write tickit simulations for Synchrotron mode
For us to be able to run system tests properly against S03 we will need a simple IOC that gives us the Synchroton Mode

We can do this using the real `db` file and then the `EpicsAdapter` in `tickit`. See https://github.com/dls-controls/tickit/tree/master/tickit/devices/femto

abbiemery:
fixes #145 

The simulation is built in tickit, to be hosted in s03. This PR is for the system tests which test the connection to the simulated device and check against it's default values.


MattPrit (approved):
Looks good to me

PR #216, opened 2022-08-18
29 filepath for lookuptable

### LINKED ISSUE #29: Receive and use filepath for DetDistToBeamXYConverter
DetDistToBeamXYConverter is currently stored in both artemis and GDA. It would be better for artemis to receive a filepath to the version stored in GDA, so that artemis doesn't need to be updated with changes to the GDA file, and artemis can be passed different files based on the configuration/beamline being used.

abbiemery:
Fixes #29 

@bentom08 I think this is sufficient on the Artemis end, but may you check to ensure I've not missed anything?

Also before the merge we need to make a change to GDA to pass the path. @DominicOram

### DIFF
--- a/src/artemis/devices/detector.py
+++ b/src/artemis/devices/detector.py
@@ -29,8 +29,8 @@ class DetectorParams:
     omega_start: float
     omega_increment: float
     num_images: int
-
     use_roi_mode: bool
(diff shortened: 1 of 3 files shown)


bentom08 (approved):
Looks fine, I'll make the change in GDA

Issue #221, opened 2022-09-21
Create devices when artemis starts


MattPrit:
Currently devices are created every time a plan is run. Some devices may persist once the plan has completed, so when a plan is repeated multiple times, the number of active devices can grow large. It may be beneficial to just create devices once when artemis starts up, and then pass these devices into the plan.

evalott100:
I've implemented this by adding global variables in artemis/devices/fast_grid_scan_composite.py and artemis/devices/eiger.py which initialized on the first call of get_plan and then reused rather than reinitialized. The eiger is the only device which takes parameters and is updated each time get_plan is called.

Issue #220, opened 2022-09-21
Can't install databroker


MattPrit:
To access data outside the run engine `databroker` is required. To install `databroker` in the artemis pipenv, `pandas` is also required (else `from databroker import Broker` results in `ModuleNotFoundError: No module named 'pandas'`); installing pandas fails with 
```
ERROR: Failed building wheel for numpy
      Failed to build numpy
      ERROR: Could not build wheels for numpy, which is required to install pyproject.toml-based projects
      [end of output]
```

evalott100:
I didn't reproduce the above error, however I managed to get Databroker working...

DominicOram:
Closing as not worth doing for now

PR #222, opened 2022-09-22
207 flux calculation

### LINKED ISSUE #207: Flux calculation
From GDA I would like to use the flux calculations that @npat113 has added in 207-Flux.

To get these working in GDA we need to:
* Take the work Neil has done and add some tests/tidy up if required
* Add a new endpoint in Artemis that calls these plans
* Update docs appropriately
* Modify GDA to instead call Artemis

MattPrit:
Fixes #207 

Changes:
- Adds a flux calculator device for calculating the flux from the energy and intensity values from the DCM and XBPM2.
- Adds methods and plans for doing this flux calculation.
- Adds a method for flux prediction based on expected 100% beam value given energy, ring current, aperture and transmission.
- Adds endpoints for both flux calculations

To run the flux calculation:
`curl -X PUT http://127.0.0.1:5000/flux_calculation/start --data-binary "@test_flux_calculation_parameters.json" -H "Content-Type: application/json"`
To run the flux prediction:
`curl -X PUT http://127.0.0.1:5000/flux_prediction/start --data-binary "@test_flux_prediction_parameters.json" -H "Content-Type: application/json"`

Issues:
- Flux value does not update fast enough. By the time the flux calculator has updated its `flux` signal (upon update of `energy` and `intensity` signals), the plan has already completed using the signals' default values, i.e. `0.0`. Adding `time.sleep` calls can result in a correct flux value being produces, but this is not reliable.
- `FluxCalculator` device persists after flux calculation plan/method has completed. The device will the continue to have its `energy`/`intensity` signals updated as the corresponding PVs change, and thus the `flux` signal will be repeatedly recalculated and updated indefinitely. With each call to the plan, a new `FluxCalculator` device is created. After several calls, multiple `FluxCalculators` will persist and repeatedly recalculate flux indefinitely. See #221 
- Can't get flux values out of plans. To get the flux values out of the run engine to send back in the response to the `PUT` command, a databroker is needed. In order to have the values in the response, currently the flux is also calculated outside of the run engine in a function as well as in the plan. See #220 
- Flux prediction should not really be a plan. To have the same pattern in all three Flask resources, the run engine is passed a null flux prediction plan, with the prediction calculation being made outside of the run engine.


MattPrit:
TODO: Update README with instructions for starting flux 'plans'

PR #223, opened 2022-09-23
Fix file path error

abbiemery:
To fix the accidental mess from #29 .

Moved the beam converter instantiation back from the post init by making the default field path just the look uptable for the unit tests. This was causing issues with `cannot pickle 'mappingproxy' object`.


MattPrit (approved):
Looks good to me

Issue #127, opened 2022-06-17
Create a check topup plan


DominicOram:
The synchrotron periodically has to periodically be topped up. On i03 they may not want to take data whilst this is happening and so will wait for the operation to finish in some circumstances. We should create a separate small plan that does this. We can use what GDA does currently as a guide but should also work closely with the scientists.

## Acceptance Criteria
* I can apply a suspender or run a plan that will pause or wait on topup to finish before continuing
* There is a topup_gate plan where I can provide an expected length of time and it will pause the plan if we must wait for the topup to happen in this time.
* There is a system test proving both of these (requires https://github.com/dls-controls/tickit-devices/issues/22)

DominicOram:
This may be better written as a suspender

DominicOram:
GDA code for topup gate is:
```python
def topup_gate(total_exposure_time):
	time_to_topup = synchrotron.seconds_before_start_of_topup()
	if _mode_precludes_gating(time_to_topup) or \
		_sufficient_time_before_topup(time_to_topup, total_exposure_time):
		return
	# Seconds of estimated beam instability following topup
	time_beam_unstable = float(LocalProperties.get("gda.mx.topup.instability_time"))
	threshold_percentage = float(LocalProperties.get("gda.mx.topup.threshold_percentage"))
	delay_required = \
		_topup_degrades_exposure(total_exposure_time, time_beam_unstable, threshold_percentage)
	if delay_required:
		_delay_acquisition_to_avoid_topup(total_exposure_time, time_to_topup)
	else:
		# Don't gate if top up time small compared to total exposure
		gate_message_format = """Total exposure time exceeds time to next topup (%d >= %d)s but
				topup instability of %ds is less than %d%% of exposure, so collecting anyway."""
	
		gate_message = gate_message_format % (
			total_exposure_time, time_to_topup, time_beam_unstable, threshold_percentage)
		logger.info(gate_message)
def _delay_acquisition_to_avoid_topup(total_exposure_time, time_to_topup):
		gate_message_format = """Total exposure time exceeds time to next topup (%d >= %d)s.
				Waiting to collect until after topup and beam has stabilised"""
		gate_message = gate_message_format % ( total_exposure_time, time_to_topup )
		logger.info(gate_message)
		top_up_pause_time = int(synchrotron.seconds_before_end_of_topup())
		msg = TopupPauseMessage(top_up_pause_time)
		MsgBus.publish(msg)
		sleep(top_up_pause_time)
[...]

PR #224, opened 2022-09-23
WIP: Beam topup plan attempt

MattPrit:
Fist pass at making use of a suspender that pauses artemis when beam is being topped up (#127 )
Suspension conditions logic is copied from GDA into a new `SynchrotronMonitor` device.
A dummy plan that does nothing but print once a second is added; an endpoint is also added so that it can be run as `curl -X PUT http://127.0.0.1:5000/dummy/start`. A suspender using the `topup_gate_signal` signal of the `SynchrotronMonitor` is added to the run engine before executing the dummy plan. The plan should be paused whenever this signal is `True`, this sometimes happens, but usually does not. Running the same plan with the same configuration in the `__main__` of `beam_topup_plan.py` gives the same behaviour, but also gives a useful error message:
```
Subscription value callback exception (Signal(name='monitor_topup_gate_signal', parent='monitor', value=True, timestamp=1663862377.5352824))
Traceback (most recent call last):
  File "/dls/science/users/zzg91958/python-artemis/.venv/lib/python3.10/site-packages/ophyd/ophydobj.py", line 474, in inner
    cb(*args, **kwargs)
  File "/dls/science/users/zzg91958/python-artemis/.venv/lib/python3.10/site-packages/bluesky/suspenders.py", line 145, in __call__
    raise RuntimeError("Could not create the ")
RuntimeError: Could not create the 
```

### DIFF
--- a/src/artemis/__main__.py
+++ b/src/artemis/__main__.py
@@ -5,15 +5,18 @@ from dataclasses import dataclass
 from enum import Enum
 from json import JSONDecodeError
 from queue import Queue
-from typing import Optional, Tuple
+from typing import Any, Generator, Optional, Tuple
 
 from bluesky import RunEngine
+from bluesky.suspenders import SuspendBoolHigh
 from dataclasses_json import dataclass_json
 from flask import Flask, request
 from flask_restful import Api, Resource
 
+from artemis.beam_topup_plan import dummy_plan, get_synchrotron_monitor_from_parameters
 from artemis.fast_grid_scan_plan import get_plan
 from artemis.parameters import FullParameters
+from artemis.topup_parameters import TopupParameters
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


evalott100:
The above problem Mathew is encountering is from /bluesky/suspenders.py, 

`    def __make_event(self):
        """Make or return the asyncio.Event to use as a bridge."""
        assert self._lock.locked()
        if self._ev is None and self.RE is not None:
            th_ev = threading.Event()

            def really_make_the_event():
                self._ev = asyncio.Event()
                th_ev.set()

            h = self.RE.loop.call_soon_threadsafe(really_make_the_event)
            if not th_ev.wait(0.1):
                h.cancel()
            return self._ev`
            
The wait time being 0.1 causes self._ev to not be set (it seems to not be enough time). Changing this to 1 seconds lets the suspender work fine.

evalott100:
Note: the problem persists with other suspenders - e.g `SuspendFloor`.

dperl-dls:
closing as stale but feel free to reopen

PR #229, opened 2022-09-29
129 add metrics

### LINKED ISSUE #129: Publish metrics about time taken for each step in a plan
As a scientist it will be very useful to see visually when specific operations are taking place so that we can easily speed up operations. Note that some operations are going to happen at the same time and so a gantt chant kind of view will probably be helpful. We should make sure to use standard tools used e.g. Prometheus/Grafana probably.

abbiemery:
Fixes #129 .

Made an early PR with the basic setup of getting a timing metric to a local prometheus instance. Will refine the prometheus build but mainly there needs to be a conversation about what metrics we actually want and where.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -83,11 +83,11 @@
         },
         "bluesky": {
             "hashes": [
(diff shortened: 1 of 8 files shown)


dperl-dls:
Superseded by #382

PR #233, opened 2022-09-30
Fix existing folder check in logging

### LINKED ISSUE #232: Artemis fails to start if log folder does not exist
During testing we found that the `bluesky` folder not existing in the log folder produced an error in Artemis.

## Acceptance Criteria
* If the log folder does not exist Artemis creates it

abbiemery:
Fixes #232


DominicOram (approved):
Looks good to me

Issue #200, opened 2022-07-29
Pull ophyd_devices into a new repo


DominicOram:
To allow for ophyd devices to be used across beamlines we should have them in a separate repo

DominicOram:
Myself, @thomascobb and @graeme-winter, decided this will look something like:
```python
# dodal/common.py
from ophyd_epics_devices import motor

class Dcm:
    def __init__(pv_prefix: str, name: str):
        self.name = name
        self.energy = motor.motor(pv_prefix + ":ENERGY", name=name + ".energy")
        self.pitch = motor.motor(pv_prefix + ":PITCH", name=name + ".pitch")


# dodal/i03.py
from ophyd.core import Connector

def dcm(name="dcm"):
    return Dcm("BL03I-OP-DCM-01", name=name)

def s1(name="s1"):
    return ApertureSlits("BL03I-AL-SLITS-01")

def s2(name="s2"):
    return BladeSlits("BL03I-AL-SLITS-02")


# dodal/test/test_i03
from dodal import i03

def test_pvs_connect():
    with i03.Connector(timeout=10, sim_mode=True):
        dcm = i03.dcm()
        s1 = i03.s1()


# startup.py
from dodal import i03 as ixx
# Or using BEAMLINE
from dodal import ixx

with ixx.Connector() as conn:
    dcm = ixx.dcm()
    # etc.
    conn.connect()

    s1 = ixx.s1()
    conn.connect()
```

DominicOram:
See https://github.com/DiamondLightSource/python-dodal/issues/2

DominicOram:
We should make sure we do #221 at the same time

DominicOram:
Closed by https://github.com/DiamondLightSource/python-artemis/pull/517

PR #249, opened 2022-10-05
Created a devices for FGSCompositeDevice and EigerDevice that are ini…

evalott100:
…tialised during the first get_plan, and updated in subsequent get_plan calls, rather than reinitialized. #221


DominicOram (commented):
In hindsight I think this might be best done as part of https://github.com/DiamondLightSource/python-artemis/issues/200, sorry. Suggesting closing this w/o merging and I have added a comment to #200

PR #265, opened 2022-10-10
(#235) Implement flyable properly

### LINKED ISSUE #235: FastGridScan should properly implement flyable
On the latest version of ophyd FastGridScan cannot be "flyed" as it does not implement all the flyable interface (e.g. collect and describe collect). Currently we don't provide any data when the flyer is kicked off, so an empty implementation should suffice but we will ultimately want to provide data.

DominicOram:
Fixes #235 

To test:
* Confirm new unit test fails on main but passes on this branch

### DIFF
--- a/src/artemis/devices/fast_grid_scan.py
+++ b/src/artemis/devices/fast_grid_scan.py
@@ -254,6 +254,12 @@ class FastGridScan(Device):
     def complete(self) -> DeviceStatus:
         return GridScanCompleteStatus(self)
 
+    def collect(self):
+        return {}
+
+    def describe_collect(self):
+        return {}
+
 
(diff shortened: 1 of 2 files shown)


dperl-dls:
This test passes in main for me

dperl-dls:
> Hmm.. Which version of `bluesky` have you got running?

1.8.3, as specified in pipfile.lock in main

DominicOram:
Apologies, try testing against the latest bluesky version (which we're now using after we removed the lock file)

dperl-dls (approved):
Yes, now I get an error `FakeFastGridScan(...) does not implement all Flyable methods` on main but no such error on this branch

PR #270, opened 2022-10-11
240 ispyb comment

### LINKED ISSUE #240: Fix comment in ispyb
When storing data in ispyb we store the comment passed through by GDA. This means that the comment is the same for both collections, which is incorrect.

## Acceptance Criteria
* Artemis itself creates the comment
* GDA no longer passes it through

dperl-dls:
Fixes #240 from the artemis side by composing the comment in store_in_ispyb

### DIFF
--- a/src/artemis/ispyb/store_in_ispyb.py
+++ b/src/artemis/ispyb/store_in_ispyb.py
@@ -147,8 +147,10 @@ class StoreInIspyb(ABC):
         )
         params["comments"] = (
             "Artemis: Xray centring - Diffraction grid scan of "
-            f"{self.full_params.grid_scan_params.x_step_size} by "
-            f"{self.full_params.grid_scan_params.y_step_size} images. "
+            f"{self.full_params.grid_scan_params.x_steps} by "
+            f"{self.full_params.grid_scan_params.y_steps} images in "
+            f"{self.full_params.grid_scan_params.x_step_size} mm by "
+            f"{self.full_params.grid_scan_params.y_step_size} mm steps. "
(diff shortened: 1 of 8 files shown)


src/artemis/ispyb/store_in_ispyb.py
```diff
@@ -135,7 +136,24 @@ def _store_data_collection_table(self, data_collection_group_id: int) -> int:
...
+        )
+        params["comments"] = (
+            "Artemis: Xray centring - Diffraction grid scan of "
+            f"{self.full_params.grid_scan_params.x_steps} by "
+            f"{self.full_params.grid_scan_params.y_steps} images in "
+            f"{self.full_params.grid_scan_params.x_step_size} mm by "
+            f"{self.full_params.grid_scan_params.y_step_size} mm steps. "
+            f"Top left: [{self.upper_left.x},{self.upper_left.y}], "
+            f"bottom right: [{bottom_right.x},{bottom_right.y}]."
+        )
```
DominicOram, line 156: Nit: Can you put this into a function for neatness e.g. something like `params["comments"] = self.construct_comment()`

DominicOram (approved):
Great, thank you. Minor comment in code

DominicOram:
Confirmed the tests worked locally, I think the CI is just having ongoing issues so we can merge w/o it passing all checks

src/artemis/ispyb/store_in_ispyb.py
dperl-dls: Yes, good idea, done

DominicOram (approved):
Great work, thank you!

PR #278, opened 2022-10-14
258 migrate to dls python3 pip skeleton

evalott100:
Set up virtual environment for using pip instead of pipenv. The dls_dev_env.sh script now sets this up. CI has also been altered to use the new environment. 

@dperl-dls set up a docker container and the relevant CI.


dperl-dls (approved):
works for me too, doesn't change any artemis code

DominicOram:
There are some issues with this:
* `cov.xml` is now in the codebase, it shouldn't be as it's a derived file, we should add to gitignore
* The README still references `pipenv` is this correct? Can it be updated to the recommended way of running Artemis now?
* There's commented out code for doing docs builds, we shoudl either remove it if we're not doing doc builds or add a ticket to reinstate them
* We've lost the `.vscode/` files, is this deliberate?
* Why do we need a `do_nothing` function in `test_log`?

PR #279, opened 2022-10-18
88 fix nonetype errors from dataclasses json

evalott100:


src/artemis/devices/eiger.py
```diff
@@ -159,7 +158,7 @@ def arm_detector(self):
         odin_status &= await_value(self.odin.meta.ready, 1)
         odin_status.wait(10)
 
-        set_and_wait(self.cam.acquire, 1, timeout=10)
+        self.cam.acquire.set(1).wait(timeout=10)
```
DominicOram, line 161: Good catch!

DominicOram (changes requested):
Thanks to looking into this, it does remove the warnings. However, couple of bits:
* I don't think we should be setting these params to optional, in production they are always required to be set and having them optional implies that's not the case
* You have some code in here for the issue around swapping `image_size`, can you remove it from here and make sure it's just on the correct branch

evalott100:
> I don't think we should be setting these params to optional, in production they are always required to be set and having them optional implies that's not the case

I agree, I think it was a quirk of the typing. I'll rewrite it to make it non-optional.

> You have some code in here for the issue around swapping image_size, can you remove it from here and make sure it's just on the correct branch

My bad, I'll start double checking this

DominicOram (approved):
Thanks, this reduces the warning significantly.

Issue #239, opened 2022-09-30
Image size is inverted


DominicOram:
During last round of Artemis testing we found that the image size we were writing to Nexus file had the width/height inverted. We should correct this

DominicOram:
There is a hotfix on the live version (`/dls_sw/i03/software/artemis`) that fixes this. Apply these correcteions and add/fix unit tests

dperl-dls:
fixed by #280

PR #280, opened 2022-10-18
#239 made sure there wasn't an inversion, wrote a unit test to check …

evalott100:
…constants set appropriately to the NexusWriter

### DIFF
--- a/src/artemis/nexus_writing/write_nexus.py
+++ b/src/artemis/nexus_writing/write_nexus.py
@@ -162,7 +162,7 @@ def create_detector_parameters(detector_params: DetectorParams) -> Dict:
         "flatfield_applied": "_dectris/flatfield_correction_applied",
         "pixel_mask": "mask",
         "pixel_mask_applied": "_dectris/pixel_mask_applied",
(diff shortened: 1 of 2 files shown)


src/artemis/nexus_writing/write_nexus.py
```diff
@@ -162,7 +162,7 @@ def create_detector_parameters(detector_params: DetectorParams) -> Dict:
         "flatfield_applied": "_dectris/flatfield_correction_applied",
         "pixel_mask": "mask",
         "pixel_mask_applied": "_dectris/pixel_mask_applied",
-        "image_size": [detector_pixels.width, detector_pixels.height],  # (fast, slow)
+        "image_size": [detector_pixels.height, detector_pixels.width],  # (fast, slow)
```
DominicOram, line 165: Should: This comment is now wrong, it should be  `# (slow, fast)` to match the `fast_axis`, `slow_axis` above.

DominicOram (changes requested):
Looks good to me, minor comment in code, otherwise good, thanks!

evalott100 (commented):
Changes made

DominicOram (approved):
Great, thanks!

Issue #236, opened 2022-09-30
Incorrectly checking filename has been written to


DominicOram:
Currently Artemis checks that we have set the filename in Odin correctly by reading `self.odin.file_writer.id` and `self.odin.meta.file_name` and comparing them to `file_prefix`. However, the PVs we read from are waveforms and have not been converted to strings so the comparison does not work.

# Acceptance Criteria
* The comparison for these PVs works
* There is a test for this

PR #281, opened 2022-10-18
Added string=True and wrote unit test for await_true #236

evalott100:


src/artemis/devices/unit_tests/test_eiger.py
```diff
@@ -234,3 +234,11 @@ def test_stage_runs_successfully(mock_await, fake_eiger):
...
     fake_eiger.odin.file_writer.file_path.put(True)
     fake_eiger.stage()
+
+
+def test_check_filename_has_been_written_to(fake_eiger):
+    from artemis.devices.status import await_value
+
+    assert await_value(
+        fake_eiger.odin.meta.file_name, fake_eiger.detector_params.full_filename
+    )
```
DominicOram, line 244: Should: This test doesn't do what you think it does. `await_value` creates a status object that will be complete when the value is correct. You can call `status.done()` on it to see if it completes. In this case it isn't completing at all because that Signal isn't being set anywhere. The assert is just checking that the object has been created. I think testing this change will actually be hard with a unit test. However, I think that the system test does catch it so happy not to remove this unit test entirely

DominicOram (changes requested):
Great, thanks. Minor comment

evalott100:
Fixes #236

DominicOram (approved):

Issue #155, opened 2022-06-28
Move FGS into smargon


DominicOram:
As of https://github.com/DiamondLightSource/python-artemis/pull/112 we have both a `Smargon` device and an `FastGridScan` device. We should move the `FastGridScan` one to be a sub device of the `Smargon` to match the IOC

evalott100:
I've done this by having fast_grid_scan as a variable in I03Smargon.

We should rethink where the I03Smargon class should go - using it in motors.py is problematic as the is_valid function in fast_grid_scan.py uses the XYZLimitBundle class as a type for the limits positional variable, which can't be imported from motors.py without a circular import error.

dperl-dls:
Closing as stale - should be reopened in Dodal repository if still relevant

PR #283, opened 2022-10-20
Put the FastGridScan device in I03Smargon. 

evalott100:
We should rethink where the I03Smargon class should go - using it in motors.py is problematic as the is_valid function in fast_grid_scan.py uses the XYZLimitBundle class as a type for the limits it imports, which can't be imported from motors.py without a circular import error. #155

### DIFF
--- a/src/artemis/devices/fast_grid_scan.py
+++ b/src/artemis/devices/fast_grid_scan.py
@@ -16,7 +16,7 @@ from ophyd import (
 from ophyd.status import DeviceStatus, StatusBase
 from ophyd.utils.epics_pvs import set_and_wait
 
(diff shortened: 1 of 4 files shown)


DominicOram (changes requested):
Looks good, I think we can solve the circular import by just putting the smargon in it's own file.

DominicOram:
Once https://github.com/DiamondLightSource/python-artemis/pull/371 is merged this will be a bit easier

evalott100:
@DominicOram I don't know if it's as simple as the above commit. Do I need to change references to fast grid scan? Would be good to talk about.

evalott100:
Closing due to the branch being too out of date with main. The new pull request is #455

Issue #255, opened 2022-10-06
Beam converter uses test file in production


DominicOram:
During testing we sound that the beam converter is always using the test lookup table, it should be using the path specified from GDA.

# Acceptance Criteria
* The path from GDA should be used
* There should be no default behavior if the table is not specified
* There are test for this

# Note
* Original work done in https://github.com/DiamondLightSource/python-artemis/pull/216

DominicOram:
Patched on to live version is a dirty fix

DominicOram:
Fixed in https://github.com/DiamondLightSource/python-artemis/pull/284

PR #284, opened 2022-10-20
Ensured proper test file can be passed into beam converter #255

evalott100:


DominicOram (changes requested):
Looks good, should fix the issue but some minor comments.

DominicOram (approved):
Great, thank you, test is much better!

PR #288, opened 2022-10-20
287: Tidy up the eiger

### LINKED ISSUE #287: Critically analyse every PV we set on the eiger
Artemis currently sets forward stream to true with no justification other than doing what GDA does. We should be more careful around what we're setting and why, not just do it because GDA does, particularly in the eiger/odin where there's a lot of PVs set

# Acceptance Criteria
* Artemis doesn't set the forward stream
* We critically analyse the PVs set and document with comments why they're required

DominicOram:
Fixes #287 

To test:
* Confirm on the original issue that all queries have been resolved
* Confirm tests still pass


dperl-dls (approved):
- confirm tests pass
- confirm that all the changes here address points raised in #287
- I can't really say that we critically examined *every* PV

Nit: would like to see the units included in set_detector_threshold docstring, there's a big difference between eV and keV

DominicOram:
> * confirm tests pass
> 
>     * confirm that all the changes here address points raised in #287
> 
>     * I can't really say that we critically examined _every_ PV
> 
> 
> Nit: would like to see the units included in set_detector_threshold docstring, there's a big difference between eV and keV

Thanks! Yes, I agree we haven't examined every PV. I guess that I mean, given a reasonable level of knowledge it is clear to me from looking at the code why we are setting every PV. But that's really subjective so if you think there are things that are unclear we should definitely change names/put in more comments.

I agree with the units, have added it (note I got this from the units of the PV so we're assuming that's correct)

PR #289, opened 2022-10-20
#275: Disable ROI after checking successful data collection

### LINKED ISSUE #257: beam centre/resolution display is wrong in ispyb
Not sure if this is an artemis issue or an ispyb issue but the error is only there for images from grid scans, images from normal data collection are not affected.

Screenshot from 2022-10-07 11-01-00
Screenshot from 2022-10-07 11-01-41

Centre of the image used for calculating resolution displayed in synchweb is off (see screenshots).

Not off by the same amount as #253, it's diagonally towards the upper left instead of only in x.

### LINKED ISSUE #275: Odin issues on unstage
For collection https://ispyb.diamond.ac.uk/dc/visit/mx31440-3/dcg/8142853 we appeared to take data correctly but we had a filewriter timeout on unstaging that caused us to stop collection.

DominicOram:
Fixes #275 

To test:
* Confirm resolution on the ticket sounds sensible and below does as suggested

### DIFF
--- a/src/artemis/devices/eiger.py
+++ b/src/artemis/devices/eiger.py
@@ -74,8 +74,8 @@ class EigerDetector(Device):
         self.odin.file_writer.timeout.put(1)
         self.odin.nodes.wait_for_filewriters_to_finish()
         self.disarm_detector()
-        self.disable_roi_mode()
         status_ok = self.odin.check_odin_state()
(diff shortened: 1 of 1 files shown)


dperl-dls (approved):
confirm that disabling roi mode is now after checking success

PR #290, opened 2022-10-20
(#286) Make wrong filename PVs read only

### LINKED ISSUE #286: Set file unwritable filenames in eiger to RO
It has cropped up a few times in GDA and Artemis where we're setting the wrong filenames in odin (there is one filename PV that we should set and the rest should propagate). To avoid future mistakes we should set the ones we don't want to set to RO with a comment as to why.

DominicOram:
Fixes #286 

To test:
* Confirm unit tests pass
* Try setting the wrong PV (e.g. add `self.odin.meta.file_name.put(file_prefix)` to `set_odin_pvs`) and confirm unit tests would then fail


dperl-dls (approved):
confirm trying to set RO PVs breaks things

PR #291, opened 2022-10-21
273 graylog running twice

evalott100:
Solved by changing the port for logging to graylog.

### DIFF
--- a/src/artemis/devices/eiger.py
+++ b/src/artemis/devices/eiger.py
@@ -53,7 +53,13 @@ class EigerDetector(Device):
         if errors:
             raise Exception("\n".join(errors))
 
-    def stage(self):
+    def pre_arm(self):
+        """
+        Setting up for the arm, to be ran concurrently.
+
+        Args: None
+        Returns: None
+        """
         self.odin.nodes.clear_odin_errors()
         status_ok, error_message = self.odin.check_odin_initialised()
         if not status_ok:
@@ -68,6 +74,7 @@ class EigerDetector(Device):
 
         odin_filename_status.wait(self.ODIN_FILEWRITER_TIMEOUT)
 
+    def stage(self):
         self.arm_detector()
 
     def unstage(self) -> bool:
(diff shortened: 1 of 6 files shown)


evalott100:
More on this:

This is the graylog input for the artemis port we were using:
image

This is the one for the normal gelf which doesn't double send logs:
image

DominicOram (changes requested):
Good spot but I think we need to dig a bit more:
* I want a separate input for Artemis, it makes it easier to filter and to throttle just Artemis if required
* If we don't know why it happened it could just start happening again

Additionally, there's a lot of code changes here that seem unrelated to the graylog stuff, can we remove them?

evalott100:
> I want a separate input for Artemis, it makes it easier to filter and to throttle just Artemis if required
This would make sense, my plan would be to change the config of the artemis port so those two other parameters match the default gelf one. If the artemis port is in use during then this would cause problems, so it'd  need to be during shutdown.

> If we don't know why it happened it could just start happening again
I pretty much agree. If changing those parameters doesn't work then we'll know that it is some unique issue with the port that isn't being shown to us. At that point I'll find a sysadmin and get  them to investigate.

> there's a lot of code changes here that seem unrelated to the graylog stuff, can we remove them? 
Yea I think I accidentally added them. I'll recommit without the port change since we want to get our original port working anyway.

evalott100:
Changing the Artemis config on graylog didn't work. I'll need to ask scicomp about this.

DominicOram:
> Changing the Artemis config on graylog didn't work. I'll need to ask scicomp about this.

Thanks, we should think of a good way to track when we're waiting on something else. Could you add a new column to the project board for `Blocked`, put the issue in it then add a comment on to the original issue about what we're waiting for

DominicOram:
This PR also contains completely unrelated code changes, can we remove them please?

evalott100:
This was caused by the setting "Remove matches from 'All messages' stream" in the steam not being selected. Now it is graylog is only logging once.

PR #292, opened 2022-10-21
164 use documents for processing

dperl-dls:


src/artemis/fgs_communicator.py
```diff
@@ -0,0 +1,68 @@
...
+    Listens to events from the RE and submits:
+    - nothing so far
+    """
+
+    def __init__(self):
+        self.reset(FullParameters())
+
+    def reset(self, parameters):
+        self.active_uid = None
+        self.params = parameters
```
DominicOram, line 28: Nit: technically we don't need to store the params as we only use them here.

src/artemis/tests/test_fgs_communicator.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+
+from artemis import fgs_communicator
+from artemis.parameters import FullParameters
+
```
DominicOram, line 5: Should: These tests are a great start, thank you! I think there's a few more we can add to make it more comprehensive though:
* We can patch out the communicator nexus writers by doing `@patch("fgs_communicator.NexusWriter` and assert that correct things are done on start/stop (we should probably do this even if we don't do the asserts as currently it means the tests are doing actual file writing)
* I think a test where we have a dummy plan that just starts/stops a run then having the FGSCommunicator subscribed to a RE that runs the plan and asserting that the nexus writer mocks are called correctly would be great. This is a bit more than a unit test I guess but proves the whole stack works
* A test that when the nexus writing throws an exception the RE doesn't continue and is in an error state that we can read out to send to GDA. (I know this is more functionality of Bluesky itself but by adding an explicit test for it you're saying it's important functionality that we have thought about)

src/artemis/tests/test_main_system.py
```diff
@@ -38,6 +38,9 @@ def abort(self):
                 raise Exception(self.error)
         self.RE_takes_time = False
 
+    def subscribe(self, *args: Any):
```
DominicOram, line 41: Should: Can you add a test somehwere in this file that an FGSCommunicator gets passed into this at some point

DominicOram (changes requested):
Great refactor, thank you! I think to avoid `main` having in progress work on it we should remove the parts that hint at ispyb/zocalo in this PR and just have them in https://github.com/DiamondLightSource/python-artemis/pull/295, comments reflect this. 

I'm also a bit torn about the `testing_plan`. While we can't run the full thing in dev easily I see why it's useful but I don't think it will ever be that useful in production so I'm not keen on adding it into `__main__.py`. Instead I think it should just have it's own `__main__` in the same file that we can use to run it. e.g. like the `snapshot_plan`

dperl-dls:
closed in favour of #295

PR #295, opened 2022-10-24
164 move ispyb and zocalo to communicator

dperl-dls:


src/artemis/fast_grid_scan_plan.py
```diff
@@ -75,36 +80,27 @@ def do_fgs():
...
+    communicator: FGSCommunicator,
+):
+    @subs_decorator(communicator)
+    def decorated():
+        yield from run_gridscan(fgs_composite, eiger, parameters)
+        yield from move_xyz(
+            fgs_composite.sample_motors, communicator.xray_centre_motor_position
+        )
 
+    yield from decorated()
```
DominicOram, line 100: Should: As discussed I think something a little more like:
```python
    yield from bps.subscribe("all", communicator)
    yield from run_gridscan(fgs_composite, eiger, parameters)
    yield from bps.unsubscribe(communicator)
    
    communicator.wait_for_results()
    yield from move_xyz(
        fgs_composite.sample_motors, communicator.xray_centre_motor_position
    )
```
Gets round having the communicator have to check what plan is running and is more explicit in when results are being waited on.

src/artemis/fgs_communicator.py
```diff
@@ -0,0 +1,144 @@
...
+from artemis.nexus_writing.write_nexus import (
+    NexusWriter,
+    create_parameters_for_first_file,
+    create_parameters_for_second_file,
+)
+from artemis.parameters import FullParameters
+from artemis.zocalo_interaction import run_end, run_start, wait_for_result
+
+
+class FGSCommunicator(CallbackBase):
```
DominicOram, line 17: Should: As discussed i think breaking these up into the different communicators would be more reusable. See comment at https://github.com/DiamondLightSource/python-artemis/issues/164#issuecomment-1262072452. It probably then makes sense for each communicator to be in the appropriate file e.g. the zocalo one in `zocalo_interaction`

src/artemis/fgs_communicator.py
```diff
@@ -0,0 +1,144 @@
...
+            raise ValueError(
+                "Non-dict object stored in descriptor list, or key not valid."
+            )
+        return associated_descriptor_doc
+
+    # TODO is this going to eat too much memory if there are a lot of these with a lot of data?
+    # It probably gets destroyed on self.reset() ?
+    def descriptor(self, doc: dict):
+        artemis.log.LOGGER.debug(f"\n\nReceived descriptor document: {doc}\n")
+        self.descriptors[doc.get("uid")] = doc
```
DominicOram, line 77: As discussed, we may not need this anymore if we no longer need a reference to the plan. Let's see if it falls out in refactoring

src/artemis/ispyb/store_in_ispyb.py
```diff
@@ -209,7 +216,12 @@ def _store_position_table(self, dc_id: int) -> int:
         return self.mx_acquisition.update_dc_position(list(params.values()))
 
     def _store_data_collection_group_table(self) -> int:
-        session_id = self.core.retrieve_visit_id(self.get_visit_string())
+        try:
+            session_id = self.core.retrieve_visit_id(self.get_visit_string())
+        except ispyb.NoResult:
+            raise Exception(
+                f"Not found - session ID for visit {self.get_visit_string()} where self.ispyb_params.visit_path is {self.ispyb_params.visit_path}"
+            )
```
DominicOram, line 224: :+1: Nice, thanks!

src/artemis/ispyb/tests/test_store_in_ispyb.py
```diff
@@ -220,9 +220,13 @@ def test_exception_during_run_results_in_bad_run_status(
...
-    with pytest.raises(Exception) as _:
-        with dummy_ispyb:
-            raise Exception
+
+    dummy_ispyb.begin_deposition()
+    dummy_ispyb.end_deposition("fail")
+
+    # with pytest.raises(Exception) as _:
+    #     with dummy_ispyb:
+    #         raise Exception
```
DominicOram, line 229: Should: Remove commented out code

test-xrc-result.json
```diff
@@ -0,0 +1,13 @@
+{
```
DominicOram, line 1: Should: Was this mistakenly added?

DominicOram (changes requested):
Really good work, thank you! As discussed I think some refactoring would be good. Also, would you be able to go through fast_grid_scan_plan and add some more comments (e.g. why do I need to subscribe something or unsubscribe it after) and docstrings on the individual plans? I think this is the most "user-facing" code we have so we should go through it as if we're a user and see if it makes sense

setup.cfg
```diff
@@ -21,6 +21,8 @@ install_requires =
     pyepics
     flask-restful
     dataclasses-json
+    mysql-connector-python<=8.0.29 #See https://github.com/DiamondLightSource/ispyb-api/issues/183
```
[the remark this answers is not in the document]
dperl-dls: I think so? I get that utf-8 error with the newest version of ispyb-api so...

src/artemis/fgs_communicator.py
dperl-dls: I agree it makes sense to do this, but I think we should discuss a bit more about how to organise it so I'd rather do that in a separate issue/PR

src/artemis/devices/motors.py
```diff
@@ -43,7 +43,7 @@ class I03Smargon(MotorBundle):
     x: EpicsMotor = Component(EpicsMotor, "X")
     y: EpicsMotor = Component(EpicsMotor, "Y")
     z: EpicsMotor = Component(EpicsMotor, "Z")
-    omega: EpicsMotor = Component(EpicsMotor, "OMEGA")
+    omega: EpicsMotor = Component(EpicsMotor, "Omega")
```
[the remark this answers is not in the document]
dperl-dls: fixed

src/artemis/fgs_communicator.py
```diff
@@ -0,0 +1,144 @@
...
+        self.ispyb = (
+            StoreInIspyb3D(ispyb_config, self.params)
+            if self.params.grid_scan_params.is_3d_grid_scan
+            else StoreInIspyb2D(ispyb_config, self.params)
+        )
+        self.active_uid = None
+        self.gridscan_uid = None
+        # TODO add flag for this or delete?
+        # useful for testing, to not duplicate files:
+        # self.params.detector_params.prefix += str(time.time())
```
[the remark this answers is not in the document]
dperl-dls: deleted

src/artemis/fgs_communicator.py
dperl-dls: yes, it's much cleaner now fgs_communicator doesn't have to know which plan it's attached to

src/artemis/ispyb/tests/test_store_in_ispyb.py
dperl-dls: removed

setup.cfg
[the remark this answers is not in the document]
DominicOram: @dperl-dls, you marked this as resolved but it's not clear why. Is my comment above incorrect?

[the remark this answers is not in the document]
dperl-dls: sorry, I thought I removed the pin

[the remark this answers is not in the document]
dperl-dls: fixed now

src/artemis/fgs_communicator.py
```diff
@@ -0,0 +1,114 @@
...
+    def start(self, doc: dict):
+        artemis.log.LOGGER.debug(f"\n\nReceived start document:\n\n {doc}\n")
+        artemis.log.LOGGER.info("Creating Nexus files.")
+        self.nxs_writer_1.create_nexus_file()
+        self.nxs_writer_2.create_nexus_file()
+
+    def event(self, doc: dict):
+        artemis.log.LOGGER.debug(f"\n\nReceived event document:\n{doc}\n")
+
+        if doc.get("name") == "ispyb_motor_positions":
```
DominicOram, line 61: Should: I don't like the hardcoded string here, it means if we change the name of the doc in the plan we have to change it here or else it will break. I think it would be better if they were both looking at a constant stored elsewhere so we could ensure their always the same.

src/artemis/tests/test_fast_grid_scan_plan.py
```diff
@@ -60,136 +64,126 @@ def test_ispyb_params_update_from_ophyd_devices_correctly():
...
+        Point3D(1, 2, 3)
+    )
+    FakeComposite = make_fake_device(FGSComposite)
+    RE(
+        move_xyz(
+            FakeComposite("test", name="fakecomposite").sample_motors, motor_position
+        )
+    )
+    x = bps_mv
+    print(x)
```
DominicOram, line 138: Nit: Remove print statement and x assignment

src/artemis/tests/test_fast_grid_scan_plan.py
```diff
@@ -60,136 +64,126 @@ def test_ispyb_params_update_from_ophyd_devices_correctly():
...
 
-    mock_ispyb_update_time_and_status.assert_has_calls(
-        [call(DUMMY_TIME_STRING, GOOD_ISPYB_RUN_STATUS, id, dcg_id) for id in dc_ids]
+    FakeEiger = make_fake_device(EigerDetector)
+    eiger: EigerDetector = FakeEiger(
+        detector_params=params.detector_params, name="eiger"
     )
-    assert mock_ispyb_update_time_and_status.call_count == len(dc_ids)
+
+    return fgs_composite, eiger, params
```
DominicOram, line 189: Should: You can remove this fixture as it is no longer used. However, you could also probably use it in some of your tests to remove boilerplate such as making the fake devices. Up to you.

src/artemis/tests/test_fgs_communicator.py
```diff
@@ -0,0 +1,265 @@
...
+
+    dc_ids = [1, 2]
+    dcg_id = 4
+
+    mock_ispyb_store_grid_scan.return_value = [dc_ids, None, dcg_id]
+    mock_ispyb_get_time.return_value = DUMMY_TIME_STRING
+    mock_ispyb_update_time_and_status.return_value = None
+
+    params = FullParameters()
+    params.detector_params.prefix += str(time.time())
```
DominicOram, line 98: Nit: Why is this required? We should probably just make sure the test isn't writing nexus files by mocking things out

src/artemis/tests/test_fgs_communicator.py
```diff
@@ -0,0 +1,265 @@
...
+    "artemis.fgs_communicator.StoreInIspyb3D.update_grid_scan_with_end_time_and_status"
+)
+def test_fgs_failing_results_in_bad_run_status_in_ispyb(
+    mock_ispyb_update_time_and_status: MagicMock,
+    mock_ispyb_get_time: MagicMock,
+    mock_ispyb_store_grid_scan: MagicMock,
+    wait_for_result: MagicMock,
+    run_end: MagicMock,
+    run_start: MagicMock,
+    dummy_3d_gridscan_args,
```
DominicOram, line 148: Nit: Do you need `dummy_3d_gridscans_args`, I don't think you're really using it? If you don't you can remove the fixture

src/artemis/tests/test_fgs_communicator.py
```diff
@@ -0,0 +1,265 @@
...
+):
+    dc_ids = [1, 2]
+    dcg_id = 4
+    mock_ispyb_store_grid_scan.return_value = [dc_ids, None, dcg_id]
+    mock_ispyb_get_time.return_value = DUMMY_TIME_STRING
+    mock_ispyb_update_time_and_status.return_value = None
+
+    params = FullParameters()
+    params.detector_params.prefix += str(time.time())
```
DominicOram, line 157: Nit: As before, don't think this is required

DominicOram (changes requested):
Love all the additional logging, some minor comments. Also I think we can do with a couple more tests. You currently have some tests that the plans produce the right docs and that given the right docs the communicator will do as expected. I think it would be good to put the two together in that:
* Mock out the external services
* Call the plan
* Assert the mocks are correctly filled in

This is more a system test than a unit test but I think we need it. I think it's fine to go in the `test_fgs..` file even though it's a system one

DominicOram:
Looks good, mostly minor stuff now. @callumforrester would you be able to take a look at this?

src/artemis/fgs_communicator.py
dperl-dls: done, #315

src/artemis/fgs_communicator.py
```diff
@@ -0,0 +1,114 @@
...
+            self.params.ispyb_params.slit_gap_size_y = doc["data"]["slit_gaps_ygap"]
+
+            artemis.log.LOGGER.info("Creating ispyb entry.")
+            self.ispyb_ids = self.ispyb.begin_deposition()
+            datacollection_ids = self.ispyb_ids[0]
+            self.datacollection_group_id = self.ispyb_ids[2]
+            for id in datacollection_ids:
+                run_start(id)
+
+        # any live update stuff goes here
```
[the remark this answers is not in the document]
dperl-dls: removed

src/artemis/fgs_communicator.py
```diff
@@ -0,0 +1,114 @@
...
+    def stop(self, doc: dict):
+        artemis.log.LOGGER.debug(f"\n\nReceived stop document:\n\n {doc}\n")
+        exit_status = doc.get("exit_status")
+
+        artemis.log.LOGGER.debug("Updating Nexus file timestamps.")
+        self.nxs_writer_1.update_nexus_file_timestamp()
+        self.nxs_writer_2.update_nexus_file_timestamp()
+
+        if self.ispyb_ids == (None, None, None):
+            raise Exception("ispyb was not initialised at run start")
```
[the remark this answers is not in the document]
dperl-dls: I don't think it's possible. We can test sending a stop document, for sure, but bluesky won't do bps.run_end if it isn't in an open run

setup.cfg
```diff
@@ -21,6 +21,8 @@ install_requires =
     pyepics
     flask-restful
     dataclasses-json
+    mysql-connector-python
+	sqlalchemy
```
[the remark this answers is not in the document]
dperl-dls: oh yeah haha

src/artemis/tests/test_fgs_communicator.py
dperl-dls: Okay, I just figured I should mock the minimum possible to be able to run the test, I guess I like the idea that it is calling the real nexus writer and not a fake one. But now that you mention it, since it's not actually what we're testing here I will remove it

src/artemis/tests/test_fast_grid_scan_plan.py
```diff
@@ -60,136 +64,126 @@ def test_ispyb_params_update_from_ophyd_devices_correctly():
     slit_gaps.xgap.sim_put(xgap_test_value)
     slit_gaps.ygap.sim_put(ygap_test_value)
 
-    RE(update_params_from_epics_devices(params, undulator, synchrotron, slit_gaps))
+    class TestCB(CallbackBase):
```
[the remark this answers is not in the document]
dperl-dls: this test is for the read_hardware_for_ispyb plan, not the communicator, that's also why it's in test_fast_grid_scan_plan, it wasn't supposed to test that the documents are handled correctly. I've renamed it to hopefully be more clear, but I don't quite understand what you meant by
>if it fails it may be that you need to update FGSCommunicator

src/artemis/tests/test_fast_grid_scan_plan.py
dperl-dls: done

src/artemis/fgs_communicator.py
```diff
@@ -0,0 +1,114 @@
...
+    Listens to documents emitted by the RE and:
+    - prepares nexus files
+    - prepares ipsyb deposition
+    - submits job to zocalo
+    """
+
+    def __init__(self):
+        self.reset(FullParameters(), write_files=False)
+
+    def reset(self, parameters: FullParameters, write_files=True):
```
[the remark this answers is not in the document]
dperl-dls: For some reason when I first wrote this I thought it was a good idea to have it reused, but on revision I agree that it's better to just create a new one each run, I've removed this

src/artemis/fast_grid_scan_plan.py
```diff
@@ -75,36 +84,31 @@ def do_fgs():
...
+    fgs_composite: FGSComposite,
+    eiger: EigerDetector,
+    parameters: FullParameters,
+    communicator: FGSCommunicator,
+):
+    """A multi-run plan which runs a gridscan, gets the results from zocalo
+    and moves to the centre of mass determined by zocalo"""
+    # our communicator should listen to documents only from the actual grid scan
+    # so we subscribe to it with our plan
+    subs_decorator(communicator)(run_gridscan(fgs_composite, eiger, parameters))
```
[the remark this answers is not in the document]
dperl-dls: not quite sure how I would just context manager to achieve this, but I have just written it out more explicitly so hopefully it's more readable

src/artemis/tests/test_fgs_communicator.py
```diff
@@ -0,0 +1,265 @@
...
+):
+
+    params = FullParameters()
+    params.detector_params.prefix += str(time.time())
+    communicator = fgs_communicator.FGSCommunicator()
+    communicator.reset(params)
+    communicator.start(test_start_document)
+
+    # awkwardly, nxs_writer_1 and nxs_writer_2 are mocked by the exact same object,
+    # so can't test them individually
```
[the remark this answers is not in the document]
dperl-dls: thanks

src/artemis/fgs_communicator.py
dperl-dls: ok, i've made them read from a yml config file which we can put plan names in, but actually do you think it's better to just have them in  a python file instead?

src/artemis/tests/test_fast_grid_scan_plan.py
dperl-dls: ok, removed

src/artemis/__main__.py
```diff
@@ -171,7 +174,9 @@ def cli_arg_parse() -> Tuple[Optional[bool], Optional[str]]:
     app, runner = create_app()
     atexit.register(runner.shutdown)
     flask_thread = threading.Thread(
-        target=lambda: app.run(host="0.0.0.0", debug=True, use_reloader=False),
+        target=lambda: app.run(
+            host="0.0.0.0", port=5005, debug=True, use_reloader=False
```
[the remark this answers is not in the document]
dperl-dls: yes, done

dperl-dls:
> * Mock out the external services
> 
>     * Call the plan
> 
>     * Assert the mocks are correctly filled in
> 
> 
> This is more a system test than a unit test but I think we need it. I think it's fine to go in the `test_fgs..` file even though it's a system one

Unfortunately I think this isn't currently possible without a better functioning S03 - it gets stuck on the eigerodin, scaninvalid, etc. The framework of such a test is there which can be easily tidied up once we have better simulation.

README.md
```diff
@@ -44,27 +44,23 @@ Starting a scan
 
 To start a scan you can do the following:
 ```
-curl -X PUT http://127.0.0.1:5000/fast_grid_scan/start --data-binary "@test_parameters.json" -H "Content-Type: application/json"
+curl -X PUT http://127.0.0.1:5005/fast_grid_scan/start --data-binary "@test_parameters.json" -H "Content-Type: application/json"
```
DominicOram, line 47: Good spot in updating the docs!

src/artemis/devices/motors.py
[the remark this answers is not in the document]
DominicOram: Must: This change will break production, we can't have it in

src/artemis/tests/test_fast_grid_scan_plan.py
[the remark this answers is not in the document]
DominicOram: I guess what I mean is that what we're testing here is that the expected documents are created but it's not clear just from this test why we care that those specific documents are created. We care because `FGSCommunicator` needs them so either using the real `FGSCommunicator` here or having a comment as to why we care about the test would make that clearer. It's fine though, happy to leave as is.

src/artemis/fgs_communicator.py
DominicOram: I think a python file with a constant in is better. With the YAML file we have a similar issue where both files need to know the name of the key. If it's a real constant you can change the name by doing a refactor in the IDE much more easily

DominicOram (approved):
Great, thank you. Had some really minor comments but I just fixed them up myself

DominicOram (approved):
I removed the `wait_for_results` on run end in favour of the explicit wait that we now have.

PR #297, opened 2022-10-24
114 set no steps to 0 at beginning of scan

### LINKED ISSUE #114: Set "Number of Z Steps" to zero at end of 3d grid scan script
The way that the low level grid scan program (on the smargon controller) decides if it is a 2d or 3d grid scan is just (pseudo-code)
```
is_3d_grid_scan = no_of_z_steps > 0
```
So if we do a 3d grid scan, and then attempt a 2d grid scan, we end up with two 3d grid scans, as the number of z steps is still greater than zero. We should put it back to zero at the end of each 3d grid scan.

evalott100:
Setting it at the end of the plan instead. Fixes #114

### DIFF
--- a/src/artemis/fast_grid_scan_plan.py
+++ b/src/artemis/fast_grid_scan_plan.py
@@ -69,6 +69,11 @@ def run_gridscan(
     )
 
     fgs_motors = fgs_composite.fast_grid_scan
+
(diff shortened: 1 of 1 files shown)


DominicOram (approved):
Thanks, I think ideally there would a be a test against this but probably overkill

PR #298, opened 2022-10-24
Added logic to raise exception and log to GDA when zocalo fails to fi…

### LINKED ISSUE #251: Behave correctly when no results returned from zocalo
When no spots are found on the grid zocalo will return a position of [nan, nan, nan]. In this case we should:

* Log to graylog that no position was returned
* Log in GDA no position was returned
* Comment in ISpyb that nothing was found
* Move to the optical centre (i.e. the position Artemis starts in)

Do not raise an error to stop GDA

evalott100:
…nd the centre. The motors return to their default position Fixes #251

### DIFF
--- a/src/artemis/fast_grid_scan_plan.py
+++ b/src/artemis/fast_grid_scan_plan.py
@@ -1,4 +1,5 @@
 import argparse
+import math
 import os
 
 import bluesky.plan_stubs as bps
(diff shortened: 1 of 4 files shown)


DominicOram (changes requested):
I think the structure of this is a bit messy, but also aware the structure that https://github.com/DiamondLightSource/python-artemis/pull/295 is going to change a lot of structure anyway. I think we should merge that first then revisit this.

DominicOram:
Now that https://github.com/DiamondLightSource/python-artemis/pull/295 is merged can you take another look at this @evalott100?

PR #307, opened 2022-10-31
305: Ensure reading PVs as strings when appropriate

### LINKED ISSUE #305: Correct mistaken merge for eiger signals
I believe ttps://github.com/DiamondLightSource/python-artemis/pull/290/files contains a mistaken merge, the signals should be both RO and also interpreted as strings.

DominicOram:
Fixes #305 

To test:
* Confirm this contains the same changes as https://github.com/DiamondLightSource/python-artemis/pull/281


dperl-dls (approved):
oops, sorry I didn't catch that #290 reverted #281

PR #319, opened 2022-11-03
276 write error into ispyb comment

dperl-dls:
Fixes 276

Bluesky handily gives us a `reason` when it emits the `stop` document. If the data collection has failed, this is the message from whatever exception ended the run. Often this message is readable but sometimes it may not be - we should be making sure that exceptions which propagate to the RunEngine have sensible messages.

Based on 164 branch so wait for that one to merge

### To test:
- check tests pass


DominicOram (changes requested):
This PR contains all the changes from #164 and so it's hard to see the relevant parts for review. Could you make a new branch from current main and cherry-pick the relevant commits onto it?

dperl-dls:
Done, #340

PR #334, opened 2022-11-04
318 oav centring create devices for oav set values at start of operation

### LINKED ISSUE #318: OAV Edge Detection: Create Devices for the OAV, camera, smargon. Write a function that sets their values at the start of operation.
Ophyd devices need to be created to hold the relevant PVs as signals. A python method will need to set the device PVs to the correct values before a run, the values will come from OAVCentring.json - found in mx-config GDA.

To calculate the grid positions we need to get the edges of the pin from the OAV image. To do this we need to read/write to a number of PVs on the OAV.

## Acceptance Criteria
* We understand and document what is required to set/read to do edge detection
* We can put the OAV in this mode and get the edges out through ophyd

evalott100:
Fixes #318


src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+            self.oav.input_pv, self.oav_parameters.input_plugin + ".MXSC"
+        )
+
+        # zoom is an integer value, stored as a float in a string so
+        # we may need a .0 on the end? GDA does this with
+        # zoomString = '%1.0dx' % float(zoom)
+        # which seems suspicious, we may want to think about doing this in a nicer way
+        yield from bps.abs_set(
+            self.oav_camera.zoom, f"{float(int(self.oav_parameters.zoom))}x", wait=True
+        )
```
DominicOram, line 153: Should: It's not just an integer, if you run `caget -d DBR_GR_ENUM BL03I-EA-OAV-01:FZOOM:MP:SELECT` you can get all the options for it, which include `2.5x` and `7.5x`. I think we should have a conversion function on the device itself that:
* Takes a requested zoom, compares it to the options (read from the PV)
* If the request is possible return the string to set
* If the request is not possible log a warning and return the closes zoom level

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+
+        blbrightness.moveTo(brightnessToUse)
+        """
+
+    def start_mxsc(self, input_plugin, min_callback_time, filename=None):
+        yield from bps.abs_set(self.oav.input_plugin_pv, input_plugin)
+        yield from bps.abs_set(self.oav.enable_callbacks_pv, 1)
+        yield from bps.abs_set(self.oav.min_callback_time_pv, min_callback_time)
+        yield from bps.abs_set(self.oav.blocking_callbacks_pv, 0)
```
DominicOram, line 174: Should: I think we could do with some comments here on why we're setting these values

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+        https://github.com/DiamondLightSource/python-artemis/issues/317
+        is solved
+         brightnessToUse = self.oav_parameters.extract_dict_parameter(
+            context, "brightness"
+        )
+
+        blbrightness.moveTo(brightnessToUse)
+        """
+
+    def start_mxsc(self, input_plugin, min_callback_time, filename=None):
```
DominicOram, line 170: Should: A docstring on this would be good. Also a comment about why having a filename of None is ok

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+        yield from bps.abs_set(self.oav.input_plugin_pv, input_plugin)
+        yield from bps.abs_set(self.oav.enable_callbacks_pv, 1)
+        yield from bps.abs_set(self.oav.min_callback_time_pv, min_callback_time)
+        yield from bps.abs_set(self.oav.blocking_callbacks_pv, 0)
+
+        # I03-323
```
DominicOram, line 176: Nit: I think this speaks for itself without the link to the issue

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+            yield from bps.abs_set(self.oav.py_filename_pv, filename, wait=True)
+            yield from bps.abs_set(self.oav.read_file_pv, 1)
+
+        # Image annotations
+        yield from bps.abs_set(self.oav.draw_tip_pv, True)
+        yield from bps.abs_set(self.oav.draw_edges_pv, True)
+
+        # Image to send downstream
+        OUTPUT_ORIGINAL = 0
+        yield from bps.abs_set(self.oav.output_array_pv, OUTPUT_ORIGINAL)
```
DominicOram, line 187: Should: I'm not sure what this means and the comment doesn't help much

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+        # zoom is an integer value, stored as a float in a string so
+        # we may need a .0 on the end? GDA does this with
+        # zoomString = '%1.0dx' % float(zoom)
+        # which seems suspicious, we may want to think about doing this in a nicer way
+        yield from bps.abs_set(
+            self.oav_camera.zoom, f"{float(int(self.oav_parameters.zoom))}x", wait=True
+        )
+
+        if (yield from bps.rd(self.oav_backlight.control)) == "Out":
+            yield from bps.abs_set(self.oav_backlight.control, "In", wait=True)
```
DominicOram, line 156: Should: Can we move this to a plan stub in the backlight file with optional wait or not. I also think using the `backlight.py` file gives you nice constanst rather than having to use strings

DominicOram (changes requested):
Really good start, sorry I may have gone a bit overboard with the comments.

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+    def __init__(self):
+        self.rgb_mode = 2
+
+    def load_parameters_from_file(
+        self,
+        context="loopCentring",
+        path=os.path.dirname(os.path.realpath(__file__)),
+    ) -> None:
+        """
+        A very termporary solution to help me do testing. Eventually we'll get
```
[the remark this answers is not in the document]
evalott100: Yep good idea. I meant to do this. I was using this method to play round without writing tests and forgot to take it out.

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+        )
+        self.min_callback_time = self.extract_dict_parameter(
+            context, "min_callback_time", fallback_value=0.08
+        )
+
+        self.direction = self.extract_dict_parameter(context, "direction")
+
+        self.max_tip_distance = self.extract_dict_parameter(context, "max_tip_distance")
+
+    def extract_dict_parameter(self, context: str, key: str, fallback_value=None):
```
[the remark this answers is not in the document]
evalott100: The extract dict parameter? Agreed on the docstring.

I wrote one which tests it has the behaviour we need - pretty much just taking the correct parameters from the desired places within the json - and revertig to the fallback_value on failure.

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+class NewDirection:
+    RIGHT_TO_LEFT = 0
+    BOTTOM_TO_TOP = 1
+    LEFT_TO_RIGHT = 2
+    TOP_TO_BOTTOM = 3
+
+
+class OAVParameters:
+    def __init__(self):
+        self.rgb_mode = 2
```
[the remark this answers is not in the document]
evalott100: One of the parameters, contained as a PV but not in the json, is `oavColourMode`. This has the value `2` for rgbMode. GDA defines this in the function where it sets parameters to make the context more clear - which I think is nice.

I'll put it outside of initialisation but inside the class while we decide what to do with it.

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+class OAVCentring:
+    def __init__(self, beamline="BL03I"):
+        self.oav = OAV(name="oav", prefix=beamline + "-DI-OAV-01:")
+        self.oav_camera = Camera(name="oav-camera", prefix=beamline + "-EA-OAV-01:")
+        self.oav_backlight = Backlight(
+            name="oav-backlight", prefix=beamline + "-MO-MD2-01:"
+        )
+        self.oav_goniometer = Goniometer(name="oav-goniometer", prefix="-MO-SGON-01:")
```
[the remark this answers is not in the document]
evalott100: Good point, it would make sense to use that.

src/artemis/devices/oav/oav_centring.py
evalott100: Tbh I'm not entirely sure. I'll ask Mark Williams.

src/artemis/devices/oav/oav_centring.py
evalott100: Good idea. We'll have to make an ophyd device for the backlight brightness anyways so it will need its own file.

DominicOram: Yh, I think brightness can eventually go in that `backlight.py` file too

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,199 @@
...
+from artemis.devices.oav.oav_detector import OAV, Backlight, Camera, Goniometer
+
+
+class Direction:
+    LEFT_TO_RIGHT = 1
+    RIGHT_TO_LEFT = 2
+    TOP_TO_BOTTOM = 3
+
+
+class NewDirection:
```
DominicOram, line 17: Nit: I think you should be able to use `IntEnum` (https://docs.python.org/3/library/enum.html#enum.IntEnum) for these

src/artemis/devices/oav/oav_centring.py
evalott100: I'm not entirely sure what these values are. I should ask Neil Patterson about them.

src/artemis/devices/oav/oav_centring.py
evalott100: In hindsight I don't think having a filename of none is okay. This is the python file for calculating the edge waveforms, which will always be necessary. It should be a fail-state if we can't find it in the json passed from GDA imo.

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,295 @@
...
+        )
+        self.min_callback_time = self._extract_dict_parameter(
+            "min_callback_time", fallback_value=0.08
+        )
+
+        self.direction = self._extract_dict_parameter("direction")
+
+        self.max_tip_distance = self._extract_dict_parameter("max_tip_distance")
+
+    def _extract_dict_parameter(self, key: str, fallback_value=None, reload_json=False):
```
DominicOram, line 72: Much more readable, thanks!

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,295 @@
...
+
+        # No fallback_value was given and the key wasn't found
+        raise KeyError(
+            f"Searched in {self.file} for key {key} in context {self.context} but no value was found. No fallback value was given."
+        )
+
+    def load_microns_per_pixel(self, zoom):
+        zoom_levels_filename = "src/artemis/oav/microns_for_zoom_levels.json"
+        f = open(zoom_levels_filename)
+        json_dict = json.load(f)
```
DominicOram, line 119: Should: We should have a context manager here to not leave the file open.

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,295 @@
...
+                ):
+                    crosshair_x_line = file_lines[i + 1]
+                    crosshair_y_line = file_lines[i + 2]
+                    break
+
+            if crosshair_x_line is None or crosshair_y_line is None:
+                pass  # TODO throw error
+
+            self.beam_centre_x = int(crosshair_x_line.split(" = ")[1])
+            self.beam_centre_y = int(crosshair_y_line.split(" = ")[1])
```
DominicOram, line 271: Nit: This seems unrealted to the issue we're trying to solve? Can we do it in a separate PR please

src/artemis/devices/oav/oav_detector.py
```diff
@@ -0,0 +1,126 @@
...
+    HDF5Plugin,
+    OverlayPlugin,
+    ProcessPlugin,
+    ROIPlugin,
+)
+
+from artemis.devices.oav.grid_overlay import SnapshotWithGrid
+
+
+class ColorMode(IntEnum):
```
DominicOram, line 19: :+1:

src/artemis/devices/oav/oav_centring.py
```diff
@@ -0,0 +1,295 @@
...
+
+        if fallback_value:
+            return fallback_value
+
+        # No fallback_value was given and the key wasn't found
+        raise KeyError(
+            f"Searched in {self.file} for key {key} in context {self.context} but no value was found. No fallback value was given."
+        )
+
+    def load_microns_per_pixel(self, zoom):
```
DominicOram, line 116: Should: Can we have a few unit tests on this function?

DominicOram (changes requested):
Thanks for addressing those issues. There are a few others I think you missed on the first time round which are still relevant and I added one or two more.

src/artemis/devices/oav/oav_centring.py
evalott100: It will need to be an integer: I'm fairly sure it currently is from I03 - if it weren't a compatible pixels to microns ration wouldn't be able to be extracted from the JSON.

DominicOram (approved):
This looks good, thank you!

PR #336, opened 2022-11-04
#312 a very basic fake zocalo

### LINKED ISSUE #312: Create a fake zocalo service for artemis testing
As part of work needed for #311 

- I think that for simplicity this should be a simple rabbitmq server app which consumes messages from artemis and sends back fake processing results, because this will be easiest for us to add more types of requests to for future experiment types (e.g., omega scans...) and it will not use a lot of computing resources (my workstation is already very low on memory running artemis + S03 + graylog...)
- Alternatively, this could be a copy of the real zocalo with fake 'recipes' instead, which would be more realistic, but more work to maintain

dperl-dls:
- consumes 'end_run()' message
- sends back "results" for dcgid 4

Fixes #312

### To test:
1. Run `module load rabbitmq/dev` and fake zocalo main.py in one terminal
2. Run `module load dials/latest`, set artemis zocalo environment to `devrmq` and end run_end() and wait_for_result() from a python shell in another terminal and see that the request is "processed" and a "result" is returned.

### DIFF
--- a/fake_zocalo/README.rst
+++ b/fake_zocalo/README.rst
@@ -1,0 +1,11 @@
+fake_zocalo
+===========================
+
+.. note::
+
+    This is meant to be used for testing artemis. Don't try to process any real
+    data with it! You will just get back (1.2, 2.3, 3.4).
+
+## To run:
+- You first need to run `module load rabbitmq/dev`, which starts the rabbitmq server and generates some credentials in ~/.fake_zocalo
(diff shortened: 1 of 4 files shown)


DominicOram (approved):
This is great! We're getting close to being able to run the full stack properly, which is very cool. I added some helper scripts and a system test that we can use once we've done https://github.com/DiamondLightSource/python-artemis/issues/356. Otherwise all good.

PR #340, opened 2022-11-08
276 new put error in ispyb

### LINKED ISSUE #276: Comment on ispyb when zocalo fails to find centre
In #251, the logic on the grid scan was changed so that when zocalo fails to find the centre, logs are made to GDA and graylog,  and an Exception is raised. This is contained in `src/artemis/fast_grid_scan_plan.py` in `run_gridscan`. We also need to send out a comment to ispyb.

dperl-dls:
Fixes #276 

replaces #319 

Bluesky handily gives us a reason when it emits the stop document. If the data collection has failed, this is the message from whatever exception ended the run. Often this message is readable but sometimes it may not be - we should be making sure that exceptions which propagate to the RunEngine have sensible messages.

### To test:
check tests pass


src/artemis/ispyb/store_in_ispyb.py
```diff
@@ -47,7 +47,7 @@ def begin_deposition(self):
         ) = self.store_grid_scan(self.full_params)
         return self.datacollection_ids, self.grid_ids, self.datacollection_group_id
 
-    def end_deposition(self, success):
+    def end_deposition(self, success, reason):
```
DominicOram, line 50: Nit: This might be a bit of overkill but I think it would be nice to create some kind of `ISpyBStatus` enum here, a bit like we do in [main](https://github.com/DiamondLightSource/python-artemis/blob/59739c26fdd3a99c853a8b96f3804cd945c5c9b1/src/artemis/__main__.py#L43) along with a helper function of `bluesky_doc_to_ispyb_status`. This has a couple of advantages:
* Pulls the magic strings of `DataCollection Successful` into an obvious constant
* Separates the interface for this class and for bluesky so we're not conflating them. e.g. the use of `fail` in here is a bluesky thing but not obviously

DominicOram: What are your thoughts?

src/artemis/ispyb/store_in_ispyb.py
```diff
@@ -102,6 +100,8 @@ def update_grid_scan_with_end_time_and_status(
             params["parentid"] = datacollection_group_id
             params["endtime"] = end_time
             params["run_status"] = run_status
+            if reason is not None and reason != "":
+                params["comments"] += " " + run_status + " reason: " + reason
```
DominicOram, line 104: Nit:
```suggestion
                params["comments"] += f" {run_status} reason: {reason}"
```
Feels a bit cleaner to me

DominicOram (approved):
Generally looks like it will work, I would be interested to get your opinion on the proposed refactor before merging though

PR #342, opened 2022-11-09
301 add zocalo environment parameter

### LINKED ISSUE #301: Be able to switch zocalo environment
For testing artemis/experimental plans off the beamline we should be able to use e.g. `devrmq` instead of `artemis`, to allow "processing" by a mock zocalo. But, `zc.activate_environment("zocalo")` is hardcoded throughout artemis, including in unit tests. 

**acceptance criteria:**
- zocalo env can be specified at runtime e.g. through parameters or --dev flag
- tests on zocalo interaction reflect this

dperl-dls:
Fixes #301 

Adds a parameter to parameters and zocalo_interaction in order to be able to specify the data processing (zocalo) environment at runtime. ~~Defaults to production "artemis" env if not specified, so it should not break any existing parameter configurations.~~ Mainly so that we can change to "devrmq" for testing and mock out zocalo; test_parameters.json is updated to reflect this.

Defaults to devrmq because it's too dangerous to be able to accidentally send requests to prod zocalo, see #356 

### To test:
1. verify tests still pass

### DIFF
--- a/src/artemis/fgs_communicator.py
+++ b/src/artemis/fgs_communicator.py
@@ -68,7 +68,7 @@ class FGSCommunicator(CallbackBase):
             datacollection_ids = self.ispyb_ids[0]
             self.datacollection_group_id = self.ispyb_ids[2]
             for id in datacollection_ids:
-                run_start(id)
+                run_start(id, self.params.zocalo_environment)
 
     def stop(self, doc: dict):
         artemis.log.LOGGER.debug(f"\n\nReceived stop document:\n\n {doc}\n")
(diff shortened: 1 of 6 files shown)


DominicOram (changes requested):
I think we should have a chat about this:
* I'm worried it now makes things annoying for using it in production
* I don't like that we have to pass the env around all over the place, maybe with the new callback structure it makes sense to pull the functions in `zocalo_interaction` into the callback class and have that hold the state of the environment

dperl-dls:
- actually, put zocalo_interaction functions into zocalo handler class
- it needs to get env information when initialised
- default to "artemis" on zocalo handler and "devrmq" in our test parameters etc

dperl-dls:
closed in favour of  #381

Issue #335, opened 2022-11-04
Move detector to robot load position


npat113:
Robot load action requires detector to be at a minimum distance of 337mm on I03 to avoid collision. This standalone task should be inserted as soon as it is possible where the next action is a robot load.

DominicOram:
Beamline configuration has changed, this is no longer required

PR #346, opened 2022-11-10
I03 basic devices

npat113:
Fixes # (issue)
basic starting devices to allow dev across branches

### To test:
everything


src/artemis/devices/I03Smargon.py
```diff
@@ -0,0 +1,25 @@
+from ophyd import Component as Cpt
+from ophyd import Device, EpicsMotor, EpicsSignal
+
+"""
+Real motors added to allow stops following pin load (e.g. real_x1.stop() )
+Stub offsets are calibration values that are required to move between calibration pin position and spine pins. These are set in EPCICS and applied via the proc.
+"""
+
+
+class I03Smargon(Device):
```
DominicOram, line 10: Should: There is already an I03 smargon in `devices/motors.py`. Could you add any additionally required PVs there? Or if you think it makes sense to have it in a separate smargon file we can do that too. Just don't want two copies

src/artemis/move_det_robot_load.py
```diff
@@ -0,0 +1,44 @@
+import bluesky.plan_stubs as bps
```
DominicOram, line 1: Should: I think this is part of https://github.com/DiamondLightSource/python-artemis/issues/335 right? Can we do it in a separate PR for that rather than here?

DominicOram (changes requested):
Couple of minor suggestions. You also have quite a few empty files:
* DCM
* beamstop
* cryostream
* fluorescence_detector_motion
* lower_gonio_stages
* qbpm1
* robot
* scintillator

I would rather not have empty files in the codebase as it leads to confusion. If you have some code for them now feel free to add to this PR, if not I suggest we remove from this PR and add when required later.

src/artemis/devices/I03Smargon.py
npat113: I think there is a discussion to be had on devices and how granular you want them in terms of file structure. Motors.py is either going to become very large, or should be split to individual devices. Should something similar to the EPICS grouping be used?

src/artemis/move_det_robot_load.py
npat113: yes, not sure how that sneaked in there

src/artemis/devices/I03Smargon.py
DominicOram: Yh, I agree that motors was probably too large a category. It made sense when the usecase was just grid scan but now we're looking wider afield it doesn't make as much sense. Happy to move it to here. In which case can you merge the one in `motors.py` and change all references? I can do this if you'd rather.

src/artemis/devices/CTAB.py
```diff
@@ -0,0 +1,24 @@
+from ophyd import Component as Cpt
+from ophyd import Device, EpicsMotor, EpicsSignalRO
+
+
+class CTAB(Device):
+    """Basic collimantion table (CTAB) device for motion plus the motion disable signal when laser curtain triggered and hutch not locked"""
+
+    inboard_y: EpicsMotor = Cpt(EpicsMotor, "-MO-TABLE-01:INBOARDY")
```
DominicOram, line 8: Nit: Do you know if there's anywhere this co-ord system is documented? I can write some down if not

src/artemis/devices/DCM.py
```diff
@@ -0,0 +1,24 @@
+from ophyd import Component as Cpt
+from ophyd import Device, EpicsMotor, EpicsSignalRO
+
+
+class DCM(Device):
+
+    # upstream_x: EpicsMotor = Cpt(EpicsMotor, '-MO-DET-01:UPSTREAMX')
+
```
DominicOram, line 8: Nit: Is this used here? We should remove it if not

src/artemis/devices/DCM.py
```diff
@@ -0,0 +1,24 @@
...
+    bragg: EpicsMotor = Cpt(EpicsMotor, "-MO-DCM-01:BRAGG")
+    roll: EpicsMotor = Cpt(EpicsMotor, "-MO-DCM-01:ROLL")
+    offset: EpicsMotor = Cpt(EpicsMotor, "-MO-DCM-01:OFFSET")
+    perp: EpicsMotor = Cpt(EpicsMotor, "-MO-DCM-01:PERP")
+    energy: EpicsMotor = Cpt(EpicsMotor, "-MO-DCM-01:ENERGY")
+    pitch: EpicsMotor = Cpt(EpicsMotor, "-MO-DCM-01:PITCH")
+    wavelength: EpicsMotor = Cpt(EpicsMotor, "-MO-DCM-01:WAVELENGTH")
```
DominicOram, line 15: Nit: Do you know of any documentation for what each of these do? If not can we write some comments? The ones in particular I'm not so sure about:
* offset
* perp

src/artemis/devices/I03Smargon.py
```diff
@@ -0,0 +1,25 @@
...
+    y: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:Y")
+    z: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:Z")
+    chi: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:CHI")
+    phi: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:PHI")
+    omega: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:OMEGA")
+
+    stub_offset_set: EpicsSignal = Cpt(EpicsSignal, "-MO-SGON-01:SET_STUBS_TO_RL.PROC")
+
+    real_x1: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:MOTOR_3")
+    real_x2: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:MOTOR_4")
```
DominicOram, line 21: Nit: Could you put in a comment about what the difference is between `x1` and `x2`?

src/artemis/devices/I03Smargon.py
```diff
@@ -0,0 +1,25 @@
...
+    real_y: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:MOTOR_1")
+    real_z: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:MOTOR_2")
+    real_phi: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:MOTOR_5")
+    real_chi: EpicsMotor = Cpt(EpicsMotor, "-MO-SGON-01:MOTOR_6")
```
DominicOram, line 25: Nit: Why do we need the real motors here? I would have thought we only ever really want to use the virtual axes?

src/artemis/devices/fluorescence_detector_motion.py
```diff
@@ -0,0 +1,7 @@
+from ophyd import Component as Cpt
+from ophyd import Device, EpicsSignal
+
+
+class FluorescenceDetector(Device):
+
+    pos: EpicsSignal = Cpt(EpicsSignal, "-EA-FLU-01:CTRL")
```
DominicOram, line 7: Nit: Pos implies a position, which I would assume is a motor but I'm guessing this is just an IN/OUT? In which case could you do something like the backlight where you define:
```
class Backlight(Device):
    OUT = 0
    IN = 1

    pos: EpicsSignal = Component(EpicsSignal, "-EA-BL-01:CTRL")
```

src/artemis/move_det_robot_load.py
DominicOram: You can remove it by doing `git rm --cached src/artemis/move_det_robot_load.py`. This will take it out of git but keep it locally if you're worried about losing it.

DominicOram (changes requested):
Thank you, mostly just requests for more documentation. Some of this may be my own lack of understanding but we should try and make it accessible to someone relatively new to DLS.

src/artemis/devices/DCM.py
npat113: no, that comment has sneaked in from elsewhere. Will remove

DominicOram (approved):
Great, thank you. I made some minor edits with moving the location of comments. If you put the comments in the right place then tooltips on IDEs will pick them up. So if you have a line that says `mv(smargon.stub_offset_set, 10)` with the comment located where it now is you can hover over `stub_offset_set` to get a description of what it is.

src/artemis/devices/I03Smargon.py
DominicOram: I will merge this PR and fix this in a subsequent issue

PR #347, opened 2022-11-11
#231 Adjust results from Zocalo

### LINKED ISSUE #231: Results from zocalo are from the top left corner
We are converting the returned results from zocalo to positions incorrectly. The value returned from zocalo defines zero as the top left of the first box, whereas the start position sent to the controller (and thus where we define zero) is the centre of the first box.

## Acceptance Criteria
* We account for this mismatch
* There are tests for this

dperl-dls:
Fixes #231

### To test:
1. Check this change functionally achieves the same thing as the fix on live
2. Confirm tests adjusted correctly and still passing


DominicOram (changes requested):
Looks correct to me. Can you change the docstrings on `wait_for_results` to talk about the co-ord system the results are returned in. Also a comment next to the -0.5 as to why we're doing it. Thanks!

DominicOram (approved):
Thanks!

PR #351, opened 2022-11-14
348 better points

### LINKED ISSUE #348: Replace Point3D with something which supports arithmetic
Currently Point3D is a NamedTuple which is sort of fine, but we have to do stuff like:

`corrected_centre = Point3D(xray_centre.x - 0.5, xray_centre.y - 0.5, xray_centre.z - 0.5)`

instead of being able to just subtract (0.5, 0.5, 0.5).

If we ever need to do more arithmetic with vectors/coordinates then we should replace this with a more robust class, potentially a simple homemade one which can just add/subtract/multiply/divide, but ideally with one from a well-maintained module which supports distances, angles and whatnot, and might be a bit optimised.

dperl-dls:
Fixes #348 

Replaces NamedTuple in artemis/utils.py with subclasses of np.ndarray to allow doing arithmetic and other numpy functions on Point2D and Point3D. Provides all the same functionality we relied on with the namedtuple.

### To test:
1. Confirm tests still pass

### DIFF
--- a/src/artemis/fgs_communicator.py
+++ b/src/artemis/fgs_communicator.py
@@ -11,6 +11,7 @@ from artemis.nexus_writing.write_nexus import (
     create_parameters_for_second_file,
 )
 from artemis.parameters import ISPYB_PLAN_NAME, FullParameters
+from artemis.utils import Point3D
 from artemis.zocalo_interaction import run_end, run_start, wait_for_result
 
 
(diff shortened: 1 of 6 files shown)


dperl-dls:
This depends on #347 , and anyway even though tests pass pytest segfaults on exit so that's not very good, so will revisit later

PR #354, opened 2022-11-15
#298 Added similar logic as before

evalott100:
Fixes #(251)

@DominicOram  accidentally messed up my branches. This will replace #298  as  a pull request (I'll close it).

### DIFF
--- a/src/artemis/fast_grid_scan_plan.py
+++ b/src/artemis/fast_grid_scan_plan.py
@@ -1,4 +1,5 @@
 import argparse
+import math
 
 import bluesky.plan_stubs as bps
 import bluesky.preprocessors as bpp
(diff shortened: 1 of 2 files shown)


DominicOram (changes requested):
Looks good, can you fix the merge conflicts and add a test that covers the new logic? Thanks!

DominicOram:
I've done the merge on this so that we can pull it in for testing today but we should still add a unit test please

DominicOram:
I think with https://github.com/DiamondLightSource/python-artemis/pull/381 merged we should move this into the Zocalo class, e.g. when you initialise the class you pass t the fallback xyz and this logic is handled in the class itself

evalott100:
@DominicOram  By fallback value you mean providing a function argument to fall back on instead of defaulting to the optical centre?

Move which part into the Zocalo class sorry, this part?

``` python
    # We move back to the centre if results aren't found
    if math.nan in subscriptions.zocalo_handler.xray_centre_motor_position:
        log_msg = f"No diffraction found, moving to optical centre {initial_xyz}"
        artemis.log.LOGGER.warn(log_msg)
        yield from move_xyz(fgs_composite.sample_motors, initial_xyz)
        raise NoCentreFoundException(f"Zocalo: {log_msg}")
```

evalott100:
@DominicOram is this the kind of thing you mean?

DominicOram (approved):
Yes, looks good to me, thanks. Happy for you to merge once you've fixed the conflicts

PR #358, opened 2022-11-17
#355 example refactoring of zebra

### LINKED ISSUE #355: Make Zebra more generic
The Zebra device is currently written very specifically for FGS, eg:

```python
class Zebra(Device):
    (...)
    def stage(self) -> List[object]:
        self.setup_fast_grid_scan()
        self.output.disable_fluo_collection()
        return super().stage()
```

We should make it more generic in order to use it for rotation scans, and anything else we might add in the future.

dperl-dls:
- for multiple Devices (for each type of scan)
- but hopefully without too much duplication

Fixes #355 

### To test:
1. Think about whether this is a good way to organise things
2. Check tests pass


dperl-dls:
On discussion with @DominicOram, decided to keep one zebra device and extract the varying logic into plans

dperl-dls:
Closed in favour of #365

PR #365, opened 2022-11-18
355 move zebra logic to plan

### LINKED ISSUE #355: Make Zebra more generic

dperl-dls:
Fixes #355

Allows us to set up the zebra in the experimental plan in order to customise it for different experiments

### To test:
1. Verify this makes sense
2. Check tests pass (ideally with S03 and script for slit gaps)


src/artemis/fast_grid_scan_plan.py
```diff
@@ -122,6 +128,9 @@ def gridscan_with_communicator(fgs_comp, det, params):
         fgs_composite.sample_motors, communicator.xray_centre_motor_position
     )
 
+    # tidy up
+    yield from set_zebra_shutter_to_manual(fgs_composite.zebra)
```
DominicOram, line 132: Should: We probably need to do this as a `finally`. I think something like [this](https://nsls-ii.github.io/bluesky/generated/bluesky.preprocessors.finalize_decorator.html?highlight=finalizer) will help

dperl-dls: ah yeah good idea,

src/artemis/device_setup_plans/setup_zebra_for_fgs.py
```diff
@@ -0,0 +1,25 @@
...
+    TTL_XSPRESS3,
+    Zebra,
+)
+
+
+def setup_zebra_for_fgs(zebra: Zebra):
+    yield from abs_set(zebra.output.out_pvs[TTL_DETECTOR], IN3_TTL)
+    yield from abs_set(zebra.output.out_pvs[TTL_SHUTTER], IN4_TTL)
+    yield from abs_set(zebra.output.out_pvs[TTL_XSPRESS3], DISCONNECT)
+    yield from abs_set(zebra.output.pulse_1_input, DISCONNECT)
```
DominicOram, line 20: Nit: I know this is an extension but it would be nice to put these all in a `group` that we can wait on later now that it's easy to do so.

src/artemis/system_tests/test_fgs_plan.py
```diff
@@ -0,0 +1,142 @@
...
+        prefix=params.beamline,
+    )
+    return fast_grid_scan_composite
+
+
+@pytest.mark.s03
+@patch("bluesky.plan_stubs.wait")
+@patch("bluesky.plan_stubs.kickoff")
+@patch("bluesky.plan_stubs.complete")
+def test_run_gridscan(
```
DominicOram, line 65: Should: This test currently fails due to the scan params being invalid

src/artemis/system_tests/test_fgs_plan.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+@pytest.mark.s03
+@patch("bluesky.plan_stubs.wait")
+@patch("bluesky.plan_stubs.kickoff")
+@patch("bluesky.plan_stubs.complete")
+@patch("artemis.fgs_communicator.FGSCommunicator")
+@patch("artemis.fast_grid_scan_plan.tidy_up_plans")
+@patch("artemis.fast_grid_scan_plan.run_gridscan_and_move")
+def test_full_plan_tidies_at_end_when_plan_fails(
```
DominicOram, line 128: Should: Both these tests fail with due to an import issue

DominicOram: Also, given we're running these as system tests anyway do we need to mock out the `tidy_up_plans`? And if we're running with `fake_zocalo` and the dev database we probably don't need to mock out the communicator either? My general philosophy with the tests is that for unit test we mock out lots so that we're just testing the thing we want to test as unit tests are quick and good to have lots of. But for system tests is the opposite we want to test as much as possible because they tend to be slower and we probably only want a few of them

src/artemis/system_tests/test_fgs_plan.py
dperl-dls: fixed

src/artemis/system_tests/test_fgs_plan.py
dperl-dls: communicator hasn't existed for a few revisions now already. fixed version uses the real callbacks.

src/artemis/device_setup_plans/setup_zebra_for_fgs.py
dperl-dls: done

src/artemis/system_tests/test_fgs_plan.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+    return fast_grid_scan_composite
+
+
+@pytest.mark.s03
+@patch("artemis.fast_grid_scan_plan.wait_for_fgs_valid")
+@patch("bluesky.plan_stubs.wait")
+@patch("bluesky.plan_stubs.kickoff")
+@patch("bluesky.plan_stubs.complete")
+def test_run_gridscan(
```
DominicOram, line 68: Nit: This test still doesn't work for me but I think it's outside the scope of this issue so going to put a skip on it and approve this PR.

DominicOram (approved):
Thank you!

PR #382, opened 2022-11-22
129 opentelemetry tracing

### LINKED ISSUE #129: Publish metrics about time taken for each step in a plan

dperl-dls:
Fixes #129
Adds an OpenTelemetry tracer to time different steps of the plan, which we can easily export to prometheus or any other database/ui for monitoring

### To test:
1. For testing, run the Jaeger all-in-one container: https://www.jaegertracing.io/docs/1.6/getting-started/
2. Run artemis in dev/S03/fake zocalo mode and run through the plan. you may need to wrap do_fgs in a try block and throw away the odin/filewriter waittimeouterror
3. Confirm tracing results show up in Jaeger UI, maybe add some more `with tracer.start_span("span_name"):` in various places


DominicOram:
@evalott100 could you review this? Mostly just need to make sure that it's obvious how to get it set up and working locally from the docs

evalott100:
> @evalott100 could you review this? Mostly just need to make sure that it's obvious how to get it set up and working locally from the docs

@DominicOram Which docs, the jaeger docs or are there some on the artemis repo I'm missing?

DominicOram:
@evalott100 All docs basically, can you _easily_ get a locally running GUI showing how long Artemis operations are taking using the information provided. If not, @dperl-dls should write more docs on the README/wiki

evalott100 (approved):
I now consider the documentation in the README sufficient.

PR #412, opened 2022-11-30
fix old import in oav centring

### LINKED ISSUE #411: Some files still importing I03Smargon from motors.py
That file doesn't exist anymore so this breaks test discovery

dperl-dls:
Fixes #411

### To test:
1. check imports/tests resolve correctly


evalott100 (approved):
No issues here

Issue #370, opened 2022-11-18
Make nexus writer deal with different experiment types


dperl-dls:
The nexus writer is currently very hacky and should be tidies in general as part of this change (As discussed in [#592)](https://github.com/DiamondLightSource/python-artemis/pull/592#discussion_r1150807798). Most of the functionality should probably end up in Nexgen, and what's left in Artemis should be more generic, and/or possibly only handled by the callbacks.

Issue #369, opened 2022-11-18
Confirm Zocalo handler deal with rotation scans


dperl-dls:
Once the data is correct in ispyb, this is just triggered in the same way as grid scan. This issue is just to confirm it works

DominicOram:
Once the data is correct in ispyb, this is just triggered in the same way as grid scan

Issue #368, opened 2022-11-18
Make ispyb handler deal with different experiment types


dperl-dls:
- `StoreInIspyb` base class should become more general so that other scan types can use it
  - `store_grid_scan` should probably move out into its own thing
- ~~we should minimise any calculation of things in `StoreInIpsyb` or deriving classes, such as the filename - instead of things like `params["file_template"] = f"{self.detector_params.prefix}_{self.run_number}_master.h5"` the full filename should be known and determined by the InternalParameters - also relevant to #370~~ now in #707

dperl-dls:
#43 #45 #372 should be done as part of this (if still necessary)

PR #416, opened 2022-12-05
413 systematize callbacks

### LINKED ISSUE #413: Systematise organisation of internal interaction and callbacks
Currently, we have a situation where `zocalo_interaction` contains the interaction and the callback code, while `write_nexus` and `ispyb_interaction` are independent, and the ispyb and nexus callbacks are together in one file. this is horribly unsystematic.

I propose that we separate the zocalo callback back out from the zocalo interaction file, and split up the callbacks from each other, so that:
- all the code needed for zocalo interaction (for any type of experiment) is in `zocalo_interaction`
- all the code needed for ispyb interaction (for any type of experiment) is in` ispyb_interaction`
- all the code needed for nexus writing (for any type of experiment) is in `write_nexus`
- and the callbacks are all individual for experiment type, eg: `gridscan_ispyb_callback`, `rotation_zocalo_callback` etc.

That way it's the same for everything, we minimise duplication of code for the actual interactions, and minimise checking what the experiment is in the callback, just subscribe to the correct one in the plan.

Any thoughts?

dperl-dls:
Fixes #413 

Reorganises the external interaction submodule so that it can better be extended for different experiment types (eg #368 #369 #370)

### To test:
1. check the structure for sanity
2. run tests


dperl-dls:
@DominicOram Would really like it if this one could be reviewed sooner rather than later - it's just shuffling stuff around, no actual changes to functionality, but it will make merging other things a pain

Issue #333, opened 2022-11-04
OAV Centring


evalott100:
GDA currently carries out OAV centring on the pin.

Acceptance Criteria:
* Rotation of the pin to find the biggest width
* Takes the centre at the biggest width, and the centre at the angle orthogonal to the biggest width, Combines them to make a 3D coordinate.
* Translates the 3D coordinates to smargon xyz motor coordinates, and moves there.

evalott100:
Closing since acceptance criteria are met, will next write system tests.

DominicOram:
I don't think we can call this done until https://github.com/DiamondLightSource/python-artemis/pull/394 is merged?

PR #418, opened 2022-12-05
V0.9.10

### LINKED ISSUE #248: Clean up and give better error on SCAN_INVALID
If the gridscan params would cause the smargon to go out of it's limits then the `SCAN_INVALID` PV goes low and the fgs kickoff times out. We're correctly checking this but we're not cleaning up well afterwards:

# Acceptance criteria
* When the scan is invalid it is reported back to GDA nicely (Note that this is a warning for GDA it shouldn't stop the queue)
* This cleans up the detector and does not leave it in an armed state

dperl-dls:
Live fixes from testing on 05/12/2022 which were not too hacky to merge
Fixes #248


DominicOram:
This is now really large because it contains changes from https://github.com/DiamondLightSource/python-artemis/issues/333 and minor changes from the testing. I think we need a new PR for those minor changes and then to tackle https://github.com/DiamondLightSource/python-artemis/issues/333 in https://github.com/DiamondLightSource/python-artemis/pull/394. I'm happy to split this out

DominicOram:
I believe all the functionality in this PR is now covered by:
* https://github.com/DiamondLightSource/python-artemis/pull/429
* https://github.com/DiamondLightSource/python-artemis/pull/425
* https://github.com/DiamondLightSource/python-artemis/pull/430
* OAV work, that we can pull in for testing as and when

So I will close this PR as won't merge

PR #423, opened 2022-12-08
 Fix initialisation of OAV device and parameter loading

### LINKED ISSUE #422: Fix initialisation of OAV device and parameter loading
# Testing criteria:
* PVs are accurately set.
* Zoom controller using the enum "[zoomlevel]x" PV instead of the percentage PV.
* Add methods for loading xml zoom levels and display.configuration.

evalott100:
Fixes #422 

### To test:
1. Check OAVParameters is capable of loading in the correct beam centre, and microns to pixels for a given zoom.
2. Check PVs are being set to the correct values on run.
3. Check the zoom level can't be set to an unknown value (other than 1.0x, 2.5x, 5.0x, 7.5x, 10x).


DominicOram (changes requested):
Looks good other than minor requested changes.

DominicOram (changes requested):
Couple of comments, otherwise fine.

DominicOram (approved):
Great, thank you!

PR #425, opened 2022-12-09
(#248) Wait for valid scan parameters before starting scan

### LINKED ISSUE #248: Clean up and give better error on SCAN_INVALID

DominicOram:
Fixes #248

This will check the scan invalid and stop us doing the scan if it is invalid. I think there are potentially nicer ways to do it in the future but we keep patching this onto the beamline anyway so should merge it then tidy later (See #426)

### To test:
1. Confirm new unit tests pass


dperl-dls (approved):
Confirmed tests passing

PR #430, opened 2022-12-09
(#389) run_artemis will now fail if not run as the correct user

### LINKED ISSUE #389: Fix run script
The `run_artemis` script is currently not working. It has the following issues:
* It runs Artemis using the old `pipenv` way
* It does not kill a running Artemis correctly
* It does not correctly `ssh`/`su`

## Acceptance Criteria
* Fix the above, if we cannot `ssh`/`su` properly then we can just check we're on the right machine/user and inform the user if not

DominicOram:
Fixes #389

### To test:
1. Run `./run_artemis.sh -b=i03` and confirm you get an error about not being on the right machine
2. Confirm `run_artemis.sh` contains changes from https://github.com/DiamondLightSource/python-artemis/pull/418/files


dperl-dls (approved):
looks right, message appears correctly on my machine

PR #434, opened 2022-12-12
Fix checking for host/user in run_artemis

DominicOram:
### To test:
1. Tested on beamline 
2. for local testing just confirm that the error message printed now shows the user/hostname when script is run


dperl-dls (approved):
Seems to work

PR #407, opened 2022-11-29
399 random order tests

dperl-dls:
Fixes #399 
Randomises the order in which tests are run, including in CI, in order to help identify any interdependent tests.

Tests will fail currently; they are fixed in #405, which is merged into main now

### To test:
0. You might need to `pip install pytest-random-order` (or run `./dls_dev_env.sh` again)
1. Run `pytest --random-order -m "not s03` 
2. Observe that some ispyb tests fail sometimes
3. Merge main
4. Rerun tests a few times and see they always pass

PR #443, opened 2022-12-13
399 redo random order tests

### LINKED ISSUE #399: Randomise tests
https://github.com/DiamondLightSource/python-artemis/issues/393 was missed because the tests had issues where they shared state between them. If we randomise the tests this should hopefully be less likely to happen.

## Acceptance Criteria
* Tests run in CI are in a random order
* This random order is reproducible and it is documented how to reproduce it (e.g. if tests fail in a particular order I can run this order locally)

dperl-dls:
Fixes #399 
Replaces #407 

Randomises the order in which tests are run, including in CI, in order to help identify any interdependent tests.

### To test:
0. see tests pass even in a random order

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -54,39 +54,11 @@ jobs:
 
 
       - name: Run tests
-        run: pytest --random-order -m "not s03"
+        run: pytest --random-order -m "not (dlstbx or s03)"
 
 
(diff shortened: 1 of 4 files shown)


evalott100 (approved):
Reviewed using the `-v` tag to print all test methods as they go, both test order and file order seem random :)

PR #447, opened 2022-12-15
446 improve error logging

### LINKED ISSUE #446: Improve error logging
We often find we are missing useful info from our logs, but if we run in `--logging-level DEBUG` then we get every single ophyd PV put logged to graylog. We should have an intermediate level where extra information from artemis and also the most important info from ophyd/epics is logged.

dperl-dls:
Fixes #446 by:
- adding a command line arg to add a callback which logs all documents
- adding an InfoLoggingDevice subclass of ophyd.Device which adds some INFO level logging around waiting for connection
- tidying up some other bits and pieces of logging

### To test:
1. Run tests


DominicOram (changes requested):
As seen on beamline, cli arg does not get correctly passed through as we're using it in the class variables but not in th `start` function

DominicOram (approved):
Looks good, I added a little more logging and fixed some typos.

PR #455, opened 2022-12-16
Put the FGS device inside I03Smargon.

### LINKED ISSUE #155: Move FGS into smargon

evalott100:
Fixes #155 

New pull request (old one is #283 ) for adding the fast grid scan device into the smargon.


DominicOram (changes requested):
Now this is in the smargon can we update the plans/tests to use it from there rather than creating a second one in the `FGSComposite`

DominicOram:
I'm going to close this w/o merging, I think we can do it in dodal instead

PR #463, opened 2022-12-19
366 simplify params

dperl-dls:
further steps in separating the external and internal params


DominicOram:
Closing without merging as this is now too out of date to be useful. We can pick the work back out in a new PR if we want

PR #464, opened 2022-12-19
(#450) Run artemis fixes

### LINKED ISSUE #450: run_artemis improvements
There are a couple of improvements we can make to `run_artemis.sh`:
* We don't need to exit if not on a beamline, it should just start a dev artemis. Change the docs to reflect this as the recommended way of starting artemis
* We need to be able to run the script from anywhere and so it should cd into the correct place for running (this should work for dev mode or if running from `artemis_v...`)
* The script should only finish when Artemis is ready to accept commands

DominicOram:
Fixes #450

### To test:
1. On main call `run_artemis` and then immediately try to push a new set of params to the server, it will fail as the server is not yet up
2. On this branch confirm server is up once `run_artemis` is finished
3. Call `run_artemis` from a different directory and confirm it still works
4. Confirm `README` is sensible

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,18 +16,11 @@ Controlling the Gridscan Externally (e.g. from GDA)
 
 Starting the bluesky runner
 -------------------------
-You can start the bluesky runner by doing the following:
-```
-source .venv/bin/activate
-python -m artemis
-```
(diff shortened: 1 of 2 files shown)


dperl-dls (approved):
script works as described on my machine

readme changes  look good

PR #467, opened 2022-12-20
435 don't trigger zocalo if scan invalid and give better message on scan_invalid

### LINKED ISSUE #435: Invalid scan parameters still trigger zocalo
If the parameters for the grid scan are invalid we still trigger zocalo and thus get an alert from zocalo saying that it can't find datafiles

## Acceptance Criteria
* Zocalo is not triggered if scan is invalid

### LINKED ISSUE #512: Improve message on scan invalid
When the scan is invalid we currently write `Unsuccessful reason: Scan parameters invalid after 0.5 seconds` to ispyb. This isn't useful to a user. Instead we should write something like `Pin exceeds travel range, unable to scan`

dperl-dls:
Fixes #435 - zocalo is now triggered by a start document emitted at the beginning of `do_fgs()`
Also Fixes #512 

### To test:
1. Run tests


dperl-dls:
ah yeah it does

DominicOram (commented):
Sorry, I've reviewed this for style but not yet for functionality. I will get back to it ASAP

src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -155,9 +177,13 @@ def run_gridscan_and_move(
...
         yield from run_gridscan(fgs_composite, detector, params)
 
-    artemis.log.LOGGER.debug("Starting grid scan")
+    artemis.log.LOGGER.info("Starting grid scan")
     yield from gridscan_with_subscriptions(fgs_composite, eiger, parameters)
 
+    # bps.create()
+    # yield from wait_for_fgs_valid(fgs_composite.fast_grid_scan)
+    # bps.save()
+
```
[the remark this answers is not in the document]
dperl-dls: oops, was meant to be deleted after checking putting it at line 134 worked...

src/artemis/external_interaction/callbacks/fgs/zocalo_callback.py
```diff
@@ -42,30 +44,39 @@ def __init__(
...
 
     def stop(self, doc: dict):
-        LOGGER.debug("Zocalo handler received stop document.")
-        if self.ispyb.ispyb_ids == (None, None, None):
-            raise ISPyBDepositionNotMade("ISPyB deposition was not initialised!")
-        datacollection_ids = self.ispyb.ispyb_ids[0]
-        for id in datacollection_ids:
-            self.zocalo_interactor.run_end(id)
-        self.processing_start_time = time.time()
+        if self.started_run:
```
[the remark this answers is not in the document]
dperl-dls: because all callbacks receive all start and stop documents. so it recieves a start document at the beginning, but if `wait_for_fgs_valid()` fails, it will then recieve a stop document. `self.started_run` is set at the beginning of `do_fgs`, rather than at the beginning of `run_gridscan_and_move_and_tidy`, so it should only do `zocalo_interactor.run_end()` if it actually started a run

src/artemis/external_interaction/callbacks/fgs/zocalo_callback.py
```diff
@@ -42,30 +44,39 @@ def __init__(
...
+            if doc.get("run_start") == self.run_gridscan_uid:
+                LOGGER.info(
+                    f"Zocalo handler received stop document, for run {doc.get('run_start')}, and started run = {self.started_run}, uid : {self.run_gridscan_uid}"
```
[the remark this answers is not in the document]
dperl-dls: no, there are many sub-runs ("run_gridscan", "do_fgs", "move_xyz" ...; set with `@bpp.set_run_key_decorator()`) and we care about the stop document associated with the run_start uid for the sub-plan which the zocalo handler is using for a trigger

this will be tidier after #468 but that is probably longer-term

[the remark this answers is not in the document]
dperl-dls: oh, whoops, I misunderstood your comment at first glance. agree.

src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -118,19 +123,25 @@ def run_gridscan(
     yield from set_fast_grid_scan_params(fgs_motors, parameters.grid_scan_params)
     yield from wait_for_fgs_valid(fgs_motors)
 
-    @bpp.stage_decorator([eiger])
```
DominicOram, line 121: Must: We still need to stage the eiger

src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -118,19 +123,25 @@ def run_gridscan(
...
         yield from bps.wait()  # Wait for all moves to complete
         yield from bps.kickoff(fgs_motors)
         yield from bps.complete(fgs_motors, wait=True)
 
     with TRACER.start_span("do_fgs"):
-        yield from do_fgs()
+        try:
+            yield from do_fgs()
+        except Exception as e:
+            artemis.log.LOGGER.info(e, exc_info=True)
```
DominicOram, line 137: Must: Why are we logging this here? If we need to it should probably be worse than `info`, also we probably want to re-raise so as to do things like kill the RE

src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -192,7 +203,11 @@ def create_devices():
     )
 
     artemis.log.LOGGER.info("Connecting to EPICS devices...")
-    fast_grid_scan_composite.wait_for_connection()
+    try:
+        fast_grid_scan_composite.wait_for_connection()
+    except Exception as e:
+        artemis.log.LOGGER.info(e, exc_info=True)
```
DominicOram, line 209: Must: As above with logging

src/artemis/external_interaction/callbacks/fgs/zocalo_callback.py
```diff
@@ -42,30 +44,34 @@ def __init__(
         ] = parameters.grid_scan_params.grid_position_to_motor_position
         self.processing_start_time = 0.0
         self.processing_time = 0.0
+        self.started_run: bool = False
```
DominicOram, line 47: Should: Do we need `started_run` now?

src/artemis/unit_tests/test_fast_grid_scan_plan.py
```diff
@@ -176,6 +224,7 @@ def test_individual_plans_triggered_once_and_only_once_in_composite_run(
             subscriptions,
         )
     )
+    # RE(bps.close_run())
```
DominicOram, line 227: Should: Remove commented out code

DominicOram (changes requested):
Sorry, again this isn't a full review but caught some more issues :(

src/artemis/experiment_plans/fast_grid_scan_plan.py
dperl-dls: wow, I really messed something up here, this was my hacky workaround for running the full plan with S03

src/artemis/experiment_plans/fast_grid_scan_plan.py
dperl-dls: but now that you mention it, logging as error and raising seems good

DominicOram: I'm not sure if here is the best place to do that though, I think we should be still getting those logs at the runner anyway? Logging here just makes the plan itself less readable

src/artemis/system_tests/test_fgs_plan.py
```diff
@@ -6,6 +6,11 @@
 
 from artemis.devices.eiger import EigerDetector
```
DominicOram, line 7: Should: As discussed, these are the same as https://github.com/DiamondLightSource/python-artemis/pull/508/files, I will review that then we can remove the changes here.

DominicOram (commented):
As discussed I'm going to write the following tests:
* Unit test - That if zocalo is triggered before there is an ispyb entry we get an exception
* System test - Given scan invalid there's no zocalo triggering but is ispyb entry (with correct comment)
* System test - given scan valid zocalo is triggered and the correct dcgid is returned

DominicOram (approved):
Thanks for this, sorry it took so long to get it merged. It looks good now and clears up quite a few different issues. I've added the tests we discussed, fixed the issue with the `stop` not getting the metadata and removed the excess logging. Feel free to merge if you think it looks good or happy to discuss further

Issue #469, opened 2022-12-20
device naming: slit_gaps


npat113:
There are 4 sets of slits on I03 with 2 versions of definition. We should be specific about which set we are using. Helpfully the set in this case is the odd one out in terms of naming. 

S1-3 slits: size and centre
S4 slits: gap and centre

have submitted a renamed version to specify s4

Issue #465, opened 2022-12-19
Unit tests are broken with latest bluesky/ophyd version 


DominicOram:
The unit tests are currently broken on a clean install with a number of `ValueError: name must be a string.` errors from ophyd. We should fix these.

## Acceptance Criteria
* Unit tests pass with a clean install
* If the problem is with ophyd appropriate issue is raised there (has happened in the past where Fake devices react differently to real ones)

DominicOram:
@noemifrisina couldn't reproduce this. I will double check whether I can

DominicOram:
Also can't reproduce, closing

PR #429, opened 2022-12-09
428 fix end writing to ispyb

DominicOram:
Fixes #428

### To test:
1. Run new system test against code in main, confirm it fails
2. Run against this branch, confirm it passes

Issue #428, opened 2022-12-09
End details not written to ISPYB


DominicOram:
During testing we found that we were not successfully writing to Ispyb at the end of the data collection.

## Acceptance Criteria
* We do successfully write at the end of data collection
* There is a test for this

PR #470, opened 2022-12-20
altered slit gaps to specify s4

npat113:
Fixes #(469)
renamed slits to specify s4

### To test:
Everything - did I break things!


src/artemis/devices/s4_slit_gaps.py
```diff
@@ -0,0 +1,7 @@
+from ophyd import Component, Device, EpicsSignal
+
+
+class S4SlitGaps(Device):
+
+    s4xgap: EpicsSignal = Component(EpicsSignal, "S4XGAP")
+    s4ygap: EpicsSignal = Component(EpicsSignal, "S4YGAP")
```
DominicOram, line 7: Must: By changing the PV from `"XGAP"` to `"S4XGAP"` this is effectively telling ophyd to look for PVs called `BL03I-AL-SLITS-04:S4XGAP`, which I don't think exist:
```
[ffv81422@ws455 tickit-devices]$ caget BL03I-AL-SLITS-04:S4XGAP
Channel connect timed out: 'BL03I-AL-SLITS-04:S4XGAP' not found.
[ffv81422@ws455 tickit-devices]$ caget BL03I-AL-SLITS-04:XGAP
BL03I-AL-SLITS-04:XGAP         0.5
```

DominicOram: Nit: I don't think I would change the `xgap`/`ygap` here necessarily `s4_slits.xgap` reads a bit better to me than `s4_slits.s4xgap` as it seems needless repetition.

DominicOram (changes requested):
Thanks Neil. I had naively assumed that we could treat the `SlitGaps` as generic e.g. we could have something like:
```
s3_slit_gaps = Component(SlitGaps, "-AL-SLITS-03:")
```
for your 3rd set of slits and so we didn't need to necessarily specify S4 in the Device name but from looking at PVs this is wrong. It seems only your S4s have a `XGAP`, others have `X` and `X:SIZE` etc. So I think this change probably makes sense. 

Other than the comment I've added in the code itself this looks like it would work fine for on the beamline but it breaks some of our tests. You can see the broken tests [here](https://github.com/DiamondLightSource/python-artemis/actions/runs/3742337989/jobs/6353144485) and run them yourself locally by doing:
```
./dls_dev_env.sh
source .venv/bin/activate
pytest
```
I think the fix shouldn't be too bad but happy to have a look with you in the new year if you like. Note that some of the tests may be broken by other things like [this](https://github.com/DiamondLightSource/python-artemis/issues/465) but I think if there are any errors around `slits` this will be from this change. I think running the tests will also highlight the `S4XGAP` issue I have mentioned.

DominicOram:
Other minor admin things (none of which are a big deal but good to check in future if you remember):
* You've written `Fixes #(469)` in the PR, it should actually be `Fixes #469` for github to correctly link but I agree the template doesn't make this obvious, I'll fix the template. I think if I write Fixes #469 here now it will remedy this...
* If you write `#ticket_num` in the commit message github will pick also do the linking e.g. see commits [here](https://github.com/DiamondLightSource/python-artemis/pull/429/commits) link to https://github.com/DiamondLightSource/python-artemis/issues/428. Other developers often forget though so meh

npat113:
No problem. Will also expand the device to include the centre and real motors as well. 

Query: Should the gap motor be a EpicsSignal or a EpicsMotor? It's a compound motion of 2 real motors. I had been working on the basis of "if it physically moves then it is a motor".

DominicOram:
> No problem. Will also expand the device to include the centre and real motors as well.

> Query: Should the gap motor be a EpicsSignal or a EpicsMotor? It's a compound motion of 2 real motors. I had been working on the basis of "if it physically moves then it is a motor".

The distinction is more based on what the PV is in EPICS, if it's a motor record best to use `EpicsMotor`, otherwise `EpicsSignal`. I normally do this by looking whether it has a standard motor screen in the synoptic or whether records like `caget BL03I-AL-SLITS-04:XGAP.VELO` exist. In the case of the gaps they are actually motor records so we should use `EpicsMotor` (even though we mistakenly weren't before)

DominicOram:
@npat113, are you happy for one of us to tidy this up and get it merged for you?

npat113:
go for it

DominicOram (approved):
I've fixed the tests, would appreciate @dperl-dls or @noemifrisina double checking tests still pass (including system tests) and code looks sensible

noemifrisina:
Code looks good to me. 
Unit tests and most system tests still fine. However both `test_fgs_plan` and `test_zocalo_system` seem to hang, without showing any errors. @dperl-dls  could you double check those please?

noemifrisina:
Update on previous comment. Tests are now all running, however `test_eiger` fails with the message:
```
FAILED src/artemis/devices/unit_tests/test_eiger.py::test_when_set_odin_pvs_called_then_full_filename_written - AssertionError: assert '0' == 'test_0'
```

DominicOram:
@noemifrisina can you try now https://github.com/DiamondLightSource/python-artemis/pull/548 is merged

noemifrisina:
Run pytest a couple of times, looks like tests all pass now

noemifrisina (approved):

PR #471, opened 2023-01-09
Move test_plan_system to the correct location

noemifrisina:
Move test_plan_system.py from unit_tests/ to system_tests/

### To test:
1. Run system tests


DominicOram (approved):
Great, thank you!

PR #474, opened 2023-01-18
Move OAV centering plan to device_setup_plans

noemifrisina:
`oav_centering_plan.py` is a plan and not a device. Move from `devices` to `device_setup_plans`


dperl-dls (approved):
seems sensible

PR #483, opened 2023-01-23
Zocalo no longer returns reversed coordinates

rjgildea:
See python-dlstbx#196

### To test:
1. Test artemis X-ray centring with latest version of Zocalo X-ray centring service
2. Confirm that artemis moves to the correct location


dperl-dls (approved):
This seems fine, I am approving but not merging yet, for testing on Thursday as discussed

PR #484, opened 2023-01-23
add external interaction classes to __init__

dperl-dls:
as in title

Will let you do 
```python
from artemis.external_interaction import ZocaloInteractor
```
etc.


noemifrisina (approved):
Looks fine to me

Issue #487, opened 2023-01-24
Log no diffraction into ispyb


DominicOram:
When we find no diffraction in the sample we should put this into the isypb comments

## Acceptance Criteria
* No diffraction results from zocalo causes message in the ispyb comments
* System/unit tests for this

PR #488, opened 2023-01-24
439: No diffraction

### LINKED ISSUE #439: Behave correctly on no diffraction
During testing it was found that we're not behaving correctly on no diffraction results from zocalo:
* Checking `math.nan in results` fails as results contains `float("nan")` objects
* The storing of the optical centre is wrong and so we cannot move there properly
* We should probably still put something in ispyb in this case

## Acceptance Criteria
* The above are fixed
* We have tests to confirm that this behaves as expected (both unit and system)

DominicOram:
Fixes #439 

Does not include ispyb interaction, see #487 for this,

### To test:
1. Confirm unit tests pass


dperl-dls (approved):
looks good

PR #491, opened 2023-01-24
#481 Add beamline field to graylog

### LINKED ISSUE #481: Log beamline in graylog
Other messages in graylog will give the beamline that they've come from. Artemis does not, making debugging a bit harder

DominicOram:
Fixes #481

### To test:
1. Run: 
```
source .venv/bin/activate
export BEAMLINE=i04-1
python -m artemis
```
1. Check the logs in https://graylog2.diamond.ac.uk/streams/631745613d8af00a49bfee55/search and confirm the latest log is from you and contains the expected beamline


noemifrisina (approved):
Looks good, Graylog returns the expected beamline and source.

dperl-dls (changes requested):
this test will break if merged with main because there is no corresponding entry in `get_beamline_prefixes()` for i04-1 and artemis will crash on launch with a not very helpful message

```    
...
File "/scratch/ziq44869/Development/python-artemis/src/artemis/experiment_plans/fast_grid_scan_plan.py", line 180, in create_devices
f"Creating devices for {prefixes.beamline_prefix} and {prefixes.insertion_prefix}"
AttributeError: 'NoneType' object has no attribute 'beamline_prefix'
```

so Must: handle this error more gracefully (eg "beamline environment variable is [i0xx]; no corresponding prefix found" or something) and
Should: add some more relevant beamlines which we are likely to use in the medium term

DominicOram:
I've added a sensible exception, I don't want to add additional beamlines as it gives an idea we support them, which we really don't yet

PR #492, opened 2023-01-24
438: Reset FGS position counter on start up

### LINKED ISSUE #438: Reset position counter before waiting on it
Following https://github.com/DiamondLightSource/python-artemis/issues/248 we added a wait earlier on in Artemis for the position counter to be reset. Currently on main this will always fail as the position counter never gets set to 0 before we wait on it.

DominicOram:
Fixes #438

### To test:
1. Run the new `test_given_valid_params_when_kickoff_then_completion_status_increases_and_finishes` against main and confirm it fails
2. Confirm it passes on this branch

### DIFF
--- a/src/artemis/devices/fast_grid_scan.py
+++ b/src/artemis/devices/fast_grid_scan.py
@@ -1,7 +1,6 @@
 import threading
 import time
 from dataclasses import dataclass
-from typing import List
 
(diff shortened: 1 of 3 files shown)


dperl-dls (approved):
Seems ok, test results are as described, but I can't run the test twice in a row - we are not cleaning up something properly I guess?

PR #495, opened 2023-01-24
480 better status logging

### LINKED ISSUE #480: Log *which* status is not complete
`[2023-01-21 01:05:06,649] Artemis eiger INFO: Waiting on stale parameters to go low
[2023-01-21 01:05:07,616] Artemis eiger INFO: Setting aquire
[2023-01-21 01:05:14,618] ophyd status WARNING: DeviceStatus(device=fgs_fast_grid_scan, done=False, success=False) has timed out
[2023-01-21 01:05:44,887] Artemis nexus_callback INFO: Updating Nexus file timestamps.
[2023-01-21 01:05:44,901] Artemis store_in_ispyb INFO: End ispyb deposition with status ‘fail’ and reason ‘Status has not completed yet.’.`

When something in an ophyd device times out we log the exception (last line of above). But there are many statuses which could not be completed, so we should report which status it is.

dperl-dls:
Fixes #480 

### To test:
1. Check tests haven't broken


DominicOram (approved):
Looks good, thanks.

Issue #451, opened 2022-12-15
Move motors into dodal


DominicOram:
We have had some issues with motors connecting to PVs, we should:
* Move them into `dodal`
* Instantiate them once at Artemis start

PR #496, opened 2023-01-24
#451 Creating devices before running plans

DominicOram:
Is a quick hotfix that will solve the underlying issue causing #451 but not a proper solution.

@dperl-dls and @noemifrisina, thoughts on this as a quick solution for Thursday? I think it's pretty messy and I suspect the tests need fixing too.


dperl-dls (commented):
I made a few small changes and fixed the tests (I think)
As you say it's a little messy, but I like the direction this is going, actually

dperl-dls (approved):

noemifrisina (approved):
This looks like a good start. I agree it looks a bit messy, but as long as it works that can be improved later on.

PR #504, opened 2023-01-27
(#503) Exceptions inside receive_results are now re-raised

### LINKED ISSUE #503: Correctly capture Exceptions in receive_results
Currently, if we get an exception in `receive_results` in Zocalo it is swallowed and end up just timing out on waiting for results. Instead we should capture the exception, pass it up the stack and stop waiting on the results.

DominicOram:
Fixes #503

### To test:
1. Confirm new test passes


dperl-dls (approved):
looks good, test passes, thanks

PR #511, opened 2023-01-27
Restore writing the analysis time to comments

### LINKED ISSUE #510: zocalo processing time removed from ispyb comment

dperl-dls:
Fixes #510 

to test: run new test

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,3 +4,6 @@ coverage:
       default:
         target: 85%    # the required coverage value
         threshold: 1%  # the leniency in hitting the target
+
+ignore:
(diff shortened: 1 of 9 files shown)


DominicOram (changes requested):
As discussed, can we fix this with a system test that uses the fake zocalo and reads out of the dev ispyb that a non-zero time has been written to the comment. Sorry, I know it's a pain but we should avoid regressions

dperl-dls:
test depends on the version of fake zocalo from #467

DominicOram (approved):
Good work, thank you!

PR #515, opened 2023-02-01
Improve logging to ispyb when there is no diffraction

### LINKED ISSUE #487: Log no diffraction into ispyb

dperl-dls:
Fixes #487

### To test:
1. Wait for #511  to be merged
2. Remove skip mark and run test

### DIFF
--- a/fake_zocalo/__main__.py
+++ b/fake_zocalo/__main__.py
@@ -2,6 +2,7 @@ import json
 import os
 import time
 from pathlib import Path
+from typing import Tuple
 
 import ispyb.sqlalchemy
 import pika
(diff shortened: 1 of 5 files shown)


DominicOram (changes requested):
This test doesn't work. I think it's because we don't have permission to rewrite the comment for `dcid` 1 in the dev database. It's a bit of a hack anyway as for a real run the `dcids` will be given to us arbitrarily which means trying to do a fake run with no diffraction will be hard. I think a solution to this could be for fake_zocalo to use some other indicator in the database to determine if it should respond with no diffraction. Maybe sample name?

DominicOram (approved):
Have updated to use sample name to determine diffraction

DominicOram:
I'm going to merge this as I'm happy with it and I only did minor changes to the tests compared to the work @dperl-dls did.

PR #517, opened 2023-02-06
451: Move devices into dodal

### LINKED ISSUE #451: Move motors into dodal

noemifrisina:
Fixes #451 

Move devices and their tests into the python-dodal repository.
Add dodal to dependency list.
(Also see #221 )

### To test:
1. Confirm tests still pass
2.


.pre-commit-config.yaml
```diff
@@ -1,48 +1,47 @@
 repos:
```
DominicOram, line 1: Nit: For future reference probably would have been good to do the isort update in a separate PR to avoid confusing this one.

setup.cfg
```diff
@@ -79,6 +79,7 @@ extend-ignore =
     F811,
     # line too long
     E501,
+    C408,
```
DominicOram, line 82: Should: Why do we need this ignore? Others have a comment as to why (though I appreciate they're not great comments)

noemifrisina: It was to avoid the pre-commit checks from failing when dict() or tuple() is called instead of {}, (). 

It's from one of the first commits though and from running the pre-commits again it looks like most of those went away. Right now it only fails in one instance in this file `.github/pages/make_switcher.py`, so it would be easy enough to fix and remove the ignore from setup.cfg.

.pre-commit-config.yaml
noemifrisina: Sorry I was mostly just trying to get this to stop screaming and didn't really think about that. Want me to open a separate PR for this so it can go into main while this is still a WIP?

DominicOram: No, no. It's fine, just an FYI for future tickets. Better to have lots of small self contained PRs than one bigger one

DominicOram (approved):
Looks good, let's go for it

PR #519, opened 2023-02-09
Tell Zocalo to no longer wait on run status

### LINKED ISSUE #518: Fix runStatus to correctly start zocalo
In the latest release there is a bug where we trigger zocalo before we have written a `runStatus` into the database. Zocalo expects the `runStatus` to be successful at the end of analysis and so this fails.

## Acceptance Criteria
* We remove this check

DominicOram:
Fixes #518 

I can't think of a brilliant way of testing this other than implementing the wait for run_status in fake_zocalo but I really don't want fake_zocalo to grow that much, particularly as we don't actually want the wait functionality.

### To test:
1. Run test

### DIFF
--- a/src/artemis/external_interaction/zocalo/zocalo_interaction.py
+++ b/src/artemis/external_interaction/zocalo/zocalo_interaction.py
@@ -76,7 +76,6 @@ class ZocaloInteractor:
         self._send_to_zocalo(
             {
                 "event": "end",
-                "ispyb_wait_for_runstatus": "1",
                 "ispyb_dcid": data_collection_id,
             }
         )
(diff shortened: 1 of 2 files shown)


src/artemis/external_interaction/zocalo/zocalo_interaction.py
```diff
@@ -76,7 +76,6 @@ def run_end(self, data_collection_id: int):
         self._send_to_zocalo(
             {
                 "event": "end",
-                "ispyb_wait_for_runstatus": "1",
```
rjgildea, line 79: It's probably worth adding in a comment (at least for now) to indicate that this is different behaviour compared to non-artemis MX data collections.

DominicOram: Sorry, I didn't mean that to sound as confrontational as it did

dperl-dls (approved):
seems good, test passes, cheers

PR #521, opened 2023-02-14
486 choose aperture based on xtal size

### LINKED ISSUE #486: Chose aperture based on crystal size
Given richer data from zocalo we should be able to chose apertures e.g. a bigger crystal should use a bigger aperture. We just need the infrastructure for this, how to do the choosing should be discussed with scientists.

dperl-dls:
Fixes #486

### To test:
Run tests

### DIFF
--- a/src/artemis/experiment_plans/fast_grid_scan_plan.py
+++ b/src/artemis/experiment_plans/fast_grid_scan_plan.py
@@ -190,7 +190,7 @@ def run_gridscan_and_move(
     # the data were submitted to zocalo by the zocalo callback during the gridscan,
     # but results may not be ready, and need to be collected regardless.
     # it might not be ideal to block for this, see #327
-    xray_centre = subscriptions.zocalo_handler.wait_for_results(initial_xyz)
+    xray_centre, bbox_size = subscriptions.zocalo_handler.wait_for_results(initial_xyz)
 
     # once we have the results, go to the appropriate position
     artemis.log.LOGGER.info("Moving to centre of mass.")
--- a/src/artemis/experiment_plans/fast_grid_scan_plan.py
+++ b/src/artemis/experiment_plans/fast_grid_scan_plan.py
@@ -11,7 +11,7 @@ from artemis.device_setup_plans.setup_zebra_for_fgs import (
     set_zebra_shutter_to_manual,
     setup_zebra_for_fgs,
 )
-from artemis.devices.aperturescatterguard import ApertureScatterguard, ApertureSize
+from artemis.devices.aperturescatterguard import ApertureScatterguard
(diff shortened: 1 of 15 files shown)


src/artemis/external_interaction/callbacks/fgs/zocalo_callback.py
```diff
@@ -87,25 +88,52 @@ def wait_for_results(self, fallback_xyz: Point3D) -> Point3D:
                 datacollection_group_id
             )
 
+            # Example result with two crystals:
```
noemifrisina, line 91: Could you remove this comment if not needed? Since you already have a similar example in the docstring of wait_for_results() in `zocal_interaction.py`, it feels a bit redundant

noemifrisina:
Seems good, all tests pass. Only needs the merge conflicts to be resolved.

noemifrisina (approved):

DominicOram:
Sorry to jump on this after it's been approved but I think it's really important we have some tests around `safe_move_within_datacollection_range` given how it's machine safety critical. I think we already have an aperture and scatterguard in s03 so adding system tests for it (and removing the patching on the general system tests) shouldn't be that bad?

src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -169,7 +221,11 @@ def gridscan_with_subscriptions(fgs_composite, detector, params):
     # the data were submitted to zocalo by the zocalo callback during the gridscan,
     # but results may not be ready, and need to be collected regardless.
     # it might not be ideal to block for this, see #327
-    xray_centre = subscriptions.zocalo_handler.wait_for_results(initial_xyz)
+    xray_centre, bbox_size = subscriptions.zocalo_handler.wait_for_results(initial_xyz)
+
+    if bbox_size is not None:
+        with TRACER.start_span("change_aperture"):
+            set_aperture_for_bbox_size(fgs_composite.aperture_scatterguard, bbox_size)
```
DominicOram, line 228: As discussed, I don't think this is the idiomatic way of doing things as we're calling a function on the device during a plan. I think the better way is to create a signal on the device (that doesn't point to PVs) that we can write/read/wait on in the plan then the device will watch this signal and do the complex logic based on it. @callumforrester, does that sound correct to you?

callumforrester: Why not just put the logic in `ApertureScatterguard.set`?, Then you can just go `yield from mv(aperture_device, aperture_device.LARGE)`, for example

DominicOram: Oh yh, that's much cleaner!

PR #424, opened 2022-12-09
366 refactor parameters

dperl-dls:
Contributes to things needed for #366 by separating external and internal parameter models
Externally supplied params are validated against a json schema
For now they still look exactly the same, but we can now start modifying the internals

**Requires small changes to GDA so that the json matches the schema**

### To test:
1. Run tests

PR #522, opened 2023-02-14
366 refactor parameters combineregistry

### LINKED ISSUE #366: Refactor parameters.py to support different experiment/scan types
Since only `grid_scan_params` would change for e.g. `rotation_scan_params` I think we should extent the current FullParameters for this, but another option is to have separate classes, which might also have advantages?

dperl-dls:
Fixes #366 
See: #424


DominicOram (approved):
This is great and good docs too, thank you! I've done some minor tidying, can you fix the conflicts then we can merge it?

DominicOram:
Created https://jira.diamond.ac.uk/browse/I03-866 to do the GDA side work

PR #529, opened 2023-02-16
Adds script to make it easier to test when there is no beam

### LINKED ISSUE #523: Create script to trick GDA into working with beam off
Currently GDA will not allow you to trigger Artemis w/o beam as the XBPM feedback fails to get a beam position. This can be worked around by modifying the threshold to some very large value.

## Acceptance Criteria
* There is a helper script for this
* There is some docs about it

DominicOram:
Fixes #523

See also https://github.com/DiamondLightSource/python-artemis/wiki/Testing-on-the-Beamline

### To test:
1. Replace line 4 of the script with `trap "echo 'caput BL03I-EA-FDBK-01:THRESHOLDPC_XBPM2 $INIT_VALUE' > test_beam_off" EXIT HUP`
3. Run the script
4. Kill it manually, confirm `test_beam_off` contains the expected output
5. Remove `test_beam_off`
6. Re-run the script
7. Log off, confirm `test_beam_off` contains the expected output


dperl-dls (approved):
works fine for me

DominicOram:
Hmm... Wasn't working for @noemifrisina though, I might merge it then we can play with it on the beamline but keep an eye on it

noemifrisina:
Sorry to jump in after merging, just wanted to give a small update. Tried it again this morning and while not immediate after killing the process, got a `test_beam_off` file with the expected value. Not sure yet what was wrong when I tested before - might have been something running on my machine - so keeping an eye on it sounds like a good idea.

PR #532, opened 2023-02-17
Bump isort version

noemifrisina:
Update isort 

### To test:
1. Check that all pre-commits run fine


DominicOram (approved):
Works for me, thanks!

Issue #528, opened 2023-02-16
Isolate aperture device logic from plan


dperl-dls:
Currently `fast_grid_scan_plan` directly calls aperture device functions in:

```python
def set_aperture_for_bbox_size(
    aperture_device: ApertureScatterguard,
    bbox_size: list[int],
):
    ...
    aperture_device.safe_move_within_datacollection_range(*aperture_size_positions)
```

Instead we should extend `ApertureScatterguard` with a Signal to handle this, so that we can instead `yield from bps.abs_set(aperture,position)`, or similar, in the plan

PR #534, opened 2023-02-17
527 528 improve and test aperture

### LINKED ISSUE #527: Need tests for aperture changing
#521 was merged without sufficient system tests to ensure the safety of the movements, we need some system tests for this

dperl-dls:
Fixes #527 and #528 

### To test:
1. Run tests against S03


DominicOram (approved):
Great work, thanks! As discussed I think a quick diagram would be good but I can do that myself, let's get this on the beamline first.

PR #546, opened 2023-02-23
524 setup zebra for rotation

### LINKED ISSUE #524: Set up the zebra for PC in the way required for a rotation scan
**Acceptance criteria:**
artemis.device_setup_plans has a plan which takes input parameters as needed for a rotation scan: angle of interest (phi/chi/omega), total scan width (deg), image width (deg), exposure time per image (s)
this plan should set up the zebra such that when the smargon is moved appropriately, the zebra triggers data collection as it should. 
these don't need to be programatically linked - i.e. this doesn't have to set up the eiger

dperl-dls:
Fixes #524 
Adds a plan which sets up the zebra for rotation data collection. Has been tested on the beamline.

### To test:
1. Run with S03 and check that the correct PVs are set to the correct values

### DIFF
--- a/src/artemis/device_setup_plans/setup_zebra_for_rotation.py
+++ b/src/artemis/device_setup_plans/setup_zebra_for_rotation.py
@@ -1,0 +1,29 @@
+import bluesky.plan_stubs as bps
+
+from artemis.devices.zebra import (
+    DISCONNECT,
+    OR1,
+    PC_PULSE,
+    TTL_DETECTOR,
+    TTL_SHUTTER,
+    TTL_XSPRESS3,
+    Zebra,
+)
+
+
+def setup_zebra_for_rotation(
+    zebra: Zebra, group="setup_zebra_for_rotation", wait=False
+):
+    # Need to:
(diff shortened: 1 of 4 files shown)


src/artemis/device_setup_plans/setup_zebra_for_rotation.py
```diff
@@ -0,0 +1,50 @@
...
+    """Set up the Zebra to collect a rotation dataset. Any plan using this is
+    responsible for setting the smargon velocity appropriately so that the desired
+    image width is achieved with the exposure time given here.
+
+    Parameters:
+        axis:           I03 axes enum representing which axis to use for position
+                        compare. Currently always omega.
+        start_angle:    Position at which the scan should begin, in degrees.
+        scan_width:     Total angle through which to collect, in degrees.
+    """
```
DominicOram, line 32: Great documentation here, thanks!

DominicOram (approved):
Looks good, thanks. I have added the suggested test as an automated system test.

PR #548, opened 2023-02-23
Zocalo will now timeout if no results are recieved - Additionally tidies up some tests

### LINKED ISSUE #545: Fix/tidy tests
There are a few minor issues with the system tests, we should fix/tidy them:
* Eiger test is failing `FAILED src/artemis/devices/unit_tests/test_eiger.py::test_when_set_odin_pvs_called_then_full_filename_written - AssertionError: assert '0' == 'test_0'`
* When fake zocalo is not running the system tests appear to hang as the timeout is very large
* The aperture/scattergaurd system tests are annoyingly slow

DominicOram:
Fixes #545

### To test:
1. Run system tests without fake_zocalo running
2. Confirm it times out after ~5 secs
3. Confirm `test_when_set_odin_pvs_called_then_full_filename_written` is less flaky
4. Confirm aperture/scatterguard tests are faster (can do this by running just these tests on main and here, I see ~60 sec speedup)

### DIFF
--- a/src/artemis/devices/system_tests/test_aperturescatterguard_system.py
+++ b/src/artemis/devices/system_tests/test_aperturescatterguard_system.py
@@ -130,8 +130,6 @@ def test_aperturescatterguard_moves_in_correct_order(
     RE.subscribe(cb)
 
     ap_sg.wait_for_connection()
-    ap_sg.aperture.y.set(0, wait=True)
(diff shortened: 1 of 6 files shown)


dperl-dls (approved):
looks much nicer, tests all pass

PR #552, opened 2023-02-27
550 print parameters better

### LINKED ISSUE #550: add __repr__ to InternalParameters
- So that the contents get logged, not just `<InternalParameters object at blah blah>`
- also consider printing external params in that log message

dperl-dls:
Fixes #550

### To test:
1. make an InternalParameters object and print it
2. check that this representation appears in logs


DominicOram (approved):
Tested by running Artemis in dev mode, sending a request and looking in local logs. Confirmed logs are now sensible. Thanks!

PR #553, opened 2023-02-28
Add a test to confirm order of zocalo end call vs disarming detector

### LINKED ISSUE #536: Automate a test for 535

DominicOram:
Fixes #536 

### To test:
1. Run test and confirm it passes
2. Confirm if you revert https://github.com/DiamondLightSource/python-artemis/pull/537/commits/44b515782fa4431df6ed2f881982ff73daa8c870


noemifrisina (approved):
Looks okay to me

PR #555, opened 2023-03-02
Collection of minor issues found from testing v2.0.0

### LINKED ISSUE #551: issues from testing 2.0.0
- setup is called for every experiment type, leading to a NotImplemented error on main
- upper left and position interpreted as dict then converted to point3D - was incorrect - should fix and test for correct values
- check up on double conversion when those values go from external -> internal
- The positions of each crystal is specified to too many decimal places

DominicOram:
Fixes #551

### To test:
1. Confirm tests still pass


dperl-dls (approved):
great

PR #562, opened 2023-03-06
Make sure fake zocalo, used for developer testing, is tidied up well

### LINKED ISSUE #547: fake_zocalo should tidy itself up at the end
To make `fake_zocalo` more robust we should do the following on exit:
* Stop the Rabbit MQ instance that the script started
* Remove the credentials

DominicOram:
Fixes #547

### To test:
1. Start then kill fake zocalo
2. Confirm `/home/$USER/.zocalo/` is cleaned
3. Start fake zocalo again (note will need a few seconds before stopping it) and confirm a new rabbit mq instance is created


noemifrisina:
Clean up works when exiting the terminal, but it doesn't seem to do anything when simply killing process with `Ctrl+C`

dperl-dls:
works for me

dperl-dls (approved):
I guess this usually works, I'm happy with merging it

DominicOram:
I'm gonna spend a bit of time debugging for @noemifrisina before merging but thanks for checking @dperl-dls

DominicOram:
After some testing it looks like for some reason the `trap` command doesn't work for @noemifrisina, I will instead do the cleanup in the keyboard interrupt exception in the python code

DominicOram:
@noemifrisina to look into the `trap` issue on their machine further. If you have no luck, I'll do the above

noemifrisina (approved):
Still not 100% sure which process was creating issues but after rebooting `trap` now seems to work fine.

PR #565, opened 2023-03-07
542 send aperture back to gda

### LINKED ISSUE #542: Send chosen aperture back to GDA
GDA holds the selected aperture in a dummy PV. If Artemis changes the hardware it is not picked up by GDA and so it will change it back. Instead Artemis should let GDA know what the change of aperture is and GDA can then set it.

## Acceptance Criteria
* Artemis sends back the new aperture to GDA
* GDA will change it's internal state to match the new aperture

DominicOram:
Fixes #542 

Also review GDA changes in https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/38342

### To test:
1. Run tests - in particular the new system test in `test_aperturescatterguard_system`


dperl-dls:
closed for https://github.com/DiamondLightSource/python-artemis/pull/613 so @DominicOram can review

PR #581, opened 2023-03-14
Force us to use a local version of dodal in dev

DominicOram:
### To test:
1. Run `dls_dev_env.sh`
2. Ensure local dodal is used


dperl-dls (approved):
much better

Issue #25, opened 2022-02-04
Eiger setting filenames incorrectly


DominicOram:
Currently staging the eiger is setting filenames incorrectly producing:
```
BL03S-EA-EIGER-01:OD:AcquisitionID 2022-02-03 13:15:37.829977 file.h5  
BL03S-EA-EIGER-01:OD:AcquisitionID 2022-02-03 13:17:08.007233 001  
BL03S-EA-EIGER-01:OD:AcquisitionID 2022-02-03 13:17:08.053712 file.h5
```
See https://github.com/DiamondLightSource/python-artemis/pull/23#issuecomment-1029060653 for more details. Note that the `.h5` issue has been fixed already

## Acceptance Criteria
* The filename PVs do not get set multiple times
* There is a system test covering this

PR #584, opened 2023-03-16
554 tidy oav

### LINKED ISSUE #554: Tidy up OAV devices and grid plan
Some OAV device namings are confusing, e.g.:

- [x] MJPG is called "snapshot"
- [x] inconsistent use of `input_plugin_pv` and `input_pv` (maybe areadetector plugins should be a class...)
- [x] chain plugins generically or at least in a consistent way
- [ ] structure of snapshot/snapshotwithgrid etc needs cleaning
- [x] `top_left_x_signal` etc. in snapshot device classes - add clarity that this refers to grid and remove redundant "signal"
- [x] take arguments for grid size, margin, # of boxes (eventually to come from json)
- [ ] deal with grid with no pin at zoom 0
- [ ] maybe average edge detection for stability, also maybe make sure rounding is correct so grid always extends to the edge of the sample
- [x] add backlight manipulation
- [x] `oav_centring_plan` should be in `experiment_plans` rather than `device_setup_plans`

also:
- [ ] look at bluesky visualisation options for live debugging

d-perl:
Fixes DiamondLightSource/hyperion#554 

Link to dodal PR (if required): [#25](https://github.com/DiamondLightSource/dodal/pull/25)

### To test:
1. Check organisation/naming of oav code is sensible/coherent
2. Run tests with both artemis and dodal branches


src/artemis/experiment_plans/full_grid_scan.py
```diff
@@ -0,0 +1,110 @@
...
+    gda_snap_1 = parameters.artemis_params.ispyb_params.xtal_snapshots_omega_start[0]
+    gda_snap_2 = parameters.artemis_params.ispyb_params.xtal_snapshots_omega_end[0]
+
+    snapshot_dir = os.path.dirname(os.path.abspath(gda_snap_1))
+    snap_1_filename = os.path.basename(os.path.abspath(gda_snap_1))
+    snap_2_filename = os.path.basename(os.path.abspath(gda_snap_2))
+
+    zoom_params_file = "/dls_sw/i03/software/gda_versions/gda_9_27/workspace_git/gda-mx.git/configurations/i03-config/xml/jCameraManZoomLevels.xml"
+    oav_json = "/dls_sw/i03/software/gda_versions/gda_9_27/workspace_git/gda-mx.git/configurations/i03-config/etc/OAVCentring.json"
+    display_config = "/dls_sw/i03/software/gda_versions/var/display.configuration"
```
DominicOram, line 79: Should: These hardcoded paths need removing, they should come in as params

src/artemis/experiment_plans/oav_grid_detection_plan.py
```diff
@@ -0,0 +1,187 @@
...
+        rotation_points (int): Test to see if the pin is widest `rotation_points` number of times on a full 180 degree rotation.
+    """
+
+    LOGGER.info("OAV Centring: Starting loop centring")
+
+    # parameters = OAVParameters(
+    #     parameters.experiment_params.centring_params_file,
+    #     parameters.experiment_params.camera_zoom_levels_file,
+    #     parameters.experiment_params.display_configuration_file,
+    # )
```
DominicOram, line 50: Should: Remove commented out code

src/artemis/experiment_plans/full_grid_scan.py
```diff
@@ -0,0 +1,110 @@
...
+    times_to_check = int(timeout / SLEEP_PER_CHECK)
+    for _ in range(times_to_check):
+        detector_state = yield from bps.rd(detector.shutter)
+        detector_z_dmov = yield from bps.rd(detector.z.motor_done_move)
+        LOGGER.info(f"Shutter state is {'open' if detector_state==1 else 'closed'}")
+        LOGGER.info(f"Detector z DMOV is {detector_z_dmov}")
+        if detector_state == 1 and detector_z_dmov == 1:
+            return
+        yield from bps.sleep(SLEEP_PER_CHECK)
+    raise Exception("Detector not finished moving")
```
DominicOram, line 59: Should: This is much like the waiting for fast grid scan to be valid, e.g. messy and horrible. I think we need to put it in the device side probably? Certainly refactor it to be more Bluesky idiomatic

DominicOram (changes requested):
Thanks for making a start on this but I think it still needs more cleaning up from prototyping. I think to merge it we need at least a unit test against some fake devices to confirm the plan runs through. Can you also address all the `must`s.

Can you write up issues for the other comments if you don't address them here? I also think this doesn't address all of the things on the issue it links. I have ticked the ones I think it fixes, can you split the others out into new issues?

src/artemis/utils/oav_utils.py
```diff
@@ -0,0 +1,19 @@
+import bluesky.plan_stubs as bps
+from dodal.devices.oav.oav_detector import OAV
+
+
+def get_waveforms_to_image_scale(oav: OAV):
+    """
+    Returns the scale of the image. The standard calculation for the image is based
+    on a size of (1024, 768) so we require these scaling factors.
```
DominicOram, line 8: Should: This comment is not correct in that we don't use the scaling factors

src/artemis/device_setup_plans/setup_oav.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # Connect MXSC output to MJPG input
+    yield from start_mxsc(
+        oav,
+        parameters.input_plugin + "." + parameters.mxsc_input,
+        parameters.min_callback_time,
+        parameters.filename,
+    )
+
+    yield from bps.abs_set(oav.snapshot.input_pv, parameters.input_plugin + ".CAM")
```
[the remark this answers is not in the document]
dperl-dls: renamed all of them to input_plugin

dperl-dls:
Fair enough, in that case, I'd like to get 561_558 merged first, since changes there will affect the work that needs to be done here quite a lot

src/artemis/experiment_plans/oav_grid_detection_plan.py
```diff
@@ -0,0 +1,187 @@
...
+        yield from bps.abs_set(oav.snapshot.top_left_x, upper_left[0])
+        yield from bps.abs_set(oav.snapshot.top_left_y, upper_left[1])
+        yield from bps.abs_set(oav.snapshot.box_width, box_size_x_pixels)
+        yield from bps.abs_set(oav.snapshot.num_boxes_x, boxes[0])
+        yield from bps.abs_set(oav.snapshot.num_boxes_y, boxes[1])
+
+        LOGGER.info("Triggering snapshot")
+
+        snapshot_filename = (
+            filenames["snap_1_filename"] if angle == 0 else filenames["snap_2_filename"]
```
DominicOram, line 124: Will be addressed in https://github.com/DiamondLightSource/python-artemis/issues/651

DominicOram (changes requested):
Looks good, thanks. I tried to spin out as much as possible into other issues, what's left is things we can hopefully quickly fix now

src/artemis/experiment_plans/full_grid_scan.py
```diff
@@ -74,10 +75,10 @@ def get_plan(
         "snap_2_filename": os.path.basename(os.path.abspath(gda_snap_2)),
     }
 
-    oav_params = OAVParameters("xrayCentring")
+    oav_params = OAVParameters("loopCentring", **oav_param_files)
```
DominicOram, line 78: Should: I think there is an issue using `loopCentring` here, spun out into https://github.com/DiamondLightSource/python-artemis/issues/653

DominicOram (approved):
Great, thanks for the additional tests

PR #591, opened 2023-03-20
(#571) Update pixels_per_micron to microns_per_pixel

### LINKED ISSUE #571: pixels per micron in parameters is actually microns per pixel
when gda adds `pixels_per_micron` to the params it sends to artemis it is using a property which is `MicronsPerPixel`

NeilSmithDLS:
Fixes #571

### To test:
1. Run the automated unit tests and ensure they still work


dperl-dls (approved):
Looks good, I don't see any other references to pixels per micron, the OAV code in DODAL is already in line with this change, and no tests are broken

Issue #31, opened 2022-02-16
Investigate bluesky service


DominicOram:
The current infrastructure for starting/stopping scans from GDA (done in https://github.com/DiamondLightSource/python-artemis/pull/28) works OK but is not very extensible. The core team have already created https://gitlab.diamond.ac.uk/daq/d2acq/services/bluesky-service that performs a similar task. We should evaluate whether it could be used on Artemis.

## Acceptance Criteria
* There is a branch where the bluesky-service is integrated with Artemis
* There are some notes about the advantages/disadvantages of using it over the current model
* A decision is made about whether we use this service

DominicOram:
Initial thoughts:

Advantages:
* We stress test bluesky-service (which is likely to be what is used on future bluesky projects)
* We will be forced to write some GDA code to communicate with the bluesky-service (again useful for future)

Disadvantages:
* The bluesky service contains more complexity than we (currently) need for Artemis

DominicOram:
@callumforrester is working on a cut down version of the Bluesky service which will work better for us

evalott100:
The gitlab repo above is an old prototype,  https://github.com/DiamondLightSource/blueapi/milestone/15, is the new repo for the bluesky service.

DominicOram:
Clearing assignee from this as we're waiting on blueapi to be ready

DominicOram:
@callumforrester has made a start on this but it needs a Hyperion dev to finish it off

DominicOram:
I think we can call this investigation done with the conclusion that `blueapi` is now in a good enough state that we can start using it, thanks to @callumforrester. Subsequent issues to integrate properly in due course

PR #595, opened 2023-03-27
561 588 improve dodal and its use in artemis

### LINKED ISSUE #561: Use dodal more idomatically
Currently we import the devices from dodal and then create them locally rather than using the `i03.py` file in dodal. We should switch to using this.

## Acceptance Criteria
* All device creation in Artemis is done through the functions in `i03.py`
* We should be using `s03.py` in the tests

### LINKED ISSUE #588: Improve creation of devices - prevent duplicates
Now that we have several plans using the same devices, they are being created and connection waited for several times at the beginning of artemis __main__ - see `test_when_blueskyrunner_initiated_then_plans_are_setup_and_devices_connected` in `test_main_system`

we should fix this in main which, in init, instead of:
```python
        for plan in PLAN_REGISTRY:
            PLAN_REGISTRY[plan]["setup"]()
```
should collect a list of devices to initialise, ignoring duplicates, and then do them all once

and fix `test_when_blueskyrunner_initiated_then_plans_are_setup_and_devices_connected` to test devices once

dperl-dls:
Fixes #561 and fixes #588 

Gets rid of FGSComposite in Dodal and reimplements it as a simple holder class in fast_grid_scan_plan with some config options
Dodal i03.py now manages device instantiation and this functionality is used in Artemis

Link to dodal PR (if required): [#31](https://github.com/DiamondLightSource/dodal/pull/31)

### To test:
1. Run tests


src/artemis/system_tests/test_fgs_plan.py
```diff
@@ -79,12 +42,24 @@ def RE():
...
 @pytest.fixture
 def fgs_composite():
-    fast_grid_scan_composite = FGSComposite(
-        insertion_prefix=SIM_INSERTION_PREFIX,
-        name="fgs",
-        prefix=SIM_BEAMLINE,
-    )
-    fast_grid_scan_composite.wait_for_connection()
+    with patch("dodal.i03.dcm"):
+        with patch("dodal.i03.oav"):
```
DominicOram, line 46: Nit: Maybe add a comment here that we're patching it cos there's no simulator. We should remove the patch once we have merged https://github.com/dls-controls/tickit-devices/pull/27

src/artemis/system_tests/test_main_system.py
```diff
@@ -200,12 +200,41 @@ def test_cli_args_parse():
...
+    undulator.return_value.wait_for_connection.assert_called_once()
+    synchrotron.return_value.wait_for_connection.assert_called_once()
+    smargon.return_value.wait_for_connection.assert_called_once()
+    s4_slits.return_value.wait_for_connection.assert_called_once()
+    oav.return_value.wait_for_connection.assert_called_once()
+    fast_grid_scan.return_value.wait_for_connection.assert_called_once()
+    eiger.return_value.wait_for_connection.assert_not_called()  # can't wait on eiger
+    dcm.return_value.wait_for_connection.assert_called_once()
+    backlight.return_value.wait_for_connection.assert_called_once()
+    aperture_scatterguard.return_value.wait_for_connection.assert_called_once()
```
DominicOram, line 240: Nit: Could we just patch `dodal.i03` then use `mock_i03.zebra.return_value.wait_for_connection.assert_called_once()` etc. in the test, reduces complexity a bit

DominicOram (changes requested):
Looks much nicer, particularly with the eiger moved around, thanks. Minor comments in code

src/artemis/system_tests/test_main_system.py
dperl-dls: if you mock dodal.i03 then the function which makes the devices doesn't exist

src/artemis/system_tests/test_fgs_plan.py
dperl-dls: removed this since they are removed from fgs composite anyway

DominicOram (approved):
Looks good, thank you!

PR #602, opened 2023-04-03
Adds flag to skip startup connection scripts

### LINKED ISSUE #600: Be able to start artemis in a mode that does not immediately connect to hardware
Currently on startup Artemis will run all the plan setups and so attempt to connect to hardware. This fails if there's no simulator running, we shouldn't necessarily need the simulator to start up and run some dummy plans.

## Acceptance Criteria
* We can pass a flag to artemis to not call setup scripts on plans at startup
* Setup scripts are still called when the plan runs
* There are docs/tests for this

olliesilvester:
Fixes #600

### To test:
1. Run new unit tests in test_main_system.py
2. Confirm artemis doesn't automatically connect to hardwere when running with the flag --skip_startup_connection, and that it does without this flag

### DIFF
--- a/run_artemis.sh
+++ b/run_artemis.sh
@@ -34,6 +34,8 @@ checkver () {
 STOP=0
 START=1
 DEPLOY=0
+SKIP_STARTUP_CONNECTION=0
+
 for option in "$@"; do
     case $option in
         -b=*|--beamline=*)
@@ -52,6 +54,9 @@ for option in "$@"; do
             ;;
         --deploy)
             DEPLOY=1
+            ;;
+        --skip_startup_connection)
+            SKIP_STARTUP_CONNECTION=1
             ;;
         --help|--info)
             echo "Options"
@@ -163,7 +168,7 @@ if [[ $START == 1 ]]; then
 
     source .venv/bin/activate
 
-    python -m artemis `if [ $IN_DEV == true ]; then echo "--dev"; fi` >$start_log_path 2>&1 &
+    python -m artemis `if [ $IN_DEV == true ]; then echo "--dev"; fi` `if [ $SKIP_STARTUP_CONNECTION == 1 ]; then echo "--skip_startup_connection"; fi`>$start_log_path 2>&1 &
 
     echo "Waiting for Artemis to boot"
 
--- a/src/artemis/__main__.py
+++ b/src/artemis/__main__.py
@@ -55,13 +55,20 @@ class BlueskyRunner:
         self.RE = RE
         if VERBOSE_EVENT_LOGGING:
             RE.subscribe(VerbosePlanExecutionLoggingCallback())
-        for plan in PLAN_REGISTRY:
-            PLAN_REGISTRY[plan]["setup"]()
+
+        if not skip_startup_connection:
+            for plan in PLAN_REGISTRY:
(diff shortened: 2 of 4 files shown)


run_artemis.sh
```diff
@@ -163,7 +168,7 @@ if [[ $START == 1 ]]; then
 
     source .venv/bin/activate
 
-    python -m artemis `if [ $IN_DEV == true ]; then echo "--dev"; fi` >$start_log_path 2>&1 &
+    python -m artemis `if [ $IN_DEV == true ]; then echo "--dev"; fi` `if [ $SKIP_STARTUP_CONNECTION == 1 ]; then echo "--skip_startup_connection"; fi`>$start_log_path 2>&1 &
```
DominicOram, line 171: This is ok for now but FYI I have added https://github.com/DiamondLightSource/python-artemis/issues/603 to tidy this

src/artemis/__main__.py
```diff
@@ -51,17 +51,25 @@ class BlueskyRunner:
...
         self.RE = RE
+        self.skip_startup_connection = skip_startup_connection
         if VERBOSE_EVENT_LOGGING:
             RE.subscribe(VerbosePlanExecutionLoggingCallback())
-        for plan in PLAN_REGISTRY:
-            PLAN_REGISTRY[plan]["setup"]()
+
+        if not self.skip_startup_connection:
+            for plan in PLAN_REGISTRY:
+                PLAN_REGISTRY[plan]["setup"]()
```
DominicOram, line 62: Nit: Might be a little cleaner to pull these two lines into a function to call here and in `start` e.g. `setup_all_plans`

DominicOram (changes requested):
Good work, thank you! Minor comments but otherwise good

src/artemis/__main__.py
```diff
@@ -51,17 +51,25 @@ class BlueskyRunner:
...
 
     def start(
         self, experiment: Callable, parameters: InternalParameters
     ) -> StatusAndMessage:
         artemis.log.LOGGER.info(f"Started with parameters: {parameters}")
+
+        if self.skip_startup_connection:
+            for plan in PLAN_REGISTRY:
```
dperl-dls, line 70: I would prefer if this only called setup() for the specified plan, rather than all of them

DominicOram: Good spot

dperl-dls: yeah, if we start it running without the flag, we want it to be ready to run any plan we send to it. but if we don't want it to setup the devices on startup, it would be better if it only runs the setup for each plan as needed. that way, for example, if the simulator has some but not all the devices, we can still run plans that has devices the simulator does have, while not trying to connect to ones it doesn't have.

src/artemis/__main__.py
```diff
@@ -51,17 +51,25 @@ class BlueskyRunner:
...
+                PLAN_REGISTRY[plan]["setup"]()
```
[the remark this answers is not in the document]
olliesilvester: Added tests for this now

src/artemis/__main__.py
olliesilvester: Makes sense - new version does this now

dperl-dls (approved):
Great, looks good to me

PR #609, opened 2023-04-06
Added error message and logging for invalid json parameters, and tests

### LINKED ISSUE #596: Log and return something nicer to GDA on bad JSON params
If you send a bad set of parameters to Artemis (e.g. `curl -X PUT http://127.0.0.1:5005/fast_grid_scan/start -H "Content-Type: application/json" --data "{\"bad\":1}"` it will currently send you back a nicely formatted web page with the error and not log.

## Acceptance Criteria
On sending a bad set of params to Artemis:
* There is a sensible return error
* An error is logged to `./tmp/dev/artemis.txt`

There should also be tests on this

olliesilvester:
Fixes #596

### To test:
1. Run new tests in test_main_system.py
2. Run `curl -X PUT http://127.0.0.1:5005/fast_grid_scan/start -H "Content-Type: application/json" --data "{\"bad\":1}" `and confirm error message is printed as well as logged in tmp/dev/artemis.txt

### DIFF
--- a/src/artemis/parameters/external_parameters.py
+++ b/src/artemis/parameters/external_parameters.py
@@ -8,8 +8,10 @@ from typing import Dict, Optional, Type, Union
 
 import jsonschema
 from dataclasses_json import DataClassJsonMixin
+from jsonschema.exceptions import ValidationError
 
 import artemis.experiment_plans.experiment_registry as registry
+import artemis.log
 from artemis.parameters.constants import (
     PARAMETER_VERSION,
     SIM_BEAMLINE,
@@ -169,8 +171,12 @@ class RawParameters:
             base_uri=f"{path.as_uri()}/",
             referrer=True,
         )
-        # TODO improve failed validation error messages
-        jsonschema.validate(dict_params, full_schema, resolver=resolver)
+        try:
+            jsonschema.validate(dict_params, full_schema, resolver=resolver)
+        except ValidationError:
+            artemis.log.LOGGER.error("Invalid json parameters")
+            raise ValidationError("Invalid Json parameters")
+
         experiment_type = EXTERNAL_EXPERIMENT_PARAM_DICT.get(
             dict_params["artemis_params"]["experiment_type"]
         )
--- a/src/artemis/system_tests/test_main_system.py
+++ b/src/artemis/system_tests/test_main_system.py
@@ -8,6 +8,7 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 from flask.testing import FlaskClient
+from jsonschema.exceptions import ValidationError
 
 from artemis.__main__ import Actions, BlueskyRunner, Status, cli_arg_parse, create_app
 from artemis.experiment_plans.experiment_registry import PLAN_REGISTRY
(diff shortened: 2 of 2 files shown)


src/artemis/parameters/external_parameters.py
```diff
@@ -169,8 +171,12 @@ def from_dict(cls, dict_params: dict[str, dict]):
             base_uri=f"{path.as_uri()}/",
             referrer=True,
         )
-        # TODO improve failed validation error messages
-        jsonschema.validate(dict_params, full_schema, resolver=resolver)
+        try:
+            jsonschema.validate(dict_params, full_schema, resolver=resolver)
+        except ValidationError:
+            artemis.log.LOGGER.error("Invalid json parameters")
+            raise ValidationError("Invalid Json parameters")
```
dperl-dls, line 178: should: It's nicer to just use `raise` here, which will reraise the original exception including the stack trace and the original place the validation failed.

dperl-dls (changes requested):
Looks good, but I think we also want to capture any `Exception` at the end of trying to process input, and log it with the stack trace. Currently, the error is not handled in `__main__.py` (around lines 156-159) and so it is passed on to the flask app - we want to catch the error there so that we don't get all that html in the log. Eg if you run `curl -X PUT http://127.0.0.1:5005/fast_grid_scan/start -d '{"bad":"data"}' -H "Content-Type: application/json"` against running artemis you'll get the unwanted output in the console.

src/artemis/parameters/external_parameters.py
olliesilvester: Actually it looks like that validate function does the same error handling anyway, so I've removed this part

src/artemis/__main__.py
```diff
@@ -148,6 +149,10 @@ def put(self, experiment: str, action: Actions):
                 status_and_message = StatusAndMessage(Status.FAILED, repr(e))
             except PlanNotFound as e:
                 status_and_message = StatusAndMessage(Status.FAILED, repr(e))
+            except ValidationError as e:
+                status_and_message = StatusAndMessage(Status.FAILED, repr(e))
+                artemis.log.LOGGER.error("Invalid json parameters")
+
```
dperl-dls, line 155: yeah so like I said, I think it would be better to have an `except Exception ...` at the end here so that we never have the flask app handling exceptions. Otherwise, the next time there's a differnt error that's not PlanNotFound or ValidationError we will have the same issue again. Also, I would like it if we include the exception stacktrace in the log message

src/artemis/system_tests/test_main_system.py
```diff
@@ -209,3 +209,14 @@ def test_when_blueskyrunner_initiated_then_plans_are_setup_and_devices_connected
     BlueskyRunner(MagicMock())
 
     mock_fgs.return_value.wait_for_connection.assert_called_once()
+
+
```
dperl-dls, line 213: nice, thanks for adding the extra test

dperl-dls (changes requested):

src/artemis/__main__.py
olliesilvester: OK, I think this has now been done

dperl-dls (approved):
Looks good, thanks!

PR #611, opened 2023-04-13
Tidy up run artemis - Removed deployment, improved help, tidier arguments

### LINKED ISSUE #603: Tidy up run_artemis
If we keep adding more arguments to artemis (e.g. https://github.com/DiamondLightSource/python-artemis/pull/602/files) `run_artemis` is going to get more and more messy. We should modify `run_artemis` to:
* Pass any args through to artemis
* When help is called, return the help from that script plus the help from artemis

Additionally, the script is no longer responsible for deployment, we should remove logic around that

olliesilvester:
Fixes #603

### To test:
1. Confirm running ./run_artemis.sh --help combines help from the start-up script and artemis
2. Check other arguments still work as usual
3. Check other cleanup changes are suitable


dperl-dls (changes requested):
It seems like dev mode is not added to the options in `run_artemis.sh`, when I try `./run_artemis.sh --dev` I get `Unknown option --dev. Use --help for info on option usage.` Also, deploy should be removed from the help menu if the command is removed.

dperl-dls (changes requested):
the `--verbose-event-logging` option also doesn't work - it's not in the start script and therefore not passed on to artemis

olliesilvester:
> the `--verbose-event-logging` option also doesn't work - it's not in the start script and therefore not passed on to artemis

Thanks, --logging-level was also missing. Hopefully these are fine now

run_artemis.sh
```diff
@@ -43,6 +54,13 @@ for option in "$@"; do
     esac
 done
 
+#Check valid logging level was chosen
+if [[ "$LOGGING_LEVEL" != "INFO" && "$LOGGING_LEVEL" != "CRITICAL" && "$LOGGING_LEVEL" != "ERROR" 
+    && "$LOGGING_LEVEL" != "WARNING" && "$LOGGING_LEVEL" != "DEBUG" ]]; then
+    echo "Invalid logging level selected, defaulting to INFO"
```
dperl-dls, line 60: nice

PR #615, opened 2023-04-18
482 Added log handler to rollover tmp/dev/artemis.txt daily or when it reaches 100mb

### LINKED ISSUE #482: Rollover log files
As a backup Artemis logs to `/dls_sw/i03/logs/bluesky`as well as graylog, currently this is in one big file, we should rollover the file daily and also on size?

olliesilvester:
Fixes #482

### To test:
1. Run tests, in particular in test_log.py 
2. Confirm tests don't create permanent txt files


DominicOram:
After https://github.com/DiamondLightSource/python-artemis/pull/626 this should be done in `dodal`. I have moved it into https://github.com/DiamondLightSource/dodal/pull/60 but still needs review

PR #623, opened 2023-04-20
348 replaced point 3d and point 2d to support arithmetic, using numpy arrays

### LINKED ISSUE #348: Replace Point3D with something which supports arithmetic

olliesilvester:
Fixes #348

Link to dodal PR: #[37](https://github.com/DiamondLightSource/dodal/pull/37)

### To test:
1. Check np.array() calls are suitable
2. Confirm all unit tests pass


src/artemis/external_interaction/callbacks/fgs/tests/test_zocalo_handler.py
```diff
@@ -86,7 +85,7 @@ def test_zocalo_called_to_wait_on_results_when_communicator_wait_for_results_cal
     callbacks = FGSCallbackCollection.from_params(dummy_params)
     mock_zocalo_functions(callbacks)
     callbacks.ispyb_handler.ispyb_ids = (0, 0, 100)
-    expected_centre_grid_coords = Point3D(1, 2, 3)
+    expected_centre_grid_coords = create_point(1, 2, 3)
```
dperl-dls, line 88: so this would just be `expected_centre_grid_coords = ndarray([1,2,3])`

src/artemis/external_interaction/callbacks/fgs/tests/test_zocalo_handler.py
```diff
@@ -171,21 +170,21 @@ def test_multiple_results_from_zocalo_sorted_by_total_count_returns_centre_and_b
...
+                expected_centre_grid_coords[0] - 0.5,
+                expected_centre_grid_coords[1] - 0.5,
+                expected_centre_grid_coords[2] - 0.5,
             )
         )
     )
-    assert found_centre == expected_centre_motor_coords
+    np.testing.assert_array_equal(found_centre, expected_centre_motor_coords)
 
     expected_bbox_size = list(map(operator.sub, [8, 8, 7], [2, 2, 2]))
```
dperl-dls, line 189: while we're numpyifying stuff, this could be `... = ndarray([8, 8, 7]) - ndarray([2, 2, 2])`, and in the real code where this happens

src/artemis/external_interaction/ispyb/ispyb_dataclass.py
```diff
@@ -13,8 +14,8 @@
     "microns_per_pixel_x": 0.0,
     "microns_per_pixel_y": 0.0,
     # gets stored as 2x2D coords - (x, y) and (x, z). Values in pixels
-    "upper_left": Point3D(x=0, y=0, z=0),
-    "position": Point3D(x=0, y=0, z=0),
+    "upper_left": create_point(0, 0, 0),
```
dperl-dls, line 17: this would just be something like `"upper_left": [0,0,0]`

dperl-dls (changes requested):
This is a good start, but I think we can simplify it more, and probably get rid of all of utils.py in the end.

If we need to check the length at some point, we could do something like
```python
assert 2 <= len(point) <= 3
```
but for the parameters, they are validated by the json schema already so we are always sure they are length 3

PR #626, opened 2023-04-25
Move more logging into dodal

### LINKED ISSUE #569: Logs from dodal are not coming through
Since moving to dodal we have lost the logs that dodal makes in graylog.

DominicOram:
Fixes #569

Link to dodal PR (if required): (https://github.com/DiamondLightSource/dodal/pull/38)

### To test:
1. Confirm unit tests pass and are sensible

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     xarray
     doct
     databroker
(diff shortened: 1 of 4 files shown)


noemifrisina (commented):
It looks like the relative path from "ARTEMIS_LOG_DIR" in `test_set_env_variable_sets_correct_file_handler` is not set and the test fails with an AssertionError

noemifrisina (approved):
Looks good, all tests now pass

Issue #43, opened 2022-03-14
Remove redundant ISpyB params


DominicOram:
ISpyB has 3 parameters:
* `focal_spot_size_at_samplex`
* `slitgap_horizontal`
* `beamsize_at_samplex`

Where the 3rd parameter is always just calculated as the smallest value from the other 2. This implies that the first 2 are redundant for data analysis. We should remove them and confirm it still works.

## Acceptance Criteria
* Redundant parameters are not added to ISpyB etc.
* This is tested on a beamline

DominicOram:
`ispyb_params` soon to be wholesale removed so this is redundant

PR #631, opened 2023-04-28
262 Move arming out of stage

### LINKED ISSUE #262: Move arming the eiger out of stage
As documented on the bluesky wiki `stage` is not expected to take a meaningful amount of time. Bluesky therefore waits on any stage being complete before continuing. Arming the detector takes ~6s and we currently do it in the `stage`, we should move it out of the `stage` so that we can do other things concurrently.

## Acceptance Criteria
* Staging the eiger takes little time
* Starting to arm the detector can be done in parrallel with other tasks
* The staging/arming are done as soon as possible in the plan

olliesilvester:
(Somewhat) fixes #262

Link to dodal PR (if required): [#43](https://github.com/DiamondLightSource/dodal/pull/43)

### To test:
For now, just check code changes are sensible

### DIFF
--- a/src/artemis/experiment_plans/fast_grid_scan_plan.py
+++ b/src/artemis/experiment_plans/fast_grid_scan_plan.py
@@ -224,9 +224,12 @@ def run_gridscan(
     yield from set_fast_grid_scan_params(fgs_motors, parameters.experiment_params)
     yield from wait_for_fgs_valid(fgs_motors)
 
+    # Start fake_stage here (async arming)
+    yield from bps.abs_set(fgs_composite.eiger, 1, group="arming")
+
     @bpp.set_run_key_decorator("do_fgs")
     @bpp.run_decorator(md={"subplan_name": "do_fgs"})
-    @bpp.stage_decorator([fgs_composite.eiger])
+    # @bpp.stage_decorator([fgs_composite.eiger])
     def do_fgs():
         yield from bps.wait()  # Wait for all moves to complete
         yield from bps.kickoff(fgs_motors)
@@ -237,6 +240,11 @@ def run_gridscan(
 
     with TRACER.start_span("move_to_z_0"):
         yield from bps.abs_set(fgs_motors.z_steps, 0, wait=False)
+
+    # Wait for arming to finish
+    artemis.log.LOGGER.debug("Waiting for arming...")
+    yield from bps.wait("arming")
+    artemis.log.LOGGER.debug("Arming finished")
 
 
 @bpp.set_run_key_decorator("run_gridscan_and_move")


src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -224,9 +224,12 @@ def run_gridscan(
     yield from set_fast_grid_scan_params(fgs_motors, parameters.experiment_params)
     yield from wait_for_fgs_valid(fgs_motors)
 
+    # Start fake_stage here (async arming)
+    yield from bps.abs_set(fgs_composite.eiger, 1, group="arming")
```
DominicOram, line 228: Must: This will call the `set`method n the `eiger` object, you want to call the `set` on the specific arming signal

src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -238,6 +241,11 @@ def do_fgs():
     with TRACER.start_span("move_to_z_0"):
         yield from bps.abs_set(fgs_motors.z_steps, 0, wait=False)
 
+    # Wait for arming to finish
+    artemis.log.LOGGER.debug("Waiting for arming...")
+    yield from bps.wait("arming")
+    artemis.log.LOGGER.debug("Arming finished")
```
DominicOram, line 247: Nit: Great use of logging but might as well call these `INFO` rather than `DEBUG`

DominicOram (changes requested):
Great, some comments in code, otherwise good work.

src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -238,6 +241,11 @@ def do_fgs():
     with TRACER.start_span("move_to_z_0"):
         yield from bps.abs_set(fgs_motors.z_steps, 0, wait=False)
 
+    # Wait for arming to finish
+    artemis.log.LOGGER.info("Waiting for arming...")
+    yield from bps.wait("arming")  # Add timeout here?
+    artemis.log.LOGGER.info("Arming finished")
```
DominicOram, line 247: Must: This needs to be done before `do_fgs`

src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -224,9 +224,12 @@ def run_gridscan(
     yield from set_fast_grid_scan_params(fgs_motors, parameters.experiment_params)
     yield from wait_for_fgs_valid(fgs_motors)
 
+    # Start stage with asynchronous arming here
+    yield from bps.abs_set(fgs_composite.eiger.do_arm, 1, group="arming")
```
DominicOram, line 228: Could: This could be done earlier now

src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -226,11 +229,18 @@ def run_gridscan(
...
     def do_fgs():
-        yield from bps.wait()  # Wait for all moves to complete
-        yield from bps.kickoff(fgs_motors)
-        yield from bps.complete(fgs_motors, wait=True)
+        try:
+            yield from bps.wait()  # Wait for all moves to complete
+            yield from bps.kickoff(fgs_motors)
+            yield from bps.complete(fgs_motors, wait=True)
+        finally:
+            yield from bps.unstage(fgs_composite.eiger)
```
DominicOram, line 238: As discussed offline with @callumforrester it's a bit odd we're manually arming but disarming with `unstage`. We should address this in https://github.com/DiamondLightSource/dodal/issues/67

callumforrester: Another thought might be to rename the arming group to something like `"deferred_stage"`, which makes the symmetry more obvious

DominicOram: I like `arm`, it makes more sense to the scientists.

Issue #636, opened 2023-05-04
Bit depth being written wrong due to dectris firmware changes


dperl-dls:
Data files are being written with 32 bit depth even when data collected at 16 bit depth, this is causing zocalo issues with wrong max pixel values. The "correct" bit depth is now included in the meta file in `_dectris/bit_depth_image`

currently patched live on beamline in `write_nexus.py` line 176 where we do

```python
    "bit_depth_readout": "_dectris/bit_depth_image",
```
instead of previously
```python
    "bit_depth_readout": "_dectris/bit_depth_readout",
```

but this is confusing, and we should fix it properly (so that fields have the same name...), also see https://github.com/cctbx/dxtbx/pull/632

DominicOram:
Ultimately blocked by https://github.com/nexusformat/definitions/issues/1258

DominicOram:
This is now internal in nexgen, we still need a nicer longterm fix

PR #637, opened 2023-05-04
Update write_nexus.py

d-perl:
see https://github.com/DiamondLightSource/python-artemis/issues/636 and https://github.com/cctbx/dxtbx/pull/632


DominicOram (approved):
Annoying but neccessary

PR #641, opened 2023-05-04
Run the eiger without setting number of triggers

### LINKED ISSUE #560: Test arming detector before grid detection
If we can arm before grid detection we can produce some speed up. Currently we need to do grid detection for working out the number of images but after discussion with @GDYendell it should be possible to set the number of images to 0 and arm before/whilst working out grids.

## Acceptance Criteria
* Run with 0 frames and see if we still get good data
* We may need to be careful in the disarming to watch to make sure the number of written frames matches the expected number before disarming.

DominicOram:
Fixes #560

Link to dodal PR (if required): #https://github.com/DiamondLightSource/dodal/pull/48

Requires the following to be merged first:
* https://github.com/DiamondLightSource/python-artemis/pull/640
* https://github.com/DiamondLightSource/python-artemis/pull/639

### To test:
1. Confirm tests pass

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     xarray
     doct
     databroker
-    dodal @ git+https://github.com/DiamondLightSource/python-dodal.git@0fc304db51bbdfefabc5d6aaa18652f61adc11f4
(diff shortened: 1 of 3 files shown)


dperl-dls (approved):
Looks good, tests seem fine. Great to have this capability now

dperl-dls:
Tests are only failing because I removed the dodal commit pin which was a merge conflict - they will pass when the linked dodal PR is merged

PR #648, opened 2023-05-11
WIP: Replace RunEngine management components with blueapi 

### LINKED ISSUE #31: Investigate bluesky service

callumforrester:
Fixes #31

Changes:
* Add blueapi as a dependency
* Remove `BlueskyRunner` class and use `RunEngineWorker` from blueapi
* Replace experiment registry functionality with `BlueskyContext` from blueapi
* Create top-level plan for fast grid scans that handles the parameter translation previously handled by the experiment registry
* Replace composite device global variable with a device stored inside `BlueskyContext`

Still to do:
* Ensure equivalent tests in blueapi as for deleted runner class
* Any extra tests in artemis
* Error handling/status updates

### To test:
1. A/B testing: confirm scan still runs as expected


.vscode/launch.json
```diff
@@ -12,10 +12,10 @@
             "args": [
                 "--dev",
                 "--verbose-event-logging",
-                "--skip_startup_connection"
             ],
             "env": {
-                "EPICS_CA_SERVER_PORT": "5066"
+                "EPICS_CA_SERVER_PORT": "5364",
```
callumforrester, line 17: Need to revert before merging

src/artemis/experiment_plans/experiment_registry.py
```diff
@@ -22,24 +22,24 @@ def do_nothing():
...
-        "setup": do_nothing,
-        "run": not_implemented,
-        "internal_param_type": RotationInternalParameters,
-        "experiment_param_type": RotationScanParams,
-    },
-}
-EXPERIMENT_NAMES = list(PLAN_REGISTRY.keys())
-EXPERIMENT_TYPE_LIST = [p["experiment_param_type"] for p in PLAN_REGISTRY.values()]
-EXPERIMENT_TYPE_DICT = dict(zip(EXPERIMENT_NAMES, EXPERIMENT_TYPE_LIST))
+# EXPERIMENT_TYPES = Union[GridScanParams, RotationScanParams]
```
callumforrester, line 25: Possibly everything in this module should go in `__main__`, since it's just the error type left

DominicOram: `main` now has multiple plans listed here. We will need to run all of them

callumforrester: Added `full_grid_scan` too

callumforrester:
From offline discussion with @DominicOram a couple of weeks ago:
- Ideally want stop capability (from `StopOrStatus`)
- Definitely want status: https://github.com/DiamondLightSource/blueapi/pull/218
- Need to ensure `Worker` and `BlueskyContext` are tested to the same extent as `BlueskyRunner`: https://github.com/DiamondLightSource/blueapi/issues/198
- Some mechanism for `WarningException`s, I was initially confused by this. After discussing with @dperl-dls and @DiamondJoseph, we felt it was better that blueapi/artemis has no knowledge of what constitutes a warning and what constitutes a fatal error to the greater UDC operation while said operation is under the control of GDA. Our idea was to split `WarningException` into several more specific exceptions, and have GDA simply know which types of error mean it can carry on and which don't. @DominicOram what do you think?

callumforrester:
Also discussed: equivalent of `skip_startup_connection` in bluesky context

DominicOram:
Yh, I'm happy for GDA to just examine the Exception

DominicOram:
Spun out https://github.com/DiamondLightSource/python-artemis/issues/679 for warning refactor

DominicOram:
This has become a bit unwieldy and out of date. I'm going to take what we have here (thanks @callumforrester) and split it into 3 or 4 different issues/PRs

PR #661, opened 2023-05-15
Tidy up device creation for full grid scan

### LINKED ISSUE #660: Make sure we initialise aperture scatterguard properly
We should make sure we don't end up with an invalid aperture scatterguard object, additionally we no longer need to wait for devices if they've come from dodal.

DominicOram:
Fixes #660

### To test:
1. Confirm tests pass

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     xarray
     doct
     databroker
-    dodal @ git+https://github.com/DiamondLightSource/python-dodal.git@35a2a5891f4d2a09f2c9bf3e2098c5c9772d0633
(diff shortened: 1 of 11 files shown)


dperl-dls (approved):
great, thanks for tidying this up! I just added a little sneaky change (making the get beamline params a bit more general, in prep for adding beamlines) and a couple more tests, so I'll approve this and leave for you to look at

DominicOram:
I like moving to the dictionary for `BEAMLINE_PARAMETER_PATHS` but I would rather limit the places we pull out the environment variable. Can we reuse either `get_beamline_name` in `dodal` or `get_beamline_name` here? This will also mean we can better handle the simulation case.

dperl-dls:
Done, now this depends on https://github.com/DiamondLightSource/dodal/pull/66
This will require setting and S03 env var as currently implemented because I didn't want to put S03 as a default in a generic function

PR #666, opened 2023-05-19
Use xray centring OAV params when doing an xray centre

### LINKED ISSUE #653: Use xray centring params for grid detection
Currently the code in https://github.com/DiamondLightSource/python-artemis/pull/584 uses `loopCentring` parameters rather than `xrayCentring` this is because it relies on `max_tip_distance` being in the file, which it uses to disregard oddly shaped pins. The code in GDA does not disregard oddly shaped pins for xray centring and so doesn't need this parameter.

## Acceptance Criteria
* Artemis uses the `xrayCentring` parameters

DominicOram:
Fixes #653

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/64

### To test:
1. Confirm existing unit tests pass

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     xarray
     doct
     databroker
-    dodal @ git+https://github.com/DiamondLightSource/python-dodal.git@9a382e750bac181b9bc47343171b7e46190e1d03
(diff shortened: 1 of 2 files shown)


noemifrisina (approved):
Tests all pass

PR #670, opened 2023-05-19
Do grid detection in Artemis

### LINKED ISSUE #373: OAV: get gridscan grid from waveforms
Artemis should be able to get the grid to do the 3D xray centre based on the OAV image.

## Acceptance Criteria
* The grid size etc. is determined from the OAV image (for 0 and 90)
* Snapshots with grids are taken through Artemis
* This data is passed on correctly to existing code, which performs the grids
* GDA is able to trigger this, as well as just the xray centre with provided parameters
* There are sufficient system/unit tests around this

### LINKED ISSUE #652: Run Grid Detection and Grid in Artemis
Following https://github.com/DiamondLightSource/python-artemis/issues/554 we have a plan that we can trigger from GDA in exactly the same way we trigger the `fast_grid_scan` and it will calculate the grids in Artemis as well as in GDA then use the GDA values. We should move this to a full scan happening in Artemis e.g:
* The plan must only accept the values it needs
* It will do the grid detection, the full grid and all the work in between (https://github.com/DiamondLightSource/python-artemis/issues/567)
* It should be easy to turn this on/off in GDA

### LINKED ISSUE #658: full grid scan testing notes
- aperture z check for safe move has no tolerance (??!!)
- running once works, twice does not
- pin tip centring dosn't happen at start of full grid scan (do we want it to?)
- sometimes num_images is negative (??!!) https://graylog2.diamond.ac.uk/messages/logs_8246/27ff3cc1-f308-11ed-9360-1866dafadc6c -> sometimes num_steps is negative - was default values for min and max grid size the wrong way round - we are hitting edge of image somtimes
- centre often wrong and snapshots often not displayed

DominicOram:
Fixes #652, fixes #670, fixes #373, fixes #658

This has been tested on the beamline and works sufficiently, it could use some tidying though. See #671, #672, #673, #674. Please review bearing these in mind

### To test:
1. Tests pass

### DIFF
--- a/src/artemis/experiment_plans/full_grid_scan.py
+++ b/src/artemis/experiment_plans/full_grid_scan.py
@@ -17,6 +17,9 @@ from artemis.experiment_plans.oav_grid_detection_plan import (
     create_devices as oav_create_devices,
 )
 from artemis.experiment_plans.oav_grid_detection_plan import grid_detection_plan
+from artemis.external_interaction.callbacks.fgs.fgs_callback_collection import (
+    FGSCallbackCollection,
+)
 from artemis.log import LOGGER
 from artemis.parameters.internal_parameters.plan_specific.fgs_internal_params import (
     GridScanParams,
@@ -24,9 +27,6 @@ from artemis.parameters.internal_parameters.plan_specific.fgs_internal_params im
 from artemis.utils.utils import Point3D
 
 if TYPE_CHECKING:
-    from artemis.external_interaction.callbacks.fgs.fgs_callback_collection import (
-        FGSCallbackCollection,
-    )
     from artemis.parameters.internal_parameters.plan_specific.grid_scan_with_edge_detect_params import (
         GridScanWithEdgeDetectInternalParameters,
         GridScanWithEdgeDetectParams,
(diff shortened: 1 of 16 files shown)


src/artemis/experiment_plans/full_grid_scan.py
```diff
@@ -38,7 +42,7 @@ def create_devices():
     i03.aperture_scatterguard().wait_for_connection()
 
 
-def wait_for_det_to_finish_moving(detector: DetectorMotion, timeout=2):
+def wait_for_det_to_finish_moving(detector: DetectorMotion, timeout=120):
```
dperl-dls, line 45: does it really need to be this long?

dperl-dls (approved):
Looks good, tests pass, thanks!

src/artemis/experiment_plans/full_grid_scan.py
DominicOram: Yep, the detector motion is slow, @neil-i03 said that 2mins is about right

DominicOram: No, it's just how long until it will timeout. If the motion finishes earlier it will move on

PR #675, opened 2023-05-22
delete logging handlers in global test teardown

### LINKED ISSUE #644: logging handler test failing

dperl-dls:
and uncomment tests

Fixes #644

### To test:
1. see tests work I guess?


DominicOram (approved):
Great, thank you! We may end up having to do something similar in dodal but we'll cross that bridge when we come to it

PR #680, opened 2023-05-23
Add test for writing nexus files early

### LINKED ISSUE #677: Write a unit test for h5 files not existing yet
We had an issue on i03 where we were writing >1000 frames and nexgen was failing as it was trying to open files that did not exist. https://github.com/dials/nexgen/pull/119 fixes the issue but we should confirm with a test in artemis that this does not regress

DominicOram:
Fixes #677

### To test:
1. Confirm tests fail after `pip install nexgen==0.6.26`
2. Confirm tests pass after `pip install nexgen==0.6.27`
3. Confirm commented issue will solve skipped test

### DIFF
--- a/src/artemis/external_interaction/unit_tests/test_write_nexus.py
+++ b/src/artemis/external_interaction/unit_tests/test_write_nexus.py
@@ -1,4 +1,5 @@
 import os
+from contextlib import contextmanager
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


dperl-dls (approved):
tests pass and fail as described, the skipped test previously worked for correct nexus files so that issue should fix it. thanks!

Issue #681, opened 2023-05-23
Convert InternalParameters to pydantic models


dperl-dls:
The existing InternalParameters can prbably be reworked into a standalone "translator" for external to internal formats, then the actual internal param classes can just be simple models

PR #682, opened 2023-05-23
678 pydantic internal params

dperl-dls:
Fixes 678
Makes all the internal param classes into pydantic models. Does not yet do it for InternalParameters because that is more complicated - new issue for this: https://github.com/DiamondLightSource/python-artemis/issues/681

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/68

### To test:
1. Run tests

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     xarray
     doct
     databroker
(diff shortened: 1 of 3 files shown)


noemifrisina (approved):
Tests pass (usual TimeoutErrors notwithstanding), looks okay

PR #691, opened 2023-05-31
681 make InternalParameters pydantic models

### LINKED ISSUE #681: Convert InternalParameters to pydantic models

dperl-dls:
Fixes #681
Replaces InternalParameters with a simple pydantic base model and separates out the preprocessing into loose functions or the validators of the InternalParameters classes

Requires GDA changes (update of parameter version to `"0.0.5"` and change version datatype from float to string/semantic version)

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/77

### To test:
1. Run tests


noemifrisina (changes requested):
Merge conflict needs to be solved (`dodal` requirement on `main` now pointing to a commit)

PR #692, opened 2023-05-31
change i03 import paths

dperl-dls:
Needed for https://github.com/DiamondLightSource/dodal/issues/69

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     xarray
     doct
     databroker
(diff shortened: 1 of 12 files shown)


noemifrisina (changes requested):
Looks like a couple of tests are failing because `active_device_is_same_type` is not in `i03.py` anymore but in `beamline_utils`, aside from that seems okay

noemifrisina (approved):

PR #694, opened 2023-06-01
Wait on pin tip centre to valid before grid detection

### LINKED ISSUE #683: Grid position wildly off
See https://ispyb.diamond.ac.uk/dc/visit/mx33658-4/dcg/9491841

DominicOram:
Fixes #683 

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/86

### To test:
1. Confirm dodal tests pass


DominicOram:
@callumforrester would be interested to hear if you think there's a cleaner way of doing this?

dperl-dls (approved):
I think there probably exists a nicer way of doing this but I think this is good enough to merge for now, it's not that bad and it fixes the problem

PR #696, opened 2023-06-01
695 nexus writer get documents from plan

### LINKED ISSUE #695: make nexus callback get parameters from documents
needed for #370 and #597

dperl-dls:
Fixes #695 

somewhat reorganises the fgs nexus writer callback, lets it get parameters through documents, and changes where in the plan subscription is done

associated dodal PR: https://github.com/DiamondLightSource/dodal/pull/89

### To test:
1. Run tests


src/artemis/external_interaction/callbacks/fgs/nexus_callback.py
```diff
@@ -30,23 +27,46 @@ class FGSNexusFileHandlerCallback(CallbackBase):
...
-        self.run_gridscan_uid: Optional[str] = None
+    def __init__(self) -> None:
+        self.parameters: FGSInternalParameters | None = None
+        self.run_gridscan_uid: str | None = None
+        self.nexus_writer_1: FGSNexusWriter | None = None
+        self.nexus_writer_2: FGSNexusWriter | None = None
 
     def start(self, doc: dict):
-        if doc.get("subplan_name") == "run_gridscan":
+        if doc.get("subplan_name") == "run_gridscan_move_and_tidy":
```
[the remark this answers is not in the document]
DominicOram: Could: Why don't we create the nexus writers here and so not need the `run_gridscan` at all. (Note if we do this it may make some of my other comments moot)

src/artemis/external_interaction/callbacks/fgs/zocalo_callback.py
```diff
@@ -47,7 +47,7 @@ def __init__(
         ] = parameters.experiment_params.grid_position_to_motor_position
         self.processing_start_time = 0.0
         self.processing_time = 0.0
-        self.run_gridscan_uid: Optional[str] = None
+        self.do_fgs: Optional[str] = None
```
DominicOram, line 50: Should: I think it's still worth having `uid` in this variable name

src/artemis/external_interaction/nexus/nexus_utils.py
```diff
@@ -0,0 +1,170 @@
...
+            grid_scan_params.x_axis.start,
+            grid_scan_params.x_axis.step_size,
+        ),
+        Axis("chi", "sam_x", "rotation", (0.006, -0.0264, 0.9996), 0.0),
+        Axis("phi", "chi", "rotation", (-1, -0.0025, -0.0056), 0.0),
+    ]
+    return Goniometer(gonio_axes, grid_scan)
+
+
+def create_rotation_goniometer_axes(
```
DominicOram, line 70: Should: Can we refactor the common parts out between this and the grid scan gonio function?

src/artemis/external_interaction/nexus/write_nexus.py
```diff
@@ -285,6 +121,117 @@ def update_nexus_file_timestamp(self):
...
                 nxsfile["entry"].create_dataset(
-                    "end_time", data=np.string_(self._get_current_time())
+                    "end_time", data=np.string_(get_current_time())
                 )
             shutil.move(temp_filename, filename)
+
+    def get_image_datafiles(self, max_images_per_file=1000):
+        return [
+            self.directory / f"{self.filename}_{h5_num + 1:06}.h5"
+            for h5_num in range(-(-self.full_num_of_images // max_images_per_file))
```
DominicOram, line 131: Should: I think `ceil` is more readable than the quirky maths here

src/artemis/external_interaction/nexus/write_nexus.py
```diff
@@ -285,6 +121,117 @@ def update_nexus_file_timestamp(self):
...
+            parameters.experiment_params,
+            grid_scan,
+        )
+        self.grid_scan = grid_scan
+
+    def _get_data_shape_for_vds(self) -> tuple[int | float, ...]:
+        ax = list(self.grid_scan.keys())[0]
+        num_frames_in_vds = len(self.grid_scan[ax])
+        nexus_detector_params: EigerDetector = self.detector.detector_params
+        return (num_frames_in_vds, *nexus_detector_params.image_size)
```
DominicOram, line 151: I think this function might fall out as being just in the parent class but this is probably an issue for the following PR

src/artemis/external_interaction/nexus/write_nexus.py
```diff
@@ -285,6 +121,117 @@ def update_nexus_file_timestamp(self):
...
+            for h5_num in range(-(-self.full_num_of_images // max_images_per_file))
+        ]
+
+
+class FGSNexusWriter(NexusWriter):
+    grid_scan: dict
+
+    def __init__(self, parameters: FGSInternalParameters, grid_scan: dict) -> None:
+        super().__init__(parameters)
+        self.goniometer = create_gridscan_goniometer_axes(
```
DominicOram, line 140: Should: I think it would be nicer to have an abstract function to get the gonio that each subclass needs to implement. It's more explicitly obvious it's needed then

src/artemis/parameters/schemas/experiment_schemas/grid_scan_params_schema.json
```diff
@@ -46,7 +46,10 @@
         },
         "omega_start": {
             "type": "number"
-        }
+        },
+        "x_axis": {},
+        "y_axis": {},
+        "z_axis": {}
```
DominicOram, line 52: Should: These are derived values from the other parameters and so I don't think they should be in here

DominicOram (changes requested):
Nice, great refactor, thanks! A few comments in code. Also, can you add a couple of nexus callback tests on the unhappy path e.g. that we get sensible errors if it does not get the docs in the right order

src/artemis/external_interaction/nexus/nexus_utils.py
dperl-dls: there's not really much to easily factor out imo, for any given experiment type it's different fields which will have to be filled in - the common stuff is already in `Axis` etc.

dperl-dls: on discussion,it is possible and it will be easier to do with a scanspec added, but I think it belongs more with the nexus writing stuff so I'll do that in the following PR #697

src/artemis/external_interaction/nexus/write_nexus.py
dperl-dls: it is implemented properly in the following PR, where it is just `(total frames, *image_size)`

src/artemis/parameters/schemas/experiment_schemas/grid_scan_params_schema.json
DominicOram: Yh but what happens if you do specify them as something that doesn't match the above params? It's confusing as an API

dperl-dls: ah yeah alright, that's a good point. turns out it's not actually hard to exclude them, I got the wrong idea from some older threads on the topic

src/artemis/external_interaction/callbacks/fgs/nexus_callback.py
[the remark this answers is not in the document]
dperl-dls: ok, I've simplified the structure somewhat while keeping the writing in the ispyb plan bit for now, with a comment linking to the issue to fix this when we get the detector bit depth

src/artemis/external_interaction/callbacks/fgs/nexus_callback.py
```diff
@@ -40,29 +43,31 @@ def start(self, doc: dict):
...
             self.nexus_writer_1, self.nexus_writer_2 = create_3d_gridscan_writers(
                 self.parameters
             )
-
-    def descriptor(self, doc):
-        if doc.get("name") == "ispyb_readings":
-            assert self.parameters is not None, "Failed to update parameters"
             assert (
                 self.nexus_writer_1 is not None and self.nexus_writer_2 is not None
             ), "Failed to create Nexus files, writers were not initialised."
```
DominicOram: Should: This assertion now seems over the top given we just created the writers

src/artemis/external_interaction/callbacks/fgs/nexus_callback.py
```diff
@@ -40,29 +43,31 @@ def start(self, doc: dict):
             )
             json_params = doc.get("hyperion_internal_parameters")
```
DominicOram, line 44: Could: I still think it would be cleaner to just do this on any doc that contains `hyperion_internal_parameters`, regardless of name.

src/artemis/external_interaction/callbacks/fgs/nexus_callback.py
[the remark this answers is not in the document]
DominicOram: I still think it would be cleaner to just do this on any doc that contains `hyperion_internal_parameters`, regardless of name.

src/artemis/external_interaction/nexus/write_nexus.py
DominicOram: I still think this

dperl-dls: will do this in the other branch

src/artemis/external_interaction/callbacks/fgs/nexus_callback.py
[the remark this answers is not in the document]
DominicOram: As discussed, I think something like this will just fall out of the wider refactor in https://github.com/DiamondLightSource/python-artemis/issues/726

DominicOram (approved):
Brilliant, thank you

PR #697, opened 2023-06-02
WIP: 370 implement rotation nexus

### LINKED ISSUE #370: Make nexus writer deal with different experiment types

dperl-dls:
Fixes #370 
Implements writing of nexus file for rotation scans
Requires GDA changes: `current_energy` parameter renamed to `current_energy_ev`

### To test:
1. Do thing x
2. Confirm thing y happens


DominicOram:
Could you rebase this once https://github.com/DiamondLightSource/python-artemis/pull/696 is merged so that we can see the woods through the trees. Issues from that PR that are fixed in this:
* Refactor the two `create_gonio` functions in `write_nexus` to be share a common function
* `_get_data_shape_for_vds` can probably be generic between each writer
* It would be nicer to have an abstract function to get the gonio that each subclass needs to implement. It's more explicitly obvious it's needed then

dperl-dls:
closing in favour of https://github.com/DiamondLightSource/python-artemis/pull/736

Issue #627, opened 2023-04-27
Move deploy location to /ixx/software/bluesky/artemis


DominicOram:
In it's currently location Artemis is not editable by beamline staff, it should be. We can ask sci comp to make a folder editable recursively but this probably shouldn't be `ixx/software` so adding another level will be useful.

DominicOram:
Should probably do https://github.com/DiamondLightSource/python-artemis/issues/572 at the same time

PR #704, opened 2023-06-06
572 deploy dodal as repo in prod fixed

### LINKED ISSUE #572: Deploy dodal as a repo in prod
Currently we deploy artemis as a repo but not dodal, this is annoying if we have to make live fixes to both.

## Acceptance Criteria
* Deploy structure is changed to include separate folders e.g. `artemis_v.../artemis` and `artemis_v..../dodal`
* Deploy script updated
* Restart Artemis in GDA updated

olliesilvester:
Fixes #572  and #627

### To test:
1. Run "deploy_artemis.py dev" and confirm issue requirements are satifsfied, and that the symlink is correct
2. Confirm beamline deployment location is correct


dperl-dls (changes requested):
If I try to run either
`python deploy/deploy_artemis.py dev` or `./deploy/deploy_artemis.sh dev` I get a:
`PermissionError: [Errno 13] Permission denied: './dls_dev_env.sh'`
In a freshly cloned repo from this branch. I think it's probably because it writes to /tmp/ and the system policy here is to not allow execution from /tmp/
I feel that the bash script should have a --help option, and either script should ideally let you specify the location

olliesilvester:
> If I try to run either `python deploy/deploy_artemis.py dev` or `./deploy/deploy_artemis.sh dev` I get a: `PermissionError: [Errno 13] Permission denied: './dls_dev_env.sh'` In a freshly cloned repo from this branch. I think it's probably because it writes to /tmp/ and the system policy here is to not allow execution from /tmp/

Are you able to run deploy_artemis.py dev from the main branch? That script also runs ./dls_dev_env.sh from /tmp/

dperl-dls:
Yeah, can't run that one either. so I guess it is technically a separate issue that this doesn't work on rhel8 but we might as well fix it

olliesilvester:
It works on my Red Hat 8 machine, but I can look into what's causing the permission error

dperl-dls:
oh, interesting. I wonder if I just need to get someone to change the security policy for my machine then.

dperl-dls (approved):
On discussion, we will ignor potential security policy interference in execution from /tmp/ since that doesn't affect many people. will revisit if it becomes a bigger problem

PR #708, opened 2023-06-06
Improve the use of the pin tip detect device

### LINKED ISSUE #699: Fix pin tip detect
The pin tip detect added in https://github.com/DiamondLightSource/python-artemis/pull/694 doesn't actually get the `triggered_tip` correctly, fix this.

DominicOram:
Fixes #699

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/88

### To test:
1. Confirm modified unit tests pass and code is cleaner

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     xarray
     doct
     databroker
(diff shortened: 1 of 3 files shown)


dperl-dls (approved):
Great, thanks for fixing this!

PR #711, opened 2023-06-07
368 generalise ispyb handling to rotation scans

### LINKED ISSUE #368: Make ispyb handler deal with different experiment types

dperl-dls:
Fixes #368

### To test:
- Run the new system test in `test_rotation_scan_plan.py`
- verify that results appear in https://ispyb-test.diamond.ac.uk/dc/visit/cm31105-4


src/artemis/external_interaction/ispyb/store_in_ispyb.py
```diff
@@ -233,77 +220,234 @@ def _store_data_collection_table(self, data_collection_group_id: int) -> int:
...
+        self.data_collection_id = data_collection_id
+        self._store_position_table(conn, data_collection_id)
+
+        return data_collection_id, data_collection_group_id
+
+    def begin_deposition(self):
+        with ispyb.open(self.ISPYB_CONFIG_PATH) as conn:
+            return self._store_scan_data(conn)
+
+    def end_deposition(self, success: str, reason: str):
```
DominicOram, line 273: Should: Is the only difference between this and the grid one how many dcids we have? Feels like we can move some more logic into the parent here

src/artemis/external_interaction/ispyb/store_in_ispyb.py
```diff
@@ -233,77 +220,234 @@ def _store_data_collection_table(self, data_collection_group_id: int) -> int:
...
+        return params
+
+    def _store_scan_data(self, conn: Connector):
+        data_collection_group_id = self._store_data_collection_group_table(conn)
+        self.data_collection_group_id = data_collection_group_id
+        data_collection_id = self._store_data_collection_table(
+            conn, data_collection_group_id
+        )
+        self.data_collection_id = data_collection_id
+        self._store_position_table(conn, data_collection_id)
```
DominicOram, line 265: Should: The logic of "store group, store dc, store position" is the same for all these. Can that go in the parent class?

src/artemis/external_interaction/ispyb/store_in_ispyb.py
```diff
@@ -3,191 +3,179 @@
...
-
-        # writing to ispyb
-        self.conn: Connector = None
-        self.mx_acquisition = None
-        self.core = None
-
-        self.datacollection_ids = None
-        self.datacollection_group_id = None
-        self.grid_ids = None
+    def __init__(self, ispyb_config, parameters=None) -> None:
```
DominicOram, line 47: Nit: The parent class doesn't need parameters to be passed in

DominicOram: Should: Every subclass needs an `experiment_type`, let's pass this in here

DominicOram (changes requested):
A few initial comments, can pick the review up again later this week

src/artemis/external_interaction/ispyb/store_in_ispyb.py
dperl-dls: I guess, but I think it would be more complicated given the 3D gridscan is quite different

src/artemis/external_interaction/unit_tests/test_store_in_ispyb.py
```diff
@@ -69,22 +85,49 @@ def test_regex_string(dummy_ispyb, visit_path: str, expected_match: str):
 
 
 @patch("ispyb.open", new_callable=mock_open)
-def test_store_grid_scan(ispyb_conn, dummy_ispyb, dummy_params):
+def test_store_rotation_scan(
```
DominicOram, line 88: Should: I think we're mocking out a bit too much here, which is why we're getting such poor coverage on `StoreRotationInIspyb`. Can we either in this test or in a new test mock just the upserts?

src/artemis/external_interaction/ispyb/store_in_ispyb.py
```diff
@@ -233,77 +244,198 @@ def _store_data_collection_table(self, data_collection_group_id: int) -> int:
...
+        self.omega_start = self.detector_params.omega_start
+
+        if self.ispyb_params.xtal_snapshots_omega_start:
+            self.xtal_snapshots = self.ispyb_params.xtal_snapshots_omega_start
+        else:
+            self.xtal_snapshots = []
+            LOGGER.warning("No xtal snapshot paths sent to ISPyB!")
+        super().__init__(ispyb_config, "SAD")
+
+    def _mutate_data_collection_params_for_experiment(
```
DominicOram, line 269: Should: I think a few additional simple tests just covering this function would be good

src/artemis/external_interaction/ispyb/store_in_ispyb.py
```diff
@@ -233,77 +244,198 @@ def _store_data_collection_table(self, data_collection_group_id: int) -> int:
...
+        return "Hyperion rotation scan"
+
+
+class StoreGridscanInIspyb(StoreInIspyb):
+    ispyb_params: GridscanIspybParams | None = None
+    data_collection_ids: tuple[int, ...] | None = None
+    upper_left: list[int] | None = None
+    y_steps: int | None = None
+    y_step_size: int | None = None
+    grid_ids: tuple[int, ...] | None = None
```
DominicOram, line 312: Should: As above, avoid class variables

src/artemis/parameters/plan_specific/rotation_scan_internal_params.py
```diff
@@ -66,7 +86,30 @@ def get_num_images(self):
...
+            "zocalo_environment",
+            "beamline",
+            "insertion_prefix",
+            "experiment_type",
+        ]
+        detector_field_keys = list(DetectorParams.__annotations__.keys())
+        # not an annotation but specified as field encoder in DetectorParams:
+        detector_field_keys.append("detector")
+        ispyb_field_keys = list(IspybParams.__annotations__.keys()) + list(
+            GridscanIspybParams.__annotations__.keys()
```
DominicOram, line 109: Should: This should be the `RotationIspybParams` right?

src/artemis/parameters/plan_specific/grid_scan_with_edge_detect_params.py
```diff
@@ -34,14 +41,34 @@ def get_num_images(self):
...
         arbitrary_types_allowed = True
         json_encoders = {
             **ArtemisParameters.Config.json_encoders,
         }
 
+    def __init__(self, **args):
+        super().__init__(**args)
+
+    @staticmethod
+    def _artemis_param_key_definitions() -> tuple[list[str], list[str], list[str]]:
```
DominicOram, line 56: Should: I feel like we've moved too much of this function to be in each different class, when only a small part of it is different for each. Could we only move the different part please?

src/artemis/parameters/plan_specific/rotation_scan_internal_params.py
DominicOram: Can you write a test that would have caught this too

DominicOram (changes requested):
Mostly good, thanks. Some suggestions for improvements in the code. Also, when running the new system test I get `src/artemis/experiment_plans/tests/test_rotation_scan_plan.py::test_ispyb_deposition_in_plan - AttributeError: armed`

src/artemis/external_interaction/callbacks/rotation/ispyb_callback.py
```diff
@@ -0,0 +1,93 @@
...
+from bluesky.callbacks import CallbackBase
+
+from artemis.external_interaction.exceptions import ISPyBDepositionNotMade
+from artemis.external_interaction.ispyb.store_in_ispyb import StoreRotationInIspyb
+from artemis.log import LOGGER, set_dcgid_tag
+from artemis.parameters.constants import ISPYB_PLAN_NAME, SIM_ISPYB_CONFIG
+from artemis.parameters.plan_specific.fgs_internal_params import FGSInternalParameters
+
+
+class RotationISPyBHandlerCallback(CallbackBase):
```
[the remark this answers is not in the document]
dperl-dls: sure

src/artemis/external_interaction/unit_tests/test_store_in_ispyb.py
dperl-dls: I don't really understand. As fast as I can tell we are only mocking the connection (which is necessary at least for tests which run in CI) and nothing else? low coverage seems more down to not testing `_mutate_data_collection_params_for_experiment`

DominicOram: I mean that later in this test we have lots of `when(x).y().thenReturn(z)`, which means we're never really testing any of the functionality of `x.y()`. I think it's fine to leave it for now though,given we have the additional tests on the mutate

DominicOram (approved):
Minor issue in logging but once I fixed that this worked. Thank you very much, sorry for the long delay on it!

Issue #718, opened 2023-06-12
GridScanWithEdgeDetectInternalParameters fails on init


DominicOram:
When trying to run a full grid scan from GDA we get the following error:

```
Artemis failed to start with message: 
TypeError("GridScanWithEdgeDetectInternalParameters.__init__() got an unexpected keyword argument 'artemis_params'")
```

## Acceptance Criteria
* GDA can run the full plan
* There is a unit test in Hyperion covering this

PR #721, opened 2023-06-12
Bump parameter version to 1.0.0

DominicOram:
### To test:
1. Confirm unit tests are still passing
2. Confirm I haven't missed anywhere we're still referring to "0.0.5"
3. Run https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/39272/1 against this branch and confirm you do not get a validation error on the parameter version (you will instead get https://github.com/DiamondLightSource/python-artemis/issues/718)

### DIFF
--- a/src/artemis/parameters/schemas/full_external_parameters_schema.json
+++ b/src/artemis/parameters/schemas/full_external_parameters_schema.json
@@ -3,7 +3,7 @@
     "type": "object",
     "properties": {
         "params_version": {
-            "const": "0.0.5"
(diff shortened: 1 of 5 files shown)


noemifrisina (changes requested):
One test failing because of an `AttributeError` in GridScanParams. The fix is already in `main`, so merging main should fix it

DominicOram:
> One test failing because of an `AttributeError` in GridScanParams. The fix is already in `main`, so merging main should fix it

Done

Issue #674, opened 2023-05-19
Add more tests around full_grid_scan


DominicOram:
Currently the tests we have are just smoke tests. Unit tests on the logic and ideally system tests on the plan would be good.

PR #722, opened 2023-06-12
Fix bug in API for grid scan with edge detection

DominicOram:
Fixes #718 

Note this is based off https://github.com/DiamondLightSource/python-artemis/pull/721, ensure that PR is reviewed/merged first

### To test:
1. Confirm new test passes and is sensible

PR #730, opened 2023-06-13
Tests for full grid scan

Tom-Willemsen:
Fixes https://github.com/DiamondLightSource/python-artemis/issues/674

Adds a unit test for `full_grid_scan`

Note: depends on https://github.com/DiamondLightSource/python-artemis/pull/722

### DIFF
--- a/src/artemis/experiment_plans/tests/test_full_grid_scan_plan.py
+++ b/src/artemis/experiment_plans/tests/test_full_grid_scan_plan.py
@@ -1,15 +1,59 @@
 from typing import Generator
-from unittest.mock import patch
+from unittest.mock import patch, MagicMock, ANY
+from artemis.external_interaction.callbacks.fgs.fgs_callback_collection import (
+    FGSCallbackCollection,
+)
 
 import pytest
+import numpy as np
+
 from dodal.beamlines.i03 import detector_motion
 from dodal.devices.aperturescatterguard import AperturePositions
+from dodal.devices.oav.oav_parameters import OAVParameters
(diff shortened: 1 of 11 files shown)


src/artemis/experiment_plans/tests/test_full_grid_scan_plan.py
```diff
@@ -48,3 +92,68 @@ def test_get_plan(test_params, mock_subscriptions, test_config_files):
...
+    mock_grid_detection_plan,
+    mock_fgs_callbacks,
+    eiger,
+    backlight,
+    detector_motion,
+    aperture_scatterguard,
+    RE,
+    test_full_grid_scan_params,
+    mock_subscriptions,
+    test_config_files,
```
DominicOram, line 114: Could: A few type hints in here would be nice

DominicOram (changes requested):
Generally good, minor tidying up I think you could do. Also some minor workflow issues:
* We usually to add the issue number in the branch name and commit messages e.g. see https://github.com/DiamondLightSource/python-artemis/pull/708. People are a bit hit and miss about it though, not a big deal.
* vscode should be sorting imports on save, do you have the following in your `settings.json`?
```
"editor.codeActionsOnSave": {
        "source.organizeImports": true
    },
    "python.sortImports.args": [
        "--profile",
        "black"
    ],
```

DominicOram (approved):
Looks good, wait until #722 is merged but otherwise fine

dperl-dls:
merging as #722  is merged now

Issue #91, opened 2022-05-03
MVP for user sign off on 3D grid scans


DominicOram:
This is an overarching ticket that will be split into separate tickets to implement details.

Other tickets in Artemis will get us to the point where a 3D grid scan is performed and then the optimum point is moved to. However there are the following things that need to be done for the system to be complete:

* Two OAV images need to be taken and passed to Synchweb for the results to be laid upon. Note that these two images should be taken together before the 3D gridscan is performed
* The grid used needs to take into account misaligned sample pins such that the rotation of the pin causes the sample to be outside of the grid.

PR #731, opened 2023-06-13
Use callbacks in grid detection plan

### LINKED ISSUE #671: Use callbacks in grid detection
Currently the full grid plan passes in empty lists to grid detection plan that then are filled in with data to pass to the FGS. Instead this should be handled with callbacks.

Some of the callbacks will be ones that cover both detection and FGS (like the ispyb params). Some will bbe picking up data to then be passed to the FGS plan.

Note: We must retain the ability to run FGS independently by providing it expected values

noemifrisina:
Fixes #671 

Link to dodal PR (if required): [#91](https://github.com/DiamondLightSource/dodal/pull/91)

### To test:
1.Run tests

### DIFF
--- a/src/artemis/experiment_plans/tests/test_full_grid_scan_plan.py
+++ b/src/artemis/experiment_plans/tests/test_full_grid_scan_plan.py
@@ -130,6 +130,9 @@ def test_detect_grid_and_do_gridscan(
         # Verify we called the grid detection plan
         mock_grid_detection_plan.assert_called_once()
 
+        # Verify callback to oav snaposhot was called
+        mock_oav_callback.assert_called_once()
+
         # Check backlight was moved OUT
         assert backlight.pos.get() == Backlight.OUT
 


src/artemis/experiment_plans/full_grid_scan.py
```diff
@@ -90,23 +93,29 @@ def detect_grid_and_do_gridscan():
             f"{detector_params.prefix}_{detector_params.run_number}_{{angle}}"
         )
 
-        out_snapshot_filenames = []
-        out_upper_left = {}
+        oav_callback = OavSnapshotCallback()
+        yield from bps.subscribe(oav_callback)
 
         yield from grid_detection_plan(
```
DominicOram, line 99: Should: I think you can do the subscribe/unsubscribe in a decorator

src/artemis/experiment_plans/oav_grid_detection_plan.py
```diff
@@ -161,13 +149,13 @@ def grid_detection_main_plan(
...
         yield from bps.trigger(oav.snapshot, wait=True)
 
-        out_snapshot_filenames.append(
-            [
-                path_join(snapshot_dir, f"{snapshot_filename}_grid_overlay.png"),
-                path_join(snapshot_dir, f"{snapshot_filename}_outer_overlay.png"),
-                path_join(snapshot_dir, f"{snapshot_filename}.png"),
-            ]
-        )
+        yield from bps.create("snapshot_directories")
```
DominicOram, line 152: Should: This isn't just directories any more

DominicOram (changes requested):
Minor comments in code, I think we should also merge https://github.com/DiamondLightSource/python-artemis/pull/730 first so that we can ensure it doesn't break tests

noemifrisina:
Done, thanks

DominicOram:
Could you:
* Fix conflicts
* Confirm test in https://github.com/DiamondLightSource/python-artemis/pull/730 still works on this
* Add new tests if you think that https://github.com/DiamondLightSource/python-artemis/pull/730 doesn't fully test the functionality

DominicOram (approved):
Looks good, thanks

PR #735, opened 2023-06-13
Fix failing system tests

noemifrisina:
Fixes a few broken system tests

### To test:
1. Check that all system tests run


dperl-dls (approved):
confirmed that all system tests pass with these changes. thanks!

PR #736, opened 2023-06-14
370 implement rotation nexus

### LINKED ISSUE #370: Make nexus writer deal with different experiment types

d-perl:
The history got too messy on the other branch so this is a new one
Fixes https://github.com/DiamondLightSource/python-artemis/issues/370
- Implements writing of nexus file for rotation scans
- simplifies the nexus_writer further as in comments from https://github.com/DiamondLightSource/python-artemis/pull/696

Requires Dodal PR: https://github.com/DiamondLightSource/dodal/tree/artemis_370_implement_rotation_nexus
Requires Nexgen PR https://github.com/dials/nexgen/pull/120
Requires GDA changes: [current_energy parameter renamed to current_energy_ev](https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/39308)

To test:
    Run tests


src/artemis/external_interaction/system_tests/test_write_rotation_nexus.py
```diff
@@ -0,0 +1,167 @@
...
+            "hyperion_internal_parameters": parameters.json(),
+        }
+    )
+    def plan():
+        yield from bps.sleep(0)
+
+    return plan()
+
+
+def test_rotation_scan_nexus_output_compared_to_existing_file(
```
DominicOram, line 63: Nice test, thanks!

src/artemis/external_interaction/unit_tests/test_write_nexus.py
```diff
@@ -1,20 +1,80 @@
...
+    create_rotation_goniometer_axes,
 )
-from artemis.parameters.external_parameters import from_file as default_raw_params
 from artemis.parameters.plan_specific.fgs_internal_params import FGSInternalParameters
+from artemis.parameters.plan_specific.rotation_scan_internal_params import (
+    RotationInternalParameters,
+)
+
+
+def test_generic_create_gonio_creates_same_gonio_as_defined_with_individual_params(
```
DominicOram, line 30: Should: This test is great to confirm what we've done in this issue but seems a bit pointless going forward given the old way will no longer exist, suggest we remove it.

src/artemis/parameters/plan_specific/rotation_scan_internal_params.py
```diff
@@ -26,8 +26,8 @@ class RotationScanParams(BaseModel, AbstractExperimentParameterBase):
     image_width: float = 0.1
     omega_start: float = 0.0
     phi_start: float = 0.0
-    chi_start: Optional[float] = None
-    kappa_start: Optional[float] = None
+    chi_start: float | None = None
+    kappa_start: float | None = None
```
DominicOram, line 30: Could: Probably a new ticket but sort of feels like all of these are optional (apart from maybe the one you are rotating)  and if not provided you just use the current value

src/artemis/parameters/plan_specific/rotation_scan_internal_params.py
```diff
@@ -37,7 +37,10 @@ class RotationScanParams(BaseModel, AbstractExperimentParameterBase):
 
     @validator("rotation_direction", pre=True)
     def _parse_direction(cls, rotation_direction: str | int):
-        return RotationDirection[rotation_direction]
+        if isinstance(rotation_direction, str):
+            return RotationDirection[rotation_direction]
+        else:
+            return RotationDirection(rotation_direction)
```
DominicOram, line 43: Nit: I would say we shouldn't allow the integer representation, only the string as it makes the API messier but not that fussed

DominicOram (changes requested):
Great, mostly the refactor we discussed and some minor comments. Also the `__init__` file under the external interactions unit tests is spelt wrong

src/artemis/parameters/plan_specific/rotation_scan_internal_params.py
dperl-dls: Agree in principle but that's an issue with the rotation scan plan, not nexus writing

src/artemis/experiment_plans/rotation_scan_plan.py
```diff
@@ -151,35 +158,31 @@ def rotation_scan_plan(
...
-        "backlight": backlight,
-    }
-    subscriptions = RotationCallbackCollection.from_params(params)
+def get_plan(parameters: RotationInternalParameters):
+    devices = create_devices()
+    subscriptions = RotationCallbackCollection.from_params(parameters)
 
-    @subs_decorator(list(subscriptions))
+    @bpp.subs_decorator(list(subscriptions))
+    @bpp.set_run_key_decorator("run_gridscan_move_and_tidy")
```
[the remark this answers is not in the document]
dperl-dls: yes, setting the run key is what makes a section generate a start doc

DominicOram (changes requested):
Some comments in code. As discussed, I think this is cleaner than before but not 100%, I've made a few tickets to clean up more such as https://github.com/DiamondLightSource/python-artemis/issues/747

src/artemis/parameters/plan_specific/fgs_internal_params.py
```diff
@@ -50,3 +52,87 @@ def _preprocess_artemis_params(
...
+                z_axis.end,
+                z_axis.full_steps,
+            ) * ~Line(
+                "sam_x",
+                x_axis.start,
+                x_axis.end,
+                x_axis.full_steps,
+            )
+            scan_path = ScanPath(spec.calculate())
+            return scan_path.consume().midpoints
```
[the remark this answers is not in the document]
dperl-dls: I did something similar, but I wasn't keen on calculating unused lines so I still kept those separate

src/artemis/parameters/internal_parameters.py
```diff
@@ -175,3 +175,14 @@ def _preprocess_artemis_params(
         cls, all_params: dict[str, Any], values: dict[str, Any]
     ):
         ...
+
+    @abstractmethod
+    def get_scan_points(cls):
```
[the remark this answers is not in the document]
dperl-dls: currently we're using it to calculate the data shape, so we would still need to do it in multiple places, but less consistently. I think we should change this in #745 instead

DominicOram (approved):
Great, thank you!

PR #737, opened 2023-06-14
Update launch.json

dperl-dls:
Commandline args were changed a while ago but not reflected in launch.json


DominicOram (approved):
Great, thanks!

PR #752, opened 2023-06-21
Add testing for fast grid scan without grid detect

### LINKED ISSUE #686: Fast grid scan without grid detect broken
We have a property on the beamline to turn grid detection in Artemis off and just use the provided values, this no longer works. We should fix it as part of the philosophy that all individual subplans should remain callable

olliesilvester:
Fixes #686 

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/92

The unit test seems to work but takes a long time to complete for some reason

### To test:
Confirm unit test is sensible

### DIFF
--- a/src/artemis/experiment_plans/tests/test_fast_grid_scan_plan.py
+++ b/src/artemis/experiment_plans/tests/test_fast_grid_scan_plan.py
@@ -400,5 +400,4 @@ def test_fgs_arms_eiger_without_grid_detect(
     )
 
     RE(bps.stage(fake_fgs_composite.eiger))
-    fake_fgs_composite.eiger.arming_status.wait(60)
     fake_fgs_composite.eiger._wait_fan_ready.assert_called_once()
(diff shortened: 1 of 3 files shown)


src/artemis/experiment_plans/tests/test_fast_grid_scan_plan.py
```diff
@@ -366,3 +368,31 @@ def test_when_exception_occurs_during_running_then_eiger_disarmed(
...
+    fake_fgs_composite.eiger.set_odin_pvs = MagicMock(return_value=get_good_status())
+    fake_fgs_composite.eiger.stale_params.sim_put(0)
+    fake_fgs_composite.eiger._wait_fan_ready = MagicMock(return_value=get_good_status())
+    fake_fgs_composite.eiger._wait_for_odin_status = MagicMock(
+        return_value=get_good_status()
+    )
+
+    RE(bps.stage(fake_fgs_composite.eiger))
+    fake_fgs_composite.eiger.arming_status.wait(60)
+    fake_fgs_composite.eiger._wait_fan_ready.assert_called_once()
```
DominicOram, line 398: Must: I think the test needs to assert that after the `stage` we're armed without needing to do a wait (to get it consistent we may need one of the mock sets to take some fraction of time)

DominicOram (changes requested):
Good test but I think we should change it slightly as in comment

DominicOram (approved):
Great, thank you

Issue #744, opened 2023-06-16
Get transmission for ispyb from the PV


DominicOram:
We currently pass the transmission for ispyb deposition in as a parameter from GDA, we should instead read the PV

DominicOram:
This should be the actual transmission, not the set point

PR #753, opened 2023-06-21
716 full grid scan sets transmission first

### LINKED ISSUE #716: Full grid scan should set transmission first
GDA currently does this but with https://github.com/DiamondLightSource/dodal/pull/85 done we can do it in Hyperion. Also we can do the reading from the PV and giving to ispyb too.

## Acceptance Criteria
* The transmission is set in the full grid scan
* The value of the transmission is read back and put in ispyb

olliesilvester:
Fixes #716 and #744 

This PR uses code in #742 , so is waiting on this

### To test:
1. Confirm both requirements are met
2. Run tests


olliesilvester:
When this is merged, the GDA code currently handling this should be removed

src/artemis/experiment_plans/full_grid_scan.py
```diff
@@ -76,7 +78,9 @@ def detect_grid_and_do_gridscan(
 ):
     # Start stage with asynchronous arming here
     yield from bps.abs_set(eiger.do_arm, 1, group="arming")
-
+    yield from bps.abs_set(
+        attenuator, parameters.artemis_params.ispyb_params.transmission
+    )
```
DominicOram, line 83: Must: We need to wait on this to be done in the gridscan. In fact it falls under the same sort of wait as `arming`, I think we rename arming (here and in the wait) to something like "ready_for_data_collection" then put the `set_transmission` in the same group

src/artemis/experiment_plans/tests/test_fast_grid_scan_plan.py
```diff
@@ -88,16 +88,17 @@ def test_read_hardware_for_ispyb_updates_from_ophyd_devices(
...
     RE.subscribe(test_ispyb_callback)
 
-    def standalone_read_hardware_for_ispyb(und, syn, slits):
+    def standalone_read_hardware_for_ispyb(und, syn, slits, attn):
         yield from bps.open_run()
-        yield from read_hardware_for_ispyb(und, syn, slits)
+        yield from read_hardware_for_ispyb(und, syn, slits, attn)
         yield from bps.close_run()
 
     RE(
```
DominicOram, line 96: Should: We should add transmission to the test values and assertions here

DominicOram (changes requested):
Looks good, thanks. I think we need to change the API to make it clearer this is now the transmission we want to set, not the readback. I suggest adding `transmission` to the experiment params for external API instead of ispyb params (but keep the ones in ispyb params for internal use)

olliesilvester:
Right now there tests are failing due to a new Pydantic version, the same as https://github.com/DiamondLightSource/python-artemis/pull/758#pullrequestreview-1510856133

DominicOram:
> Right now the tests are failing due to a new Pydantic version, the same as [#758 (review)](https://github.com/DiamondLightSource/python-artemis/pull/758#pullrequestreview-1510856133)

Should now be fixed with `main` merged in. I will re-review.

DominicOram (approved):
Great, thank you

PR #755, opened 2023-06-22
Add deadtime optimisation

### LINKED ISSUE #754: Add deadtime optimisation for attenuator
Attenuation optimisation using total counts is added in https://github.com/DiamondLightSource/python-artemis/pull/742 , but this algorithm is normally ran using the dead time optimisation, so we need to add this

olliesilvester:
Fixes #754 

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/102

### To test:
1. Confirm deadtime optimisation logic is the same of that in the GDA script mx-config/scripts/attenuation_optimiser.py, in the run_xspress3mini_optimisation_routine()

2. Confirm new tests run and are sensible

### DIFF
--- a/src/artemis/experiment_plans/optimise_attenuation_plan.py
+++ b/src/artemis/experiment_plans/optimise_attenuation_plan.py
@@ -55,11 +55,6 @@ def check_parameters(
             )
         )
 
-    if upper_limit < lower_limit:
-        raise ValueError(
-            f"Upper limit {upper_limit} must be greater than lower limit {lower_limit}"
-        )
-
     if default_high_roi < default_low_roi:
         raise ValueError(
             f"Upper roi {default_high_roi} must be greater than lower roi {default_low_roi}"
--- a/src/artemis/experiment_plans/optimise_attenuation_plan.py
+++ b/src/artemis/experiment_plans/optimise_attenuation_plan.py
@@ -117,7 +117,7 @@ def do_device_optimise_iteration(
     """Set transmission, set number of images on xspress3mini, arm xspress3mini and zebra"""
     yield from bps.abs_set(attenuator, transmission, group="set_transmission")
     yield from bps.abs_set(xspress3mini.set_num_images, 1, wait=True)
-    arm_devices(xspress3mini, zebra)
+    yield from arm_devices(xspress3mini, zebra)
 
 
 def deadtime_optimisation(
--- a/src/artemis/experiment_plans/optimise_attenuation_plan.py
+++ b/src/artemis/experiment_plans/optimise_attenuation_plan.py
@@ -64,21 +64,22 @@ def arm_devices(xspress3mini, zebra):
     yield from arm_zebra(zebra)
 
 
-# readout scaler values
-def read_scaler_values(xspress3mini: Xspress3Mini) -> dict:
-    # Then get timeseriescontrol
-
-    scaler_values = {}
-    scaler_values["time"] = xspress3mini.channel_1.time.get()
-    scaler_values["reset_ticks"] = xspress3mini.channel_1.reset_ticks.get()
-    scaler_values["reset_count"] = xspress3mini.channel_1.reset_count.get()
-    scaler_values["all_event"] = xspress3mini.channel_1.all_event.get()
-    scaler_values["all_good"] = xspress3mini.channel_1.all_good.get()
-    scaler_values["pileup"] = xspress3mini.channel_1.pileup.get()
-    scaler_values["total_time"] = xspress3mini.channel_1.total_time.get()
-
-    # TODO: If we only use total time and reset ticks, this function may not be needed. Check if we use the other readings at all
-    return scaler_values
+def is_deadtime_optimised(
+    direction, deadtime, deadtime_threshold, transmission
+) -> tuple[bool, bool]:
+    flip_direction = False
+    if direction:
+        if deadtime >= deadtime_threshold:
+            flip_direction = True
+            return False, flip_direction
+        else:
+            # TODO: is this number the 10% cap?
+            if transmission >= 0.9:
+                return True, flip_direction
+    else:
+        if deadtime <= deadtime_threshold:
+            return True, flip_direction
+    return False, flip_direction
 
 
 def deadtime_optimisation(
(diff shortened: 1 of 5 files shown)


src/artemis/experiment_plans/optimise_attenuation_plan.py
```diff
@@ -0,0 +1,290 @@
...
+    target, upper_limit, lower_limit, default_high_roi, default_low_roi
+):
+    if target < lower_limit or target > upper_limit:
+        raise (
+            ValueError(
+                f"Target {target} is outside of lower and upper bounds: {lower_limit} to {upper_limit}"
+            )
+        )
+
+    if upper_limit < lower_limit:
```
olliesilvester, line 58: I think this check is implied in the if statement before, so I will remove this

src/artemis/experiment_plans/optimise_attenuation_plan.py
```diff
@@ -58,6 +79,87 @@ def arm_devices(xspress3mini, zebra):
...
+) -> tuple[bool, bool]:
+    flip_direction = False
+    if direction:
+        if deadtime >= deadtime_threshold:
+            flip_direction = True
+            return False, flip_direction
+        else:
+            # The 0.9 is hardcoded in GDA
+            if transmission >= 0.9:
+                return True, flip_direction
```
DominicOram, line 93: Must: Are we sure this is right? It looks wrong to me because if the deadtime is less than threshold and the transmission less than 90% the function returns nothing

src/artemis/experiment_plans/optimise_attenuation_plan.py
```diff
@@ -58,6 +79,87 @@ def arm_devices(xspress3mini, zebra):
     yield from arm_zebra(zebra)
 
 
+def deadtime_is_transmission_optimised(
```
DominicOram, line 82: Could: It might be cleaner if the function returned whether you're optimised and then which direction to move in, rather than the flip

DominicOram: Could: It would be even more clear if we just separate it out into two functions, one to work out if we're there and one to work out if to flip. It might lead to some repeated work but clarity is more important than efficiency here

src/artemis/experiment_plans/optimise_attenuation_plan.py
```diff
@@ -58,6 +79,87 @@ def arm_devices(xspress3mini, zebra):
...
+        is_optimised, flip_direction = deadtime_is_transmission_optimised(
+            direction, deadtime, deadtime_threshold, transmission
+        )
+
+        if is_optimised:
+            optimised_transmission = transmission
+            break
+
+        if flip_direction:
+            direction = not direction
```
DominicOram, line 154: Should: We need some tests that cause the direction to be flipped

src/artemis/experiment_plans/tests/test_optimise_attenuation_plan.py
```diff
@@ -148,3 +218,22 @@ def test_arm_devices_runs_correct_functions(RE: RunEngine):
     RE(arm_devices(xspress3mini, zebra))
     xspress3mini.arm.assert_called_once()
     optimise_attenuation_plan.arm_zebra.assert_called_once()
+
+
+def test_deadtime_calc_new_transmission_gets_correct_value():
+    assert deadtime_calc_new_transmission(True, 0.05, 2) == 0.1
+    assert deadtime_calc_new_transmission(False, 0.05, 2) == 0.025
+    assert deadtime_calc_new_transmission(True, 1, 2) == 1
```
DominicOram, line 226: Nit: Better done using parameterize, will be clearer when one fails

src/artemis/experiment_plans/optimise_attenuation_plan.py
DominicOram: Should: Can we have some separate tests on this function

DominicOram (changes requested):
Good work, thank you. Some comments in code

src/artemis/experiment_plans/optimise_attenuation_plan.py
```diff
@@ -58,6 +79,87 @@ def arm_devices(xspress3mini, zebra):
...
+):
+    """Set transmission, set number of images on xspress3mini, arm xspress3mini and zebra"""
+    yield from bps.abs_set(attenuator, transmission, group="set_transmission")
+    yield from bps.abs_set(xspress3mini.set_num_images, 1, wait=True)
+    yield from arm_devices(xspress3mini, zebra)
+
+
+def deadtime_optimisation(
+    attenuator, xspress3mini, zebra, transmission, increment, deadtime_threshold
+):
```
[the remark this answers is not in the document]
olliesilvester: I'll add a docstring to the total counts version as well

src/artemis/experiment_plans/optimise_attenuation_plan.py
```diff
@@ -58,6 +79,87 @@ def arm_devices(xspress3mini, zebra):
...
+        if deadtime <= deadtime_threshold:
+            return True, flip_direction
+    return False, flip_direction
+
+
+def deadtime_calc_new_transmission(direction, transmission, increment):
+    if direction:
+        transmission *= increment
+        if transmission > 0.999:
+            transmission = 1
```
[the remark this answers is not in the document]
olliesilvester: Yes, Neil said there is currently a 10% cap on transmission although I couldn't find this on the GDA code. I'll add it in

src/artemis/experiment_plans/tests/test_optimise_attenuation_plan.py
```diff
@@ -48,6 +57,7 @@ def get_good_status():
     return status
 
 
+@pytest.mark.skip(reason="Flakey test which is refactored in another PR")
```
olliesilvester, line 60: Skipped this test for now as it's refactored in https://github.com/DiamondLightSource/python-artemis/pull/763 (although still isn't working properly after that)

olliesilvester: Coverage is low due to this test being skipped, but #763 should fix that

src/artemis/experiment_plans/optimise_attenuation_plan.py
```diff
@@ -58,35 +86,230 @@ def arm_devices(xspress3mini, zebra):
...
+
+        total_time = xspress3mini.channel_1.total_time.get()
+        reset_ticks = xspress3mini.channel_1.reset_ticks.get()
+
+        LOGGER.info(f"Current total time = {total_time}")
+        LOGGER.info(f"Current reset ticks = {reset_ticks}")
+        deadtime = 0
+
+        """ Deadtime is the time after each event during which the detector cannot record another event.
+            The reset ticks PV stops ticking while the detector is unable to process events, so the absolute difference between the total time and the
```
olliesilvester, line 220: I'm not 100% sure about this description of the reset ticks, but given the formula and definition for deadtime it looks like it must be something like this

DominicOram: I don't think we should have comments in that we're not sure about. Let's look more inhttps://github.com/DiamondLightSource/python-artemis/issues/776

src/artemis/experiment_plans/optimise_attenuation_plan.py
```diff
@@ -58,35 +86,240 @@ def arm_devices(xspress3mini, zebra):
...
+        if deadtime >= deadtime_threshold:
+            direction = Direction.NEGATIVE
+    return direction
+
+
+def calculate_new_direction(direction: Direction, deadtime, deadtime_threshold):
+    if direction == Direction.POSITIVE:
+        if deadtime > deadtime_threshold:
+            direction = Direction.NEGATIVE
+    return direction
```
DominicOram, line 100: Should: These functions are basically the same, only one is used, can we remove the other?

DominicOram (approved):
One minor issue

src/artemis/experiment_plans/optimise_attenuation_plan.py
olliesilvester: Yes, looks like i accidentally (basically) duplicated that function

PR #758, opened 2023-06-26
Apply flatfield by default when changing zoom level

DominicOram:
Fixes https://github.com/DiamondLightSource/dodal/issues/103

This mainly just removes the need for Artemis to set the plugins for the OAV as this is handled inside the device when zoom level is changed

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/104

### To test:
1. Confirm tests still pass

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     xarray
     doct
     databroker
-    dodal @ git+https://github.com/DiamondLightSource/python-dodal.git@5cb5579b3eb29bdba8d5bf966ec965ed3aeba873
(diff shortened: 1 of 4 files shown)


noemifrisina (approved):

noemifrisina:
Merge conflict to be addressed before merging

noemifrisina (changes requested):
It looks like the failing tests are due to changes in Pydantic v2, which was just released.

I think we could pin the pydantic version to `pydantic<2.0` in `setup.cfg` for this one and fix the issue in a separate PR

DominicOram:
Broken test is due to https://github.com/DiamondLightSource/python-artemis/pull/763

noemifrisina (approved):
Thanks, looks okay aside from merge conflict

PR #772, opened 2023-06-30
Unstage detector if there is an issue in grid detection

### LINKED ISSUE #765: Disarm detector if there is an issue in grid detection
Given that we start arming before grid detection we should clean it up after if there is an issue during detection. There should also be a unit test for this

DominicOram:
Fixes #765

### To test:
1. Confirm unit tests pass

### DIFF
--- a/src/artemis/experiment_plans/tests/test_full_grid_scan_plan.py
+++ b/src/artemis/experiment_plans/tests/test_full_grid_scan_plan.py
@@ -14,6 +14,7 @@ from artemis.experiment_plans.full_grid_scan import (
     create_devices,
     detect_grid_and_do_gridscan,
     get_plan,
+    start_arming_then_do_grid,
     wait_for_det_to_finish_moving,
 )
 from artemis.parameters.plan_specific.grid_scan_with_edge_detect_params import (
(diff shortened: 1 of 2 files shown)


src/artemis/experiment_plans/tests/test_full_grid_scan_plan.py
```diff
@@ -146,3 +140,38 @@ def test_detect_grid_and_do_gridscan(
...
+                    backlight=MagicMock(),
+                    eiger=eiger,
+                    aperture_scatterguard=MagicMock(),
+                    detector_motion=MagicMock(),
+                    oav_params=OAVParameters("xrayCentring", **test_config_files),
+                )
+            )
+
+        # Check detector was armed
+        eiger.do_arm.set.assert_called_once_with(1)
```
noemifrisina, line 175: It looks like the exception was raised before the detector was armed as this fails with an `AssertionError` because`set` is never called.

```
 FAILED src/artemis/experiment_plans/tests/test_full_grid_scan_plan.py::test_grid_detection_running_when_exception_raised_then_eiger_unstaged - AssertionError: Expected 'set' to be called once. Called 0 times.
```

noemifrisina (changes requested):

DominicOram:
Works on my machine and in CI, could be `dodal` version? Can you check you're on at least https://github.com/DiamondLightSource/python-dodal.git@3106d1f2e4f97bde92720d115c8fbb827ad374b1 in `dodal`?

Issue #113, opened 2022-05-26
Position in syncweb is being calculated wrong


DominicOram:
In e.g. https://ispyb.diamond.ac.uk/dc/visit/cm31105-2/dcg/7450520 image 1 and image 40 should have the same x value. They currently have it set at -19 and +361

bentom08:
Probably something to do with the "snaked" ispyb field in each scan. Synchweb position coords seem to be assuming a non-snaked scan.

bentom08:
On further investigation this appears to be an issue with all grid scans in synchweb. Probably needs to be brought up with the synchweb devs to resolve.

bentom08:
Jira ticket raised with scisoft https://jira.diamond.ac.uk/browse/SCI-14169

DominicOram:
Given this is a synchweb issue I'm going to close it on our side

PR #785, opened 2023-07-10
632 get flux from pv

### LINKED ISSUE #632: Get flux from PV
GDA now gets it's flux from a PV (`BL03I-MO-MAPT-01:Y:FLUX`). Artemis should do the same, rather than get it passed in from GDA

noemifrisina:
Fixes #632 

Link to dodal PR (if required): [#113](https://github.com/DiamondLightSource/dodal/pull/113)

### To test:
1. Check that all tests still pass

### DIFF
--- a/src/artemis/system_tests/test_main_system.py
+++ b/src/artemis/system_tests/test_main_system.py
@@ -300,7 +300,7 @@ def test_cli_args_parse():
     assert test_args == ("DEBUG", True, True, True)
 
 
-@patch("dodal.beamlines.i03.flux")
+@patch("dodal.beamlines.i03.Flux")
 @patch("dodal.beamlines.i03.DetectorMotion")
 @patch("dodal.beamlines.i03.OAV")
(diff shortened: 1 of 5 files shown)


DominicOram (changes requested):
Good work, thank you. The test is failing because you need to patch the device in `test_when_blueskyrunner_initiated_then_plans_are_setup_and_devices_connected`, can you please add it in. We should add it in to S03 though but we can merge this without doing so

src/artemis/system_tests/test_main_system.py
```diff
@@ -300,6 +300,7 @@ def test_cli_args_parse():
     assert test_args == ("DEBUG", True, True, True)
 
 
+@patch("dodal.beamlines.i03.flux")
```
DominicOram: Must: You need to patch the Device, not teh function. I've fixed it :)

DominicOram (approved):
Great, thank you!

PR #786, opened 2023-07-11
779 fix fluo optimise

### LINKED ISSUE #779: Test fluo optimisation
Successfully tested on i03 today, found the following issues:
* Set transmission works
* Optimise doesn't actually use the zebra, manually opens/closes shutter
* Therefore added sample shutter in for i03 and used this instead of zebra (TODO: remove zebra from optimise)
* Acquire/Erase states needed to be set as numbers, not strings
* The comment about no callback until collection finished is wrong instead, logic is:
```python
arm_status = await_value_in_list(self.detector_state, self.detector_busy_states)
self.erase.put(EraseState.ERASE.value)
arm_status &= self.acquire.set(AcquireState.ACQUIRE.value)
arm_status.wait(1) # Detector now taking data
return await_value(self.acquire_rbv, 0) # This will return don when the collection is finished 
```
* Deadtime logic - need to wait for it to go above the threshold before flipping direction then returning when below
* Implement high limit (default 0.1) (for both algorithms)
* Finally, we got some reasonable results but they're not exactly the same as GDA. Further work is needed (outside this issue)to finish off.

## Acceptance Criteria
* The above is tidied up into a PR and merged (with tests)

olliesilvester:
Fixes #779 

Link to dodal PR (if required): [#114](https://github.com/DiamondLightSource/dodal/pull/114)

### To test:
1. Check issue requirements are all addressed
2. Confirm new tests run and are sensible

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     xarray
     doct
     databroker
-    dodal @ git+https://github.com/DiamondLightSource/python-dodal.git@2dbcb05d2ce6223880c24fca3c6b8a9b88b2ba03
+    dodal @ git+https://github.com/DiamondLightSource/python-dodal.git@b625b9e3abadc26b095e72701a008795f0c680c0
     pydantic<2.0 # See https://github.com/DiamondLightSource/python-artemis/issues/774
 
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     xarray
     doct
     databroker
-    dodal @ git+https://github.com/DiamondLightSource/python-dodal.git@3106d1f2e4f97bde92720d115c8fbb827ad374b1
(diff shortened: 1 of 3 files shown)


DominicOram (changes requested):
The tests are failing because `LOGGER` is a global static object, you have to be careful asserting against it as you can end up affecting other tests. A better way to do it is to setup a mock locking handler e.g. like in `test_isypb_handler`:
```python
@pytest.fixture
def mock_emit():
    import logging

    test_handler = logging.Handler()
    test_handler.emit = MagicMock()  # type: ignore
    LOGGER.addHandler(test_handler)

    yield test_handler.emit

    LOGGER.removeHandler(test_handler)


def test_log(mock_emit):
    LOGGER.info("Hello")
    latest_record = mock_emit.call_args.args[-1]
    assert latest_record.levelname == "INFO"
    assert latest_record.msg == "Hello"
```

PR #802, opened 2023-07-13
Correct the grid overlay drawing

### LINKED ISSUE #770: Confirm drawn overlay correct
When we draw the grid on the OAV we want the centre of each box to indicate where data was taken. When we supply the grid to the fast motion control program we supply the first point to be taken and the subsequent steps. We need to make sure these are as expected.

## Acceptance Criteria
* We reason about the overlay and see if we think it is correct
* We write a `test_grid_scan` plan that we can a single 2D scan against a foil and confirm positions

DominicOram:
Fixes #770

### To test:
1. Confirm tests are sensible and pass


noemifrisina (approved):
Tests passing now, thanks

PR #807, opened 2023-07-17
Increase timeout when restarting artemis

### LINKED ISSUE #761: Artemis not restarting
We had issues with it looking like Artemis did not restart. In fact it did start but took ~20s.

1) Increase timeout on restart
1) Add more logging to work out why it was taking so long - basically just log after every step of `run_artemis` into the start log with timestamps

noemifrisina:
Fixes #761


DominicOram (approved):
Great, thanks

PR #814, opened 2023-07-18
Ensure proper use of class variables

### LINKED ISSUE #810: Ensure we're not holding state in class variables
See issue fixed in https://github.com/DiamondLightSource/python-artemis/commit/b9d46d54c738e281310777bdad738c51c61498c4. We should go through the code and ensure we're not doing the same elsewhere, this is particularly a problem where we're e.g. appending but we should fix it more generally too

DominicOram:
Fixes #810

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/123

### To test:
1. Confirm class variables are removed and tests still pass
2. Confirm exceptions are listed in the issue

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     xarray
     doct
     databroker
-    dodal @ git+https://github.com/DiamondLightSource/python-dodal.git@b625b9e3abadc26b095e72701a008795f0c680c0
+    dodal @ git+https://github.com/DiamondLightSource/python-dodal.git@36e0efe08f7a741796654bb07119595f9bab3411
     pydantic<2.0 # See https://github.com/DiamondLightSource/python-artemis/issues/774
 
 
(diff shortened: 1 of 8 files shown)


dperl-dls (approved):
Nit: FGSComposite isn't really a singleton, we just only create it once, there's no enforcement for that. All the objects it holds are singletons though. Anyway, removing the class variables from FGSComposite doesn't change anything (they are all overwritten with instance vars on init) so you might as well remove those too if you feel like it

DominicOram:
> Nit: FGSComposite isn't really a singleton, we just only create it once, there's no enforcement for that. All the objects it holds are singletons though. Anyway, removing the class variables from FGSComposite doesn't change anything (they are all overwritten with instance vars on init) so you might as well remove those too if you feel like it

Yh, I think I meant that all the devices under it are singletons anyway so it's effectively a singleton. But I've changed it now anyway, thanks.

Tests are failing due to https://github.com/DiamondLightSource/python-artemis/issues/847

PR #817, opened 2023-07-19
remove dataclasses_json

### LINKED ISSUE #815: remove dataclasses_json (and maybe dataclasses)
currently we have a mixture of pydantic models with some legacy `dataclasses_json` here and there (notably `__main__.py`) - we should streamline this and remove the dependency on `dataclasses-json. this should basically just work as is by changing them to `pydantic.dataclass` es, and possibly changing some `.to_json()` to `.json()`

maybe we can also change other basic dataclasses but that's less important as they are in the standard lib and not a dependency

dperl-dls:
Fixes #815

### To test:
1. Run tests

### DIFF
--- a/src/artemis/parameters/plan_specific/grid_scan_with_edge_detect_params.py
+++ b/src/artemis/parameters/plan_specific/grid_scan_with_edge_detect_params.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from typing import Any
 
 import numpy as np
-from dataclasses_json import DataClassJsonMixin
 from dodal.devices.detector import TriggerMode
 from dodal.parameters.experiment_parameter_base import AbstractExperimentParameterBase
 from pydantic import validator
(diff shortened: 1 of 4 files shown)


src/artemis/parameters/plan_specific/grid_scan_with_edge_detect_params.py
```diff
@@ -18,7 +17,7 @@
 
 
 @dataclass
-class GridScanWithEdgeDetectParams(DataClassJsonMixin, AbstractExperimentParameterBase):
+class GridScanWithEdgeDetectParams(AbstractExperimentParameterBase):
```
DominicOram, line 20: Query: `RotationScanParams` inherits from `BaseModel`. Do we have a preference on when to do that and when to use `pydantic.dataclass`? From a google I think maybe `BaseModel` gives you more? So I would suggest we standardise on them all using `BaseModel`? This may be a different ticket...

DominicOram (approved):
Looks good to me, thanks. Query in code, take it or leave it as part of this issue

src/artemis/parameters/plan_specific/grid_scan_with_edge_detect_params.py
dperl-dls: Yes, maybe it's better to use models for all of them, I just changed the ones which were still dataclass because it was easier

DominicOram: Cool, I'll make an issue

PR #821, opened 2023-07-20
Control zoom and flatfield correctly in tip/grid detect

### LINKED ISSUE #820: Control zoom correctly
As of https://github.com/DiamondLightSource/dodal/pull/104, setting the zoom level on the device will apply/remove the flatfield. However, Artemis currently uses the underlying PV rather than setting it at the device level and so the flatfield is not getting set correctly.

DominicOram:
Fixes #820

### To test:
1. Confirm test passes


noemifrisina (approved):
Looks okay

PR #827, opened 2023-07-24
Blueapi: Use bluesky context for plans

### LINKED ISSUE #826: Blueapi: Store plans in context rather than plan registry
In the usual use of Blueapi the context is what stores the plans, lets use that rather than the registry

## Acceptance Criteria
* We use `blueapi`'s introspection mechanism to get put the plans in the context
* We remove the plans from the experiment registry (keeping everything else there for now)
* We use the same mechanism to run the plans, just pull them out of the context instead

DominicOram:
Fixes #826

### To test:
1. Confirm tests still pass
2. Confirm you can `run_artemis` against the simulator, write to an endpoint and that this starts a plan. (you may need to comment out some of the devices that are not yet in s03)


DominicOram:
Some notes:
* This doesn't necessarily make our code any nicer but it's a step to using `blueapi` in general, which will make things nicer
* The use of the `Any` type in the plans is a temporary thing until we're using `blueapi` properly

src/artemis/system_tests/test_main_system.py
```diff
@@ -467,3 +463,9 @@ def test_warn_exception_during_plan_causes_warning_in_log(
     assert response_json["message"] == 'WarningException("D\'Oh")'
     assert response_json["exception_type"] == "WarningException"
     assert caplog.records[-1].levelname == "WARNING"
+
+
+def test_when_context_created_then_contains_expected_number_of_plans():
```
Tom-Willemsen, line 468: Isn't this a bit needlessly brittle? Any time we add a plan we'll now need to adjust the expected number of plans here?

Some options to consider:
- Check that a few known plans are in the `plan_functions` list
- Mock out some of the import machinery so that we can mock the expected plans in the test, and therefore they won't rely on the actual plans

src/artemis/__main__.py
```diff
@@ -192,6 +192,15 @@ def put(self, plan_name: str, action: Actions):
         return asdict(status_and_message)  # type: ignore
 
 
+def setup_context() -> BlueskyContext:
+    context = BlueskyContext()
+
+    import artemis.experiment_plans as artemis_plans
```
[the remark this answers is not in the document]
DominicOram: That wasn't the intention. Not sure why I did this, it may have been cargo-culted from the PR this is based on. Moved it.

src/artemis/system_tests/test_main_system.py
DominicOram: Good spot, thanks!

setup.cfg
```diff
@@ -19,6 +19,7 @@ package_dir =
 install_requires =
     bluesky
     pyepics
+    blueapi
```
dperl-dls, line 22: should update this to github main, although that may be complicated by pinned dependency conflicts

dperl-dls (approved):
This basically all makes sense, as it's an intermediate step I don't want to get too far in the weeds on it, tests pass and external calls still trigger the plan correctly

setup.cfg
DominicOram: I think we fix this properly by doing https://github.com/DiamondLightSource/dodal/issues/145

PR #831, opened 2023-07-28
369 rotation call zocalo

### LINKED ISSUE #369: Confirm Zocalo handler deal with rotation scans

dperl-dls:
Fixes #369 

adds a simple callback to trigger `run_start` and `run_end` in zocalo for rotation scans

### To test:
1. Run tests

### DIFF
--- a/src/artemis/external_interaction/callbacks/rotation/zocalo_callback.py
+++ b/src/artemis/external_interaction/callbacks/rotation/zocalo_callback.py
@@ -33,7 +33,7 @@ class RotationZocaloCallback(CallbackBase):
             LOGGER.info(
                 f"Zocalo handler received stop document, for run {doc.get('run_start')}."
             )
-            if self.ispyb.ispyb_ids[0]:
+            if self.ispyb.ispyb_ids[0] is not None:
                 self.zocalo_interactor.run_end(self.ispyb.ispyb_ids[0])
             else:
                 raise ISPyBDepositionNotMade("ISPyB deposition was not initialised!")
--- a/src/artemis/external_interaction/callbacks/rotation/zocalo_callback.py
+++ b/src/artemis/external_interaction/callbacks/rotation/zocalo_callback.py
@@ -11,6 +11,10 @@ from artemis.log import LOGGER
 
 
 class RotationZocaloHandlerCallback(CallbackBase):
+    """Simple callback which sends the ISPyB IDs for a rotation data collection to
+    zocalo. Both run_start() and run_end() are sent when the collection is done.
+    Triggers on the 'stop' document for 'rotation_scan_main'."""
+
     def __init__(
         self,
         zocalo_environment: str,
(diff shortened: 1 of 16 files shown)


src/artemis/external_interaction/callbacks/rotation/zocalo_callback.py
```diff
@@ -0,0 +1,36 @@
...
+            self.run_uid = doc.get("uid")
+
+    def stop(self, doc: dict):
+        if doc.get("run_start") == self.run_uid:
+            LOGGER.info(
+                f"Zocalo handler received stop document, for run {doc.get('run_start')}."
+            )
+            if self.ispyb.ispyb_ids[0] is not None:
+                self.zocalo_interactor.run_start(self.ispyb.ispyb_ids[0])
+                self.zocalo_interactor.run_end(self.ispyb.ispyb_ids[0])
```
DominicOram, line 34: Should: For grid scans we start this in `start`, why not here too?

src/artemis/external_interaction/callbacks/rotation/zocalo_callback.py
```diff
@@ -0,0 +1,36 @@
...
+
+from artemis.external_interaction.callbacks.rotation.ispyb_callback import (
+    RotationISPyBHandlerCallback,
+)
+from artemis.external_interaction.exceptions import ISPyBDepositionNotMade
+from artemis.external_interaction.zocalo.zocalo_interaction import ZocaloInteractor
+from artemis.log import LOGGER
+
+
+class RotationZocaloHandlerCallback(CallbackBase):
```
DominicOram, line 13: Should: A bit of a docstring would be good, this is probably vaguely user facing

DominicOram (changes requested):
Looks good, minor comments. I think there may be a possibility to share more with the FGS one but I actually like keeping it a bit separate in this case

src/artemis/external_interaction/callbacks/rotation/zocalo_callback.py
dperl-dls: That's what I did first; however, start is before ispyb data is read, and so there is no deposition and no IDs yet. I thought about adding yet another layer to the plan to manage this, but since the result of this isn't time-sensitive, I thought it wasn't worth cluttering up the plan over.

DominicOram: Can we make an issue to think about it maybe? I guess it doesn't matter if we're not doing parallel analysis but if we move to parallel analysis at some point it would be nice if it just worked

dperl-dls: I don't think it will ever be necessary, but we should discuss with scientists

DominicOram (approved):
Great, thank you!

PR #842, opened 2023-07-31
Fix issue with transmission being set incorrectly

### LINKED ISSUE #835: Artemis setting transmission incorrectly
Over the weekend Artemis set the transmission to 10,000% and wrote 1% into ispyb. Transmission should be 100% in both.

## Acceptance Criteria
* Work out where the conversion error is coming from
* Rename all variables (particularly in the external APIs) to be clear about whether they are percentage or decimal

DominicOram:
Fixes #835

Review GDA too https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/39641

Also see https://github.com/DiamondLightSource/dodal/pull/133

### To test:
1. Run tests
1. Confirm GDA calls Artemis correctly still using the above gerrit change see instructions [here](https://github.com/DiamondLightSource/python-artemis/wiki/Running-a-Dummy-Mode-GDA-against-Artemis)

### DIFF
--- a/src/artemis/experiment_plans/full_grid_scan.py
+++ b/src/artemis/experiment_plans/full_grid_scan.py
@@ -92,7 +92,7 @@ def start_arming_then_do_grid(
     yield from bps.abs_set(eiger.do_arm, 1, group="ready_for_data_collection")
     yield from bps.abs_set(
         attenuator,
-        parameters.artemis_params.ispyb_params.transmission,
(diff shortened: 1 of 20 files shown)


DominicOram:
Linting issues fixed in https://github.com/DiamondLightSource/python-artemis/pull/841

dperl-dls (approved):
Great. Quite apart from fixing a bug, I love to see units in names.

PR #846, opened 2023-08-01
Allow specifying the grid size in microns

### LINKED ISSUE #845: Allow setting grid scan width in microns
The width of grid scan to do for the pin should be settable as it changes for multipin and single pin. It makes sense to set this in microns rather than pixels to avoid the dependency on what zoom level you're at.

DominicOram:
Fixes #845

### To test:
1. Confirm tests still pass

### DIFF
--- a/live_test_rotation_params.json
+++ b/live_test_rotation_params.json
@@ -1,5 +1,5 @@
 {
-    "params_version": "2.1.0",
+    "params_version": "2.2.0",
     "artemis_params": {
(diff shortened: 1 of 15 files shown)


dperl-dls (approved):
probably should update params version, otherwise great

DominicOram:
Given parameter change could you review https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/39676 too

PR #855, opened 2023-08-07
834 separate setup from rotation scan

### LINKED ISSUE #834: Take out setup from rotation scans into an outer plan
There should be a plan which is *just* the things required for a rotation scan, assuming everything is in place
- `move_xyz` should be combined with the one used in gridscan, and moved out one plan layer
- setup sample environment too

dperl-dls:
Fixes #834

### To test:
1. Run tests

### DIFF
--- a/src/artemis/experiment_plans/fast_grid_scan_plan.py
+++ b/src/artemis/experiment_plans/fast_grid_scan_plan.py
@@ -27,6 +27,7 @@ from dodal.devices.eiger import DetectorParams
 from dodal.devices.fast_grid_scan import set_fast_grid_scan_params
 
 import artemis.log
+from artemis.device_setup_plans.manipulate_sample import move_x_y_z
 from artemis.device_setup_plans.read_hardware_for_setup import read_hardware_for_ispyb
 from artemis.device_setup_plans.setup_zebra import (
     set_zebra_shutter_to_manual,
@@ -117,28 +118,6 @@ def set_aperture_for_bbox_size(
         yield from bps.abs_set(aperture_device, aperture_size_positions)
 
     yield from set_aperture()
-
-
-@bpp.set_run_key_decorator("move_xyz")
-@bpp.run_decorator(md={"subplan_name": "move_xyz"})
-def move_xyz(
-    sample_motors,
-    xray_centre_motor_position: np.ndarray,
-    md={
-        "plan_name": "move_xyz",
-    },
-):
-    """Move 'sample motors' to a specific motor position (e.g. a position obtained
-    from gridscan processing results)"""
-    artemis.log.LOGGER.info(f"Moving Smargon x, y, z to: {xray_centre_motor_position}")
-    yield from bps.mv(
-        sample_motors.x,
-        xray_centre_motor_position[0],
-        sample_motors.y,
-        xray_centre_motor_position[1],
-        sample_motors.z,
-        xray_centre_motor_position[2],
-    )
 
 
 def wait_for_fgs_valid(fgs_motors: FastGridScan, timeout=0.5):
@@ -254,10 +233,7 @@ def run_gridscan_and_move(
     # once we have the results, go to the appropriate position
     artemis.log.LOGGER.info("Moving to centre of mass.")
     with TRACER.start_span("move_to_result"):
-        yield from move_xyz(
-            fgs_composite.sample_motors,
-            xray_centre,
-        )
+        yield from move_x_y_z(fgs_composite.sample_motors, *xray_centre)
 
 
 def get_plan(
--- a/src/artemis/experiment_plans/rotation_scan_plan.py
+++ b/src/artemis/experiment_plans/rotation_scan_plan.py
@@ -15,10 +15,16 @@ from dodal.devices.zebra import RotationDirection, Zebra
 from ophyd.device import Device
 from ophyd.epics_motor import EpicsMotor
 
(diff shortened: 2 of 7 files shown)


src/artemis/experiment_plans/tests/test_rotation_scan_plan.py
```diff
@@ -358,11 +391,11 @@ def test_rotation_plan_smargon_settings(setup_and_run_rotation_plan_for_tests_st
     assert smargon.x.user_readback.get() == expt_params.x
     assert smargon.y.user_readback.get() == expt_params.y
     assert smargon.z.user_readback.get() == expt_params.z
-    assert omega_vel_set.call_count == 3
-    omega_vel_set.assert_has_calls(
-        [call(DEFAULT_MAX_VELOCITY), call(rotation_speed), call(DEFAULT_MAX_VELOCITY)]
+    assert omega_set.call_count == 6
+    omega_set.assert_has_calls(
+        [call(DEFAULT_MAX_VELOCITY), call(rotation_speed), call(DEFAULT_MAX_VELOCITY)],
+        any_order=True,
```
DominicOram, line 397: Should: As I put in `conftest` this is an example of where having these as one mock is poor

src/artemis/experiment_plans/tests/conftest.py
```diff
@@ -178,15 +194,45 @@ def fake_create_devices(
...
+    eiger.stage = MagicMock()
+    eiger.unstage = MagicMock()
+    mock_omega_sets = MagicMock(return_value=Status(done=True, success=True))
+
+    mock_arm_disarm = MagicMock(
+        side_effect=zebra.pc.arm.armed.set, return_value=Status(done=True, success=True)
+    )
+    zebra.pc.arm.set = mock_arm_disarm
+    smargon.omega.velocity.set = mock_omega_sets
+    smargon.omega.set = mock_omega_sets
```
DominicOram, line 224: Should: This feels wrong, it makes the tests much more complicated as it's now one mock for both position and velocity. I think separate the mocks then you can assert things more nicely

src/artemis/experiment_plans/tests/test_fast_grid_scan_plan.py
```diff
@@ -187,43 +187,70 @@ def test_results_adjusted_and_passed_to_move_xyz(
...
     move_aperture.assert_has_calls(
-        [call_large, call_large, call_medium], any_order=True
+        [ap_call_large, ap_call_large, ap_call_medium], any_order=True
+    )
+
+    mv_call_large = call(
+        fake_fgs_composite.sample_motors, 0.05, 0.15000000000000002, 0.25
+    )
+    mv_call_medium = call(
+        fake_fgs_composite.sample_motors, 0.05, 0.15000000000000002, 0.25
```
DominicOram, line 205: Should: I think `pytest.approx(0.15)` is more readable than `0.15000000000002`

src/artemis/experiment_plans/rotation_scan_plan.py
```diff
@@ -282,7 +239,22 @@ def rotation_scan_plan_with_stage_and_cleanup(
...
+                params.artemis_params.detector_params.detector_distance,
+            )
+            LOGGER.info("moving to position (if specified)")
+            yield from move_x_y_z(
+                devices["smargon"],
+                params.experiment_params.x,
+                params.experiment_params.y,
+                params.experiment_params.z,
+            )
             LOGGER.info("setting up and staging eiger...")
```
DominicOram, line 258: Should: This comment is in the wrong place now, we started staging the eiger way before this

src/artemis/experiment_plans/fast_grid_scan_plan.py
```diff
@@ -254,10 +233,7 @@ def run_gridscan_and_move(
     # once we have the results, go to the appropriate position
     artemis.log.LOGGER.info("Moving to centre of mass.")
     with TRACER.start_span("move_to_result"):
-        yield from move_xyz(
-            fgs_composite.sample_motors,
-            xray_centre,
-        )
+        yield from move_x_y_z(fgs_composite.sample_motors, *xray_centre)
```
DominicOram, line 236: Must: The previous function waited, this now doesn't I think we probably want to still

src/artemis/experiment_plans/rotation_scan_plan.py
```diff
@@ -282,7 +239,22 @@ def rotation_scan_plan_with_stage_and_cleanup(
...
+            LOGGER.info("setting up sample environment...")
+            yield from setup_sample_environment(
+                devices["detector_motion"],
+                devices["backlight"],
+                devices["attenuator"],
+                params.artemis_params.ispyb_params.transmission_fraction,
+                params.artemis_params.detector_params.detector_distance,
+            )
+            LOGGER.info("moving to position (if specified)")
+            yield from move_x_y_z(
```
DominicOram, line 252: Must: This is never waited on, feels like it could cause issues

DominicOram (changes requested):
Great, I think this cleans things up a bit. Some comments in code

src/artemis/experiment_plans/fast_grid_scan_plan.py
DominicOram: I still think we probably want to do this

DominicOram (approved):
Looks good, thanks

PR #890, opened 2023-09-11
Idea: use `ruff` as a replacement for `flake8` + `isort` ?

Tom-Willemsen:
Test for adding `ruff` as a replacement for `flake8` + `isort`.

Main advantage is speed (approx 100x faster):

```
$ time isort --check .

real	0m0.420s
user	0m0.349s
sys	0m0.040s

$ time flake8 src/

real	0m0.859s
user	0m2.157s
sys	0m0.154s

$ time ruff .

real	0m0.011s
user	0m0.006s
sys	0m0.006s
```

One code change is required due to a difference in opinion between ruff and `isort` about constructs like:

```python
from somewhere import thing1
from somewhere import thing2 as aliased_name
from somewhere import thing3
```
which ruff prefers formatted as:
```python
from somewhere import thing1, thing3
from somewhere import thing2 as aliased_name
```

Other than that it is a 1:1 replacement in terms of linting rules.

This seems to be the direction a lot of modern python projects are moving in, and a better dev experience with `pre-commit` etc is always nice.

### DIFF
--- a/.github/pages/make_switcher.py
+++ b/.github/pages/make_switcher.py
@@ -59,7 +59,7 @@ def get_versions(ref: str, add: Optional[str], remove: Optional[str]) -> List[st
 def write_json(path: Path, repository: str, versions: str):
     org, repo_name = repository.split("/")
     struct = [
-        dict(name=version, url=f"https://{org}.github.io/{repo_name}/{version}/")
+        {"name": version, "url": f"https://{org}.github.io/{repo_name}/{version}/"}
         for version in versions
     ]
     text = json.dumps(struct, indent=2)
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -7,33 +7,26 @@ on:
     # Run every Monday at 8am to check latest versions of dependencies
     - cron: "0 8 * * WED"
 
-
 jobs:
   lint:
-     # pull requests are a duplicate of a branch push if within the same repo.
-     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.repository
-     runs-on: ubuntu-latest
-     steps:
-        - name: Setup Python
-          uses: actions/setup-python@v4
-          with:
-            python-version: '3.10'
-            architecture: x64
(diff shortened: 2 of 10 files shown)


DominicOram (commented):
This worked for me but required a bit of faffing to get vscode happy. Had to change `Python: Select Linter` to None and install the ruff extension. @dperl-dls, do you have strong opinions on this? One issue is that we're now using a different formatter to `dodal`.

Tom-Willemsen:
> One issue is that we're now using a different formatter to dodal.

I agree if we decide to use this we should also do an equivalent migration on `dodal` - I only opened the one PR initially in order to get opinions and not waste too much time if we decide we don't like this.

DominicOram:
> > One issue is that we're now using a different formatter to dodal.
> 
> I agree if we decide to use this we should also do an equivalent migration on `dodal` - I only opened the one PR initially in order to get opinions and not waste too much time if we decide we don't like this.

In which case we should probably ask @callumforrester

dperl-dls:
I don't have a very strong opinion on this but I have noticed some slowdown with flake8 so I am mildly in favour

callumforrester:
@coretl, weren't you looking at migrating the skeleton to ruff at some point?

coretl:
we are planning to, but haven't had time yet: https://github.com/DiamondLightSource/python3-pip-skeleton/issues/142

DominicOram:
Is a better way to do this to change the skeleton then merge it in here and `dodal` then?

coretl:
@Tom-Willemsen just volunteered to do this and I said yes please...

Tom-Willemsen:
See https://github.com/DiamondLightSource/python3-pip-skeleton/pull/147

Tom-Willemsen:
This PR is now a cherry-pick/merge of https://github.com/DiamondLightSource/python3-pip-skeleton/pull/147 .

DominicOram (approved):
Looks good to me, thanks. Can we merge it into dodal too please?

PR #901, opened 2023-09-20
734 grid using callbacks

### LINKED ISSUE #734: Get grid parameters out of grid detection using callbacks
Following https://github.com/DiamondLightSource/python-artemis/pull/731 the snapshot info has been moved into callbacks but we should do the same for setting the grid scan parameters

subinsaji:
Fixes #734 

Tests using:
```python
def _fake_grid_detection(
    parameters: OAVParameters,
    snapshot_template: str,
    snapshot_dir: str,
    grid_width_microns: float = 0,
    box_size_um: float = 0.0
```
have been skipped by using `@pytest.mark.skip` as `_fake_grid_detection` does not have real enough parameters to pass the tests. I will raise a new issue for this.

### To test:
1. Confirm (relevant) tests passes

### DIFF
--- a/src/hyperion/experiment_plans/tests/test_grid_detection_plan.py
+++ b/src/hyperion/experiment_plans/tests/test_grid_detection_plan.py
@@ -1,6 +1,5 @@
 from unittest.mock import MagicMock, call, patch
 
-import numpy as np
 import pytest
 from bluesky.run_engine import RunEngine
 from dodal.beamlines import i03
@@ -49,8 +48,8 @@ def fake_devices(smargon: Smargon, backlight: Backlight):
         mock_image = MagicMock()
         mock_image_class.open.return_value = mock_image
         yield oav, smargon, backlight, mock_image
-        
-        
+
+
 @patch("dodal.beamlines.beamline_utils.active_device_is_same_type", lambda a, b: True)
 @patch("bluesky.plan_stubs.wait")
 def test_grid_detection_plan_runs_and_triggers_snapshots(
@@ -146,7 +145,6 @@ def test_given_when_grid_detect_then_upper_left_and_start_position_as_expected(
 
     cb = OavSnapshotCallback()
     RE.subscribe(cb)
-
     RE(
         grid_detection_plan(
             parameters=params,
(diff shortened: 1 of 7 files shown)


src/hyperion/experiment_plans/tests/test_grid_detection_plan.py
```diff
@@ -158,9 +156,9 @@ def test_given_when_grid_detect_then_upper_left_and_start_position_as_expected(
     assert cb.out_upper_left[0] == [8, 2]
     assert cb.out_upper_left[1] == [8, 2]
 
-    assert gridscan_params.x_start == 0.0005
-    assert gridscan_params.y1_start == -0.0001
-    assert gridscan_params.z1_start == -0.0001
+    assert gridscan_params.x_start == 0.1
```
dperl-dls, line 159: Why did these values change?

dperl-dls (approved):
Looks great. Because we have discussed this so much in person I don't think there's any more to do for this change.

DominicOram (commented):
Looks great, some comments in there

src/hyperion/external_interaction/callbacks/oav_snapshot_callback.py
```diff
@@ -6,8 +6,20 @@ def __init__(self, *args) -> None:
         super().__init__(*args)
         self.snapshot_filenames: list = []
         self.out_upper_left: list = []
+        self.box_widths: list = []
+        self.no_of_boxes: list = []
```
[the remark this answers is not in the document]
subinsaji: I just did a quick test and debug. They are not used so it will be removed

src/hyperion/experiment_plans/tests/test_grid_detection_plan.py
subinsaji: `gridscan_params = GridScanParams()` is located in fast_grid_scan.py. 
The values for `x_start`, `y1_start` and `z1_start` are defined in the `GridScanParams` class to begin with.
I change it to 0.1 since that is what the test is expecting in order to pass.
Is it wise to change the values of 0.0005, -0.0001, -0.0001 for `x_start`, `y1_start` and `z1_start` respectively in  `GridScanParams()`?

dperl-dls: No, that won't fix the problem.
Previously, the default parameter of `x_start = 0.1` in GridScanParams was modified in `grid_detection_plan` in line 167: `  out_parameters.x_start = start_positions[0][0]`, so the default value wasn't used.
What is different about the way `start_positions[0][0]` used to be calculated in `grid_detection_plan` compared to how `GridDetectionCallback` calculates `start_positions` now?

PR #907, opened 2023-09-28
#906 check limits during pin tip centring

### LINKED ISSUE #906: move_pin_into_view moves out of smargon x limits
when the pin is too long 'move_pin_into_view' moves out of the smargon x limit during OAV centring

https://ispyb.diamond.ac.uk/dc/visit/mx32736-17/id/11772912
https://graylog2.diamond.ac.uk/messages/logs_8415/81e663d1-5d84-11ee-ae6d-1866dafadc6c

we should check the limits and raise a warningexception to exit the experiment gracefully

dperl-dls:
- move to the limit if normal nudge would take it over since into-view steps are large
- update tests with limits

Fixes #906

### To test:
1. Run tests

### DIFF
--- a/src/hyperion/experiment_plans/pin_tip_centring_plan.py
+++ b/src/hyperion/experiment_plans/pin_tip_centring_plan.py
@@ -57,27 +57,29 @@ def move_pin_into_view(
 
         if tip_x_px == 0:
             min_limit = min(smargon.x.limits)
-            if float(smargon.x.user_readback.get()) - step_size_mm < min_limit:
+            smargon_x = yield from bps.rd(smargon.x.user_readback)
+            if float(smargon_x) - step_size_mm < min_limit:
                 LOGGER.warning(
-                    f"Pin is too long, moving -{step_size_mm} mm would cross limits, "
+                    f"Pin tip is off screen, and moving -{step_size_mm} mm would cross limits, "
                     f"moving to minimum limit {min_limit}"
                 )
                 yield from bps.mv(smargon.x, min_limit)
                 yield from bps.sleep(OAV_REFRESH_DELAY)
                 break
-            LOGGER.warning(f"Pin is too long, moving -{step_size_mm}mm")
+            LOGGER.warning(f"Pin tip is off screen, moving -{step_size_mm} mm")
             yield from bps.mvr(smargon.x, -step_size_mm)
         elif tip_x_px == oav.mxsc.pin_tip.INVALID_POSITION[0]:
+            smargon_x = yield from bps.rd(smargon.x.user_readback)
             max_limit = max(smargon.x.limits)
-            if float(smargon.x.user_readback.get()) - step_size_mm > max_limit:
+            if float(smargon_x) + step_size_mm > max_limit:
                 LOGGER.warning(
-                    f"Pin is too short; moving {step_size_mm} mm would cross limits, "
+                    f"Pin tip is off screen, and moving {step_size_mm} mm would cross limits, "
                     f"moving to maximum limit {max_limit}"
                 )
                 yield from bps.mv(smargon.x, max_limit)
                 yield from bps.sleep(OAV_REFRESH_DELAY)
                 break
-            LOGGER.warning(f"Pin is too short, moving {step_size_mm}mm")
+            LOGGER.warning(f"Pin tip is off screen, moving {step_size_mm}mm")
             yield from bps.mvr(smargon.x, step_size_mm)
         else:
             return (tip_x_px, tip_y_px)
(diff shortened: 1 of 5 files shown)


src/hyperion/parameters/constants.py
```diff
@@ -14,6 +14,7 @@
 # this one is for making depositions:
 DEV_ISPYB_DATABASE_CFG = "/dls_sw/dasc/mariadb/credentials/ispyb-dev.cfg"
 PARAMETER_SCHEMA_DIRECTORY = "src/hyperion/parameters/schemas/"
+OAV_REFRESH_DELAY = 0.3
```
DominicOram, line 17: Nit: I feel like this should be in the dodal device?

src/hyperion/experiment_plans/pin_tip_centring_plan.py
```diff
@@ -49,16 +56,34 @@ def move_pin_into_view(
...
-            LOGGER.warning(f"Pin is too long, moving -{step_size}mm")
-            yield from bps.mvr(smargon.x, -step_size)
+            min_limit = min(smargon.x.limits)
+            if float(smargon.x.user_readback.get()) - step_size_mm < min_limit:
+                LOGGER.warning(
+                    f"Pin is too long, moving -{step_size_mm} mm would cross limits, "
+                    f"moving to minimum limit {min_limit}"
+                )
+                yield from bps.mv(smargon.x, min_limit)
+                yield from bps.sleep(OAV_REFRESH_DELAY)
```
DominicOram, line 66: Should: Given we're doing this sleep again below does this mean it sleeps twice? I don't think that's necessary

src/hyperion/experiment_plans/pin_tip_centring_plan.py
```diff
@@ -49,16 +56,34 @@ def move_pin_into_view(
         tip_x_px, tip_y_px = yield from bps.rd(oav.mxsc.pin_tip)
 
         if tip_x_px == 0:
-            LOGGER.warning(f"Pin is too long, moving -{step_size}mm")
-            yield from bps.mvr(smargon.x, -step_size)
+            min_limit = min(smargon.x.limits)
+            if float(smargon.x.user_readback.get()) - step_size_mm < min_limit:
```
DominicOram, line 60: Must: We should be using `rd` in the plan, not `get`

src/hyperion/experiment_plans/pin_tip_centring_plan.py
```diff
@@ -49,16 +56,34 @@ def move_pin_into_view(
...
+                yield from bps.mv(smargon.x, min_limit)
+                yield from bps.sleep(OAV_REFRESH_DELAY)
+                break
+            LOGGER.warning(f"Pin is too long, moving -{step_size_mm}mm")
+            yield from bps.mvr(smargon.x, -step_size_mm)
         elif tip_x_px == oav.mxsc.pin_tip.INVALID_POSITION[0]:
-            LOGGER.warning(f"Pin is too short, moving {step_size}mm")
-            yield from bps.mvr(smargon.x, step_size)
+            max_limit = max(smargon.x.limits)
+            if float(smargon.x.user_readback.get()) - step_size_mm > max_limit:
```
DominicOram, line 72: Must: This should be a `+` not a `-`

DominicOram (changes requested):
Comments in code. Also, as a could, I feel like we might be able to make this cleaner by doing something like:
* Invert step size if too long (not too short)
* Calculate ideal absolute move ourselves
* Clip to limits
* Do the absolute move

src/hyperion/experiment_plans/pin_tip_centring_plan.py
DominicOram: Must: Also, can we get a test to cover this?

src/hyperion/experiment_plans/pin_tip_centring_plan.py
dperl-dls: no, break gets out of the for loop so we won't hit the other sleep

DominicOram: Yep, sorry, my bad. I think this leads credence to the general comment of finding a slightly nicer way of doing it then. Given that we're repeating this line and the logic isn't 100% obvious

DominicOram (approved):
Looks good, thank you. I still think it could be a bit cleaner but it's a nit pick

dperl-dls:
I agree it could be cleaner, and I'll be happy if someone improves it, but it works correctly and I don't have time right now

PR #915, opened 2023-10-03
905 851 rotation scan improvements

### LINKED ISSUE #851: Don't use both wavelength and energy
we should supply only energy (, in experiment params,) and calculate wavelength only at the last moment

### LINKED ISSUE #905: Issues identified during rotation scan testing
- several changes which were previously applied to live artemis have not been merged into hyperion - including the run number in nexus filenames and ispyb depositions, auxillary angles should be optional in json schema
- many ispybparams fields should be optional too

dperl-dls:
Fixes #905 and fixes #851

Mostly changes made on the beamline as part of testing, of which most are ISPyB deposition fixes and one hack of OAV snapshot paths - issue to fix here (https://github.com/DiamondLightSource/hyperion/issues/914)

Also includes no longer adding wavelength as external parameter and instead calculating it from energy as I didn't want to find it in GDA to send...

### To test:
1. Run tests


src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
```diff
@@ -37,6 +37,11 @@ def __init__(self, parameters: GridscanInternalParameters):
     def append_to_comment(self, comment: str):
         self._append_to_comment(self.ispyb_ids[0], comment)
 
+    def start(self, doc: dict):
```
DominicOram, line 40: Should: I think a test on this would be good

src/hyperion/external_interaction/callbacks/rotation/zocalo_callback.py
```diff
@@ -26,7 +26,7 @@ def __init__(
 
     def start(self, doc: dict):
         LOGGER.info("Zocalo handler received start document.")
-        if doc.get("subplan_name") == "rotation_scan_main":
+        if self.run_uid is None:
```
DominicOram, line 29: Must: This is all about making sure the `isypb` deposition happens before `zocalo` triggers, right? I think we need a very explicit test that confirms this, otherwise we'll end up accidentally hitting it again

src/hyperion/external_interaction/ispyb/store_in_ispyb.py
```diff
@@ -34,13 +34,96 @@
...
+#     "voltage",
+#     "objaperture",
+#     "c1aperture",
+#     "c2aperture",
+#     "c3aperture",
+#     "c1lens",
+#     "c2lens",
+#     "c3lens",
+# ]
+#
```
DominicOram, line 118: Must: Remove this please

src/hyperion/unit_tests/test_utils.py
```diff
@@ -0,0 +1,20 @@
...
+        assert convert_eV_to_angstrom(test_energies[i]) == pytest.approx(
+            test_wavelengths[i]
+        )
+
+
+def test_a_to_ev_converter():
+    for i in range(len(test_wavelengths)):
+        assert convert_angstrom_to_eV(test_wavelengths[i]) == pytest.approx(
+            test_energies[i]
+        )
```
DominicOram, line 20: Nit: It's nicer to use `paramterize` as it's more obvious which one has failed if only one has

DominicOram (changes requested):
Thanks, some comments in code. But otherwise works well

DominicOram (approved):
Great, thank you!

PR #917, opened 2023-10-04
913 flux is not being read from hardware

### LINKED ISSUE #913: Flux is not being read from the hardware
Due to a bug in https://github.com/DiamondLightSource/hyperion/issues/632 we're actually still using the GDA provided flux, rather than reading it from the hardware. You can see this if you change `flux_test_value` in `test_read_hardware_for_ispyb_updates_from_ophyd_devices` to be any value other than 10.0

subinsaji:
Fixes #913
Added `flux_flux_reading` into **test_flyscan_xray_centre_plan.py** mock subscriptions and **ispyb_callback_base.py** so flux can be properly read.

### To test:
1. Run tests

### DIFF
--- a/src/hyperion/experiment_plans/tests/test_flyscan_xray_centre_plan.py
+++ b/src/hyperion/experiment_plans/tests/test_flyscan_xray_centre_plan.py
@@ -156,6 +156,7 @@ def test_results_adjusted_and_passed_to_move_xyz(
                 "s4_slit_gaps_xgap": 0,
                 "s4_slit_gaps_ygap": 0,
                 "attenuator_actual_transmission": 0,
+                "flux_flux_reading": 10,
             },
         }
     )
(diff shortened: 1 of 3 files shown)


src/hyperion/experiment_plans/tests/test_flyscan_xray_centre_plan.py
```diff
@@ -327,6 +329,7 @@ def test_logging_within_plan(
                 "s4_slit_gaps_xgap": 0,
                 "s4_slit_gaps_ygap": 0,
                 "attenuator_actual_transmission": 0,
+                "flux_flux_reading": 10,
```
DominicOram, line 332: Could: It's annoying that this dictionary is basically the same in 3 places, can you pull it into a constant?

DominicOram (approved):
Brilliant, thank you! Minor suggestion in code but otherwise good.

PR #923, opened 2023-10-06
913 Created a single dictionary that is pulled from a constant

### LINKED ISSUE #913: Flux is not being read from the hardware

subinsaji:
Fixes #913 
Made a dictionary `mock_subscription_dict` that is pulled from a constant and tidying that up.

### To test:
1. Run Tests


DominicOram (approved):
Great, thank you, much cleaner

PR #925, opened 2023-10-09
Add a check topup plan

### LINKED ISSUE #127: Create a check topup plan

noemifrisina:
Fixes #127 
Adds a small plan to check the topup and sleep until it's done if too soon.

Link to dodal PR (if required): [#190](https://github.com/DiamondLightSource/dodal/pull/190)

### To test:
1. Check that topup plan is called in a sensible location
2. Check that tests make sense and pass


src/hyperion/experiment_plans/rotation_scan_plan.py
```diff
@@ -199,6 +200,13 @@ def rotation_scan_plan(
 
     yield from arm_zebra(zebra)
 
+    # Check topup gate
+    yield from check_topup_and_wait_if_necessary(
+        i03.synchrotron(),
+        detector_params,
+        ops_time=10.0,  # Additional time to account for rotation, is s
```
Tom-Willemsen, line 207: I think this should come from some kind of property/parameter long-term? It may be different across different beamlines which each eventually use this plan. Or alternatively scientists/developers may need to adjust this periodically. i.e. I don't think this should be hardcoded. Can be spun into an issue if you prefer.

src/hyperion/experiment_plans/rotation_scan_plan.py
```diff
@@ -199,6 +200,13 @@ def rotation_scan_plan(
...
+    )  # TODO figure out a correct value for opts_time
```
Tom-Willemsen, line 208: Can you create a corresponding github issue for this TODO and link it here? Otherwise the todo will get lost.

src/hyperion/device_setup_plans/check_topup.py
```diff
@@ -0,0 +1,72 @@
...
+        Total exposure time less than time to next topup. Proceeding with collection.
+        """
+    )
+    return False
+
+
+def check_topup_and_wait_if_necessary(
+    synchrotron: Synchrotron,
+    params: DetectorParams,
+    ops_time: float = 30.0,  # Account for xray centering, rotation speed, etc
```
Tom-Willemsen, line 44: I'm not sure it makes sense to have a default at *this* level - my impression is that this plan should be generic across any collection type, and so it doesn't seem right to have a default here that *assumes* it's a particular type of collection we're running?

src/hyperion/device_setup_plans/check_topup.py
```diff
@@ -0,0 +1,72 @@
...
+        params (DetectorParams): The detector parameters, used to determine length\
+            of scan.
+        ops_time (float, optional): Additional time to account for various operations,\
+            eg. x-ray centering. In seconds. Defaults to 30.0.
+    """
+    if _in_decay_mode(
+        synchrotron.top_up.start_countdown.get()
+    ) or not _gating_permitted(synchrotron.machine_status.synchrotron_mode.get()):
+        time_to_wait = 0
+        # yield from bps.null()
```
Tom-Willemsen, line 60: Should be able to remove this, I think.

src/hyperion/device_setup_plans/check_topup.py
```diff
@@ -0,0 +1,72 @@
...
+            params.exposure_time * params.full_number_of_images + ops_time
+        )
+        time_to_topup = synchrotron.top_up.start_countdown.get()
+        time_to_wait = (
+            synchrotron.top_up.end_countdown.get()
+            if _delay_to_avoid_topup(tot_exposure_time, time_to_topup)
+            else 0.0
+        )
+
+    yield from bps.sleep(time_to_wait)
```
Tom-Willemsen, line 72: I guess it's probably fine to emit a `("sleep", 0)` message, even if it's a little bit weird. If we *wanted* to avoid it we could do something like `yield from bps.sleep(time_to_wait) if time_to_wait > 0 else bps.null()`...

src/hyperion/device_setup_plans/check_topup.py
```diff
@@ -0,0 +1,72 @@
...
+    Args:
+        synchrotron (Synchrotron): Synchrotron device.
+        params (DetectorParams): The detector parameters, used to determine length\
+            of scan.
+        ops_time (float, optional): Additional time to account for various operations,\
+            eg. x-ray centering. In seconds. Defaults to 30.0.
+    """
+    if _in_decay_mode(
+        synchrotron.top_up.start_countdown.get()
+    ) or not _gating_permitted(synchrotron.machine_status.synchrotron_mode.get()):
```
Tom-Willemsen, line 58: Aren't we meant to use `bps.rd()` rather than `.get()` directly to read ophyd signals? (comment applies anywhere in this PR we have `.get()`)?

src/hyperion/device_setup_plans/check_topup.py
```diff
@@ -0,0 +1,72 @@
...
+        return True
+    return False
+
+
+def _delay_to_avoid_topup(total_exposure_time, time_to_topup):
+    if total_exposure_time > time_to_topup:
+        LOGGER.info(
+            """
+            Total exposure time + time needed for x ray centering exceeds time to
+            next top up. Collection delayed until top up done.
```
Tom-Willemsen, line 29: Parameter name is inconsistent with logging statement - parameter implies it's only counting exposure time, but logging statement says it's exposure + ops time.

src/hyperion/device_setup_plans/check_topup.py
```diff
@@ -0,0 +1,72 @@
...
+from dodal.devices.detector import DetectorParams
+from dodal.devices.synchrotron import Synchrotron, SynchrotronMode
+
+from hyperion.log import LOGGER
+
+ALLOWED_MODES = [SynchrotronMode.USER, SynchrotronMode.SPECIAL]
+
+
+def _in_decay_mode(time_to_topup):
+    if time_to_topup == -1:
```
Tom-Willemsen, line 11: This feels like a bit of a magic number. From context I'm guessing it's what the synchrotron PV reports when the synchrotron is in decay mode, maybe naming this constant would make it clearer.

src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -179,7 +180,12 @@ def run_gridscan(
 
     fgs_motors = fgs_composite.fast_grid_scan
 
-    # TODO: Check topup gate
+    # Check topup gate
+    yield from check_topup_and_wait_if_necessary(
+        fgs_composite.synchrotron,
+        parameters.hyperion_params.detector_params,
+    )
+
```
Tom-Willemsen, line 188: Query: could we actually delay this until a bit later in the plan?

Intuitively I might expect that we can arm the detector even if we're waiting for a topup, so that as soon as the topup finishes we're ready to start collecting.

Or if not could you add a brief comment explaining why in the code?

src/hyperion/device_setup_plans/check_topup.py
Tom-Willemsen, line 72: Slight concern about getting a timer for an external process, and assuming that the external process will have finished if we sleep for the timer duration. This feels like a race condition waiting to happen...

Should we consider a polling loop *after* the main sleep to ensure the topup has actually completed?

src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
DominicOram: Yes, the latest we can do it is just before `yield from bps.kickoff(fgs_motors)`

src/hyperion/experiment_plans/rotation_scan_plan.py
noemifrisina: Made an issue out of these comments. See #932

Tom-Willemsen (approved):
Looks fine in principle. Probably should be tested on a beamline at some point but happy for you to merge before then.

Tom-Willemsen (approved):

src/hyperion/device_setup_plans/check_topup.py
```diff
@@ -0,0 +1,80 @@
...
+def wait_for_topup_complete(synchrotron):
+    start = yield from bps.rd(synchrotron.top_up.start_countdown)
+    while start == 0:
+        yield from bps.sleep(0.1)
+        start = yield from bps.rd(synchrotron.top_up.start_countdown)
+
+
+def check_topup_and_wait_if_necessary(
+    synchrotron: Synchrotron,
+    params: DetectorParams,
```
DominicOram, line 52: Should: It's possible that the detector parameters doesn't actually have the correct number of frames in when we're free running the detetcor. I think it would be better for this function just to take the expected time the collection will take, the individual collections can calculate it

src/hyperion/device_setup_plans/check_topup.py
```diff
@@ -0,0 +1,80 @@
...
+    tot_run_time = params.exposure_time * params.full_number_of_images + ops_time
+    end_topup = yield from bps.rd(synchrotron.top_up.end_countdown)
+    time_to_wait = (
+        end_topup if _delay_to_avoid_topup(tot_run_time, time_to_topup) else 0.0
+    )
+
+    yield from bps.sleep(time_to_wait)
+
+    check_start = yield from bps.rd(synchrotron.top_up.start_countdown)
+    if check_start == 0:
```
DominicOram, line 79: Should: It's not immediately obvious that 0 means it's being topped up, code you pull it into a constant to make it more obvious?

DominicOram (changes requested):
Looks good but I'm worried about how it will behave when the detector is being free run

DominicOram (changes requested):
Thanks, couple of minor comments, otherwise good.

Issue #924, opened 2023-10-09
Failed to move to find tip


DominicOram:
For collection https://ispyb.diamond.ac.uk/dc/visit/mx33658-16/dcg/10470342 Hyperion should have moved `x` around to try and find the tip, it didn't. From the logs it looks like Hyperion thought it found a tip. We should:

* Add more logging
* Watch out for this happening again, if it does look into averaging tip detection amongst more frames

DominicOram:
We have seen this a couple of additional times when running overnight

DominicOram:
https://ispyb.diamond.ac.uk/samples/sid/5003238 initially found a tip at (608, 360) but then couldn;t find the loop. We could also fix this by tweaking the edge detection settings. Suggest we find some time in shutdown to leave the edge detection on with no pin then see if we can catch when it gives a false positive

DominicOram:
Good solution will be wait until the pin tip is "stable"

dperl-dls:
@dperl-dls waiting on ophyd2 oav stuff

DominicOram:
Hotfix in https://github.com/DiamondLightSource/dodal/pull/242 works fine for now, let's call this done

PR #926, opened 2023-10-09
Add logging tip when pin is found and tidy up move to find pin

DominicOram:
See #924 (not yet fixed but should help to debug)

### To test:
1. Confirm that the tip location will be logged whenever one is found
2. Confirm that the changes done (as suggested in https://github.com/DiamondLightSource/hyperion/pull/907#pullrequestreview-1653087229) are more readable and pass the tests still


noemifrisina (approved):
Looks good to me, logging message shows up and tests pass.

PR #931, opened 2023-10-10
Write end_time_estimated in nexus file

### LINKED ISSUE #887: Write predicted end time in nexus file
We have seen more zocalo issues when we're putting the end time into the nexus file whilst zocalo is validating it. Of the form:
```
{'beamline': 'i03',
 'dc_comments': 'Hyperion: Xray centring - Diffraction grid scan of 30 by 13 '
                'images in 20.0 um by 20.0 um steps. Top left (px): [484,243], '
                'bottom right (px): [2206,989].',
 'file': '/dls/i03/data/2023/cm33866-4/xraycentring/TestProteinaseK/protk_4/protk_4_25_master.h5',
 'ispyb_expected_images': '390',
 'ispyb_wavelength': '0.976284',
 'reason': 'Unhandled KeyError exception reading '
           '/dls/i03/data/2023/cm33866-4/xraycentring/TestProteinaseK/protk_4/protk_4_25_master.h5'}
```
We can get round having to edit the nexus file after the collection if we instead put in the predicted time for the experiment. We can use the same time we use for the topup gate for this

noemifrisina:
Fixes #887 

Writes `end_time_estimated` field in the nexus file at the same time as `start_time` instead of `end_time`, removing the need to reopen the file at the end of the collection.
This is also more in line with the NXmx standard.

### To test:
1. Check tests relating to the nexus write/handler/callback still make sense and pass

### DIFF
--- a/src/hyperion/external_interaction/nexus/nexus_utils.py
+++ b/src/hyperion/external_interaction/nexus/nexus_utils.py
@@ -64,11 +64,10 @@ def create_goniometer_axes(
 
 
 def get_start_and_predicted_end_time(time_expected: float) -> tuple[str, str]:
+    time_format = r"%Y-%m-%dT%H:%M:%SZ"
     start = datetime.utcfromtimestamp(time.time())
     end_est = start + timedelta(seconds=time_expected)
-    return start.strftime(r"%Y-%m-%dT%H:%M:%SZ"), end_est.strftime(
-        r"%Y-%m-%dT%H:%M:%SZ"
-    )
+    return start.strftime(time_format), end_est.strftime(time_format)
 
 
 def create_detector_parameters(detector_params: DetectorParams) -> Detector:
--- a/src/hyperion/external_interaction/nexus/nexus_utils.py
+++ b/src/hyperion/external_interaction/nexus/nexus_utils.py
@@ -63,13 +63,12 @@ def create_goniometer_axes(
     return Goniometer(gonio_axes, scan_points)
 
 
-def get_current_time():
-    return datetime.utcfromtimestamp(time.time()).strftime(r"%Y-%m-%dT%H:%M:%SZ")
-
-
-def get_predicted_end_time(start: str, delta: float):
-    end = datetime.strptime(start, r"%Y-%m-%dT%H:%M:%SZ") + timedelta(seconds=delta)
-    return end.strftime(r"%Y-%m-%dT%H:%M:%SZ")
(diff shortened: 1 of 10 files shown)


src/hyperion/external_interaction/nexus/write_nexus.py
```diff
@@ -79,6 +77,9 @@ def create_nexus_file(self):
         initialised.
         """
         start_time = get_current_time()
+        est_end_time = get_predicted_end_time(
+            start_time, self.detector.exp_time * self.full_num_of_images
+        )
```
DominicOram, line 82: Should: It's a bit of a faff to get current time, convert to string then convert back to get the end time. I think we could just do it in one function like `start_time, est_end_time = get_start_and_predicted_end_time(time_expected)`, which will be simpler

DominicOram (changes requested):
This is great, thank you. Nice to remove so much code! One small comment in code.

src/hyperion/external_interaction/nexus/write_nexus.py
noemifrisina: Good point, done

src/hyperion/external_interaction/nexus/nexus_utils.py
```diff
@@ -63,8 +63,12 @@ def create_goniometer_axes(
...
 
 
-def get_current_time():
-    return datetime.utcfromtimestamp(time.time()).strftime(r"%Y-%m-%dT%H:%M:%SZ")
+def get_start_and_predicted_end_time(time_expected: float) -> tuple[str, str]:
+    start = datetime.utcfromtimestamp(time.time())
+    end_est = start + timedelta(seconds=time_expected)
+    return start.strftime(r"%Y-%m-%dT%H:%M:%SZ"), end_est.strftime(
+        r"%Y-%m-%dT%H:%M:%SZ"
+    )
```
DominicOram, line 71: Nit: Slightly cleaner I think, take it or leave it.
```suggestion
def get_start_and_predicted_end_time(time_expected: float) -> tuple[str, str]:
    time_format = r"%Y-%m-%dT%H:%M:%SZ"
    start = datetime.utcfromtimestamp(time.time())
    end_est = start + timedelta(seconds=time_expected)
    return start.strftime(time_format), end_est.strftime(time_format)
```

DominicOram (approved):
Great, thank you! Minor suggestion in code, take it or leave it.

PR #936, opened 2023-10-12
Recenter smargon with stub offsets when xray center found

### LINKED ISSUE #794: Stub offset motion
TBD

DominicOram:
Fixes #794

Also removed duplicate test

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/204

### To test:
1. Confirm tests pass


src/hyperion/experiment_plans/tests/test_flyscan_xray_centre_plan.py
```diff
@@ -321,11 +309,31 @@ def test_logging_within_plan(
...
+    run_gridscan: MagicMock,
+    fake_fgs_composite: FlyScanXRayCentreComposite,
+    test_fgs_params: GridscanInternalParameters,
+    RE: RunEngine,
+):
+    move_xyz.side_effect = Exception()
+    mock_subscriptions = MagicMock()
+    mock_subscriptions.zocalo_handler.wait_for_results.return_value = ((0, 0, 0), None)
+
+    with pytest.raises(Exception):
```
Tom-Willemsen, line 328: I think it would be better to make the side effect of `move_xyz` some _specific_ exception type, and then only assert that one exception is thrown. This will catch (and pass) on all sorts of things currently - including programming errors like misnamed attributes and the like...

DominicOram:
Tested this on the beamline, there seems to be a race condition between the move and setting the stubs. The apply stubs button on the synoptic gets greyed out sometimes, need to check on the edm when this happens and maybe use similar logic in Hyperion

DominicOram:
The greying out happens when the `BL03I-MO-SGON-01:CENTER_CS.DISP` is set to 1, which is done when x, y and z stop moving. This should be done when we're finished on `move_xyz` but isn't. We should check on the beamline:
* The state of `BL03I-MO-SGON-01:CENTER_CS.DISP` when the race condition occurs
* If it is 1 check if ophyd respects that it's disabled

DominicOram:
Managed to reproduce using:
```python
from dodal.beamlines.i03 import smargon
from dodal.devices.smargon import StubPosition
from hyperion.device_setup_plans.manipulate_sample import move_x_y_z
from bluesky.run_engine import RunEngine
from random import uniform
my_smargon = smargon()

def my_plan():
    x = uniform(-1, 1)
    y = uniform(-1, 1)
    z = uniform(-1, 1)

    yield from bps.mv(my_smargon.omega, 90, wait=True)

    yield from move_x_y_z(my_smargon, x, y, z, wait=True)

    yield from bps.mv(
        my_smargon.stub_offsets, StubPosition.CURRENT_AS_CENTER
    )
    print("Finished applying")

    yield from bps.sleep(10)
    yield from bps.mv(
        my_smargon.stub_offsets, StubPosition.RESEET_TO_ROBOT_LOAD
    )
    yield from move_x_y_z(my_smargon, 0, 0, 0, wait = True)
    yield from bps.mv(my_smargon.omega, 0)


RE = RunEngine()
RE(my_plan())
```
On the beamline against https://github.com/DiamondLightSource/dodal/pull/204/commits/ff3bdc5bdc8fd6b310c65924a667f432754472c2. Still need to:
* Tidy this up
* Add tests
* Set to center isn't actually finished until positions are 0, 0, 0
* Do this in v2 instead, it will be simpler

DominicOram:
Spent some time trying to do this in ophyd-async but came up on a number of problems so closing this PR in favour of a simpler solution

PR #937, opened 2023-10-17
891 read flux tranmission put into ispyb during collection

### LINKED ISSUE #891: Read flux and transmission (and put into ispyb) during collection
Currently we create the ispyb entry quite early and put the flux and transmission into ispyb at this point. This limits us in having to set the transmission early, which we would rather not do so as to maintain the xbpm feedback for as long as possible. However, we do want to make the ispyb entry early so that it shows up in synchweb and we can update it if we encounter issues.

We should split up making the entry and filling in the flux so that we:
* Make the entry early
* Read the flux and transmission whilst we're collecting and update the database with these

We can then move setting the transmission to just before collection

subinsaji:
Fixes #891

1. Split `reading hardware` into two event streams; `read_hardware_for_ispyb_pre_collection` and `read_hardware_for_ispyb_during_collection `and subsequently modifying and adding new code in order to fix various tests.  

2. Merge conflicts fixed and tests pass accordingly after updating the venv dependencies.

### To test:
1. Run tests


src/hyperion/parameters/constants.py
```diff
@@ -3,6 +3,7 @@
 SIM_BEAMLINE = "BL03S"
 SIM_INSERTION_PREFIX = "SR03S"
 ISPYB_PLAN_NAME = "ispyb_readings"
```
dperl-dls, line 5: Nit: I would like it if these names reflected the plans more closely - e.g. `ISPYB_PREPLAN_READ` and 'ISPYB_MIDPLAN_READ' or something like that, to show that the two are related

dperl-dls (approved):
This is great! I just have one minor comment I think would make it more readable - feel free to do it or not - otherwise looks good to me.

PR #938, opened 2023-10-17
Ophyd pin-tip centring

Tom-Willemsen:
Requires https://github.com/DiamondLightSource/dodal/pull/289

Provide support for running both Ophyd + AD MXSC pin tip centring workflows in hyperion.

To test:
* Confirm using the ophyd device is externally selectable
* Confirm the issues in https://github.com/DiamondLightSource/hyperion/pull/938#issuecomment-1777284170 are addressed:
    * https://github.com/DiamondLightSource/dodal/issues/290
    * https://github.com/DiamondLightSource/hyperion/issues/1068
    * https://github.com/DiamondLightSource/hyperion/issues/1069
* Confirm tests pass


DominicOram:
Tested on beamline today. Broadly works but need to:
* Tidy up
* Set the PVA plugin to look at PROC when at ZOOM 1, we won't need to do this if we get https://jira.diamond.ac.uk/browse/I03-912 done
* Allow turning this on to be user settable
* New issue - write a snapshot to file if no pin tip was found (must the same frame that was used for detection)
* Potentially new issue - use this device for the grid detection too

Tom-Willemsen:
Handed over to @dperl-dls .

DominicOram:
I'll pick this up now so that @dperl-dls can focus on rotation scan stuff

src/hyperion/device_setup_plans/setup_oav.py
```diff
@@ -163,3 +190,21 @@ def wait_for_tip_to_be_found(mxsc: MXSC) -> Generator[Msg, None, Tuple[int, int]
...
+
+
+def wait_for_tip_to_be_found_ophyd(
+    ophyd_pin_tip_detection: PinTipDetection,
+) -> Generator[Msg, None, Tuple[int, int]]:
+    found_tip = yield from bps.rd(ophyd_pin_tip_detection)
+
+    LOGGER.info("Pin tip not found, waiting a second and trying again")
+
+    if found_tip == ophyd_pin_tip_detection.INVALID_POSITION:
```
dperl-dls, line 202: Do I understand correctly that this will be adressed in https://github.com/DiamondLightSource/hyperion/issues/1069?

dperl-dls (approved):
Looks good, thanks for tidying this up

src/hyperion/device_setup_plans/setup_oav.py
DominicOram: Yep

PR #941, opened 2023-10-23
Add units to dwell time

### LINKED ISSUE #940: Add units to dwell time
In reviewing https://github.com/DiamondLightSource/hyperion/pull/925/files it wasn't clear if dwell time was in seconds, we should give the variable a unit

DominicOram:
Fixes #940

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/213

### To test:
1. Confirm tests pass

### DIFF
--- a/live_test_rotation_params.json
+++ b/live_test_rotation_params.json
@@ -1,5 +1,5 @@
 {
-    "params_version": "3.0.0",
+    "params_version": "4.0.0",
     "hyperion_params": {
         "beamline": "BL03I",
         "insertion_prefix": "SR03I",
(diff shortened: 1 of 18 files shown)


noemifrisina (approved):
Looks okay, thanks! 

Question: is there also a change in GDA to go along with this?

DominicOram:
Good spot, yes we will need a GDA change too

DominicOram:
Actually, that code in GDA will be removed soon anyway (https://github.com/DiamondLightSource/hyperion/issues/823) so we should just do that instead

PR #942, opened 2023-10-23
Initial VMXm FGS plan

Tom-Willemsen:
First pass at a FGS plan for VMXm - to test and refine on the beamline.

Based largely on existing FGS plan with references to the following devices removed:
- `ap_sg`
- `flux`
- `s4_slits`
- `undulator`
- `xbpm`

`Attenuator` replaced with `VmxmAttenuator`

Nothing has been done about either shutter - understand that the fast shutter is being controlled by motion program, and that slow shutter can be controlled by GDA prior to calling into hyperion? This assumption might be wrong, check it on the beamline.

Needs https://github.com/DiamondLightSource/dodal/pull/211


src/hyperion/experiment_plans/vmxm_flyscan_xray_centre_plan.py
```diff
@@ -0,0 +1,305 @@
...
+        GridscanInternalParameters,
+    )
+
+    parameters = GridscanInternalParameters(**external_parameters.from_file())
+    subscriptions = XrayCentreCallbackCollection.from_params(parameters)
+
+    context = setup_context(wait_for_connection=True)
+    composite = create_devices(context)
+
+    RE(flyscan_xray_centre(composite, parameters))
```
Tom-Willemsen, line 305: Fix or remove

src/hyperion/external_interaction/callbacks/ispyb_callback_base.py
```diff
@@ -39,38 +39,39 @@ def descriptor(self, doc: dict):
         self.descriptors[doc["uid"]] = doc
 
     def start(self, doc: dict):
+        LOGGER.info(f"ISPyB handler received start document {doc}.")
```
Tom-Willemsen, line 42: Remove

src/hyperion/external_interaction/callbacks/ispyb_callback_base.py
```diff
@@ -39,38 +39,39 @@ def descriptor(self, doc: dict):
...
         if self.uid_to_finalize_on is None:
             self.uid_to_finalize_on = doc.get("uid")
 
     def event(self, doc: dict):
         """Subclasses should extend this to add a call to set_dcig_tag from
         hyperion.log"""
 
-        LOGGER.debug("ISPyB handler received event document.")
+        LOGGER.info(f"ISPyB handler received event document {doc}.")
```
Tom-Willemsen, line 50: Remove

src/hyperion/external_interaction/callbacks/ispyb_callback_base.py
```diff
@@ -39,38 +39,39 @@ def descriptor(self, doc: dict):
...
+            # ]
+            # self.params.hyperion_params.ispyb_params.slit_gap_size_y = doc["data"][
+            #     "s4_slit_gaps_ygap"
+            # ]
+            # self.params.hyperion_params.ispyb_params.transmission_fraction = doc[
+            #     "data"
+            # ]["attenuator_actual_transmission"]
+            # self.params.hyperion_params.ispyb_params.flux = doc["data"][
+            #     "flux_flux_reading"
+            # ]
```
Tom-Willemsen, line 74: Make generic enough to accept missing data

src/hyperion/external_interaction/callbacks/xray_centre/callback_collection.py
```diff
@@ -28,16 +28,16 @@ class XrayCentreCallbackCollection(AbstractPlanCallbackCollection):
...
     def from_params(cls, parameters: InternalParameters):
         nexus_handler = GridscanNexusFileCallback()
         ispyb_handler = GridscanISPyBCallback(parameters)
-        zocalo_handler = XrayCentreZocaloCallback(parameters, ispyb_handler)
+        # zocalo_handler = XrayCentreZocaloCallback(parameters, ispyb_handler)
         callback_collection = cls(
             nexus_handler=nexus_handler,
             ispyb_handler=ispyb_handler,
-            zocalo_handler=zocalo_handler,
+            # zocalo_handler=zocalo_handler,
```
Tom-Willemsen, line 41: Probably make a `VmxmFlyscanCallbackCollection` that doesn't contain zocalo.

src/hyperion/external_interaction/callbacks/xray_centre/nexus_callback.py
```diff
@@ -55,12 +55,13 @@ def descriptor(self, doc):
...
-                vds_start_index=nexus_data_1["data_shape"][0],
-            )
+            # self.nexus_writer_2 = NexusWriter(
+            #     self.parameters,
+            #     **nexus_data_2,
+            #     vds_start_index=nexus_data_1["data_shape"][0],
+            # )
             self.nexus_writer_1.create_nexus_file()
-            self.nexus_writer_2.create_nexus_file()
+            # self.nexus_writer_2.create_nexus_file()
```
Tom-Willemsen, line 67: Find a way to dynamically configure this for a single nexus writer. Or rename this one to `dual_nexus_callback` and make a `single_nexus_callback`.

src/hyperion/external_interaction/nexus/nexus_utils.py
```diff
@@ -10,7 +10,7 @@
 from hyperion.external_interaction.ispyb.ispyb_dataclass import IspybParams
 
 
-def create_goniometer_axes(
+def create_vmxm_goniometer_axes(
```
Tom-Willemsen, line 13: Fix `i03`-specific assumptions.

src/hyperion/unit_tests/test_lookup_table.txt
```diff
@@ -1,5 +1,5 @@
 # Beam converter lookup table for testing
 
 Units det_dist beam_x beam_y
-100.0 150.0 160.0
-200.0 151.0 165.0
+300 123.35 130.64
+593 123.35 130.64
```
Tom-Willemsen, line 5: Put back

test_parameters.json
```diff
@@ -3,13 +3,13 @@
     "hyperion_params": {
         "beamline": "BL03S",
         "insertion_prefix": "SR03S",
-        "detector": "EIGER2_X_16M",
+        "detector": "EIGER2_X_9M",
```
Tom-Willemsen, line 6: Put back whole file.

src/hyperion/external_interaction/callbacks/xray_centre/callback_collection.py
```diff
@@ -24,15 +29,17 @@
...
+    Note: currently specific to i03 as it creates i03's goniometer axes."""
 
     nexus_handler: GridscanNexusFileCallback
     ispyb_handler: GridscanISPyBCallback
     zocalo_handler: XrayCentreZocaloCallback
 
     @classmethod
     def from_params(cls, parameters: InternalParameters):
-        nexus_handler = GridscanNexusFileCallback()
+        nexus_handler = GridscanNexusFileCallback(create_i03_goniometer_axes)
```
Tom-Willemsen, line 42: Still feels like we should be passing this from "higher up" the call chain. But probably fine for now. Possibly create an issue to think about where this comes from.

src/hyperion/external_interaction/callbacks/xray_centre/nexus_callback.py
```diff
@@ -55,12 +59,56 @@ def descriptor(self, doc):
...
+                create_goniometer_func=self.create_goniometer_func,
                 **nexus_data_2,
                 vds_start_index=nexus_data_1["data_shape"][0],
             )
             self.nexus_writer_1.create_nexus_file()
             self.nexus_writer_2.create_nexus_file()
+
+
+class Gridscan2DNexusFileCallback(CallbackBase):
+    """Similar to above, but for a 2D gridscan"""
```
Tom-Willemsen, line 80: I think the duplication with the class above shouldn't block merging in the short term, but let's write an issue to improve this in the longer term.

src/hyperion/external_interaction/nexus/nexus_utils.py
```diff
@@ -65,6 +65,57 @@ def create_goniometer_axes(
...
 
 
+def create_vmxm_goniometer_axes(
+    omega_start: float,
+    scan_points: dict | None,
+    x_y_z_increments: tuple[float, float, float] = (0.0, 0.0, 0.0),
+    chi: float = 0,
+    phi: float = 0,
+) -> Goniometer:
+    """Returns a Nexgen 'Goniometer' object with the dependency chain of I03's Smargon
```
Tom-Willemsen, line 75:
```suggestion
    """Returns a Nexgen 'Goniometer' object with the dependency chain of VMXm's axes
```

Comment is obviously wrong!

src/hyperion/external_interaction/callbacks/ispyb_callback_base.py
```diff
@@ -42,44 +42,45 @@ def descriptor(self, doc: dict):
...
-            ]
+            ].get("attenuator_actual_transmission", 0.0)
+            self.params.hyperion_params.ispyb_params.flux = doc["data"].get(
+                "flux_flux_reading", 0.0
+            )
 
-            LOGGER.info("Creating ispyb entry.")
-            self.ispyb_ids = self.ispyb.begin_deposition()
-            LOGGER.info(f"Recieved ISPYB IDs: {self.ispyb_ids}")
+        LOGGER.info("Creating ispyb entry.")
```
olliesilvester, line 81: How come this part is no longer within the if statement?

Tom-Willemsen: If I recally correctly, VMXm only uses the `READ_HARDWARE` plan so they need to create the ispyb entry on that event - otherwise they won't get an entry at all. For `i03`, we need to check whether this is right. 

It feels like a bit of a hack anyway to have this ispyb stuff listening on a specific type of message - but I think David or Dom have been doing some work to pull the ispyb interactions into an ophyd device? If that's the case then this may go away.

src/hyperion/external_interaction/callbacks/xray_centre/callback_collection.py
```diff
@@ -41,3 +48,23 @@ def from_params(cls, parameters: InternalParameters):
             zocalo_handler=zocalo_handler,
         )
         return callback_collection
+
+
+@dataclass(frozen=True, order=True)
+class VmxmFastGridScanCallbackCollection(AbstractPlanCallbackCollection):
```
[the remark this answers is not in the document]
Tom-Willemsen: See [my comment of the equivalent bit of code for `i03`](https://github.com/DiamondLightSource/hyperion/pull/942/files/35ee736c34742df5a55882a3ea05c57110cdeaac#r1404226961):

> Still feels like we should be passing this from "higher up" the call chain. But probably fine for now. Possibly create an issue to think about where this comes from.

So yes I agree with you :)

olliesilvester:
Things we need to change before we merge. You wrote all this stuff in the comments anyway, this is mainly for me

- In the deploy script, make sure the permissions go to the correct beamline (eg vmxm don't get permissions on i03 and vice versa)
- Contact ispyb to get separate credentials for Vmxm and i04 (https://github.com/DiamondLightSource/hyperion/issues/954), then we can change `ISPYB_CONFIG_PATH` accordingly
- Get a way for Hyperion to ignore certain plans on certain beamlines / not require all plans to sucessfully connect to all devices. (https://github.com/DiamondLightSource/hyperion/pull/942/files#r1404216293) I prefer the first of these two solutions.
- Make sure the eiger_params in `get_start_and_predicted_end_time` are generic enough for other beamlines
- There are also a bunch of merge conflicts which we will eventually need to sort out. It probably makes sense to fix the above things before doing this though.

Other things which aren't as urgent but we should still do at some point:

- Find the right set of abstractions between `vmxm_flyscan_xray_centre_plan` and the regular `flyscan_xray_centre_plan`. Also rename this for VMXm since it isn't actually doing any xray centring - instead the grid scan is the actual data collection.
- The ispyb callback class should use a beamline-specific config, since otherwise we have to put default values into ispyb entries for other beamlines - values for devices which don't exist for that beamline.
- Make it clear throughout hyperion as to which parts are generic and which parts are beamline specific
- Think of a nicer way to create the nexus file handler with beamline specific goniometer axes. Right now the callback collection classes are beamline specific, but only because the goniometer axes are beamline specific
- Create abstractions for the `GridscanNexusFilecallback`, since there's currently a lot of duplication

src/hyperion/external_interaction/callbacks/xray_centre/callback_collection.py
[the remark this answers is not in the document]
Tom-Willemsen: Don't put _too_ much effort into this though, because I think David (or Dom?) is looking at pulling all or most of these callbacks into ophyd devices which would then make this obsolete anyway.

PR #943, opened 2023-10-23
(#891) Renamed plans to reflect more closely what they do

### LINKED ISSUE #891: Read flux and transmission (and put into ispyb) during collection

5u6in:
Fixes #891

Changed names of `ISPYB_PLAN_NAME `and  `ISPYB_UPDATING_COLLECTION `to `ISPYB_HARDWARE_READ_PLAN` and I`SPYB_TRANSMISSION_FLUX_READ_PLAN` respectively following suggesting from https://github.com/DiamondLightSource/hyperion/pull/937#discussion_r1368406664

### To test:
1. Confirm Tests

### DIFF
--- a/src/hyperion/device_setup_plans/read_hardware_for_setup.py
+++ b/src/hyperion/device_setup_plans/read_hardware_for_setup.py
@@ -5,8 +5,8 @@ from dodal.beamlines.i03 import Attenuator, Flux, S4SlitGaps, Synchrotron, Undul
 
 import hyperion.log
 from hyperion.parameters.constants import (
-    ISPYB_PLAN_NAME,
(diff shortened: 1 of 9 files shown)


dperl-dls (approved):
Great, this is much more readable, thank you

PR #953, opened 2023-11-02
Fix flaky test_main_system tests

### LINKED ISSUE #939: Flakey unit test from test_main_systems.py in CI/CD
Prompted by test that are run in GitHub's CI/CD for pull request #937 :

- Tests pass locally using `python -m pytest "not s03"`
- However in GitHub's CI it took multiple tries for it to pass ([attempt 1](https://github.com/DiamondLightSource/hyperion/actions/runs/6562710062/attempts/1), [attempt 2](https://github.com/DiamondLightSource/hyperion/actions/runs/6562710062/attempts/2), [attempt 3](https://github.com/DiamondLightSource/hyperion/actions/runs/6562710062/attempts/3)). Passes in latest [attempt 4](https://github.com/DiamondLightSource/hyperion/actions/runs/6562710062)
- In other times the tests took less number of runs for it go between failing and passing
- Test fails in `test_given_started_when_RE_stops_on_its_own_with_error_then_error_reported` in **test_main_system.py**

If possible, document any common patterns/anti-patterns that are causing these tests to be unstable so that we can avoid them in future.

dperl-dls:
Fixes #939 

On discussion with @DominicOram we decided just to remove the offending test as diagnosing the issue will take too much time, and we won't be using the blueskyrunner for much longer, and we know it works well on the beamline.

### To test:
1. Run `pytest --random-order -m "not s03" a few times
2. Confirm tests in `test_main_system` don't fail

### DIFF
--- a/src/hyperion/__main__.py
+++ b/src/hyperion/__main__.py
@@ -221,6 +221,9 @@ class StopOrStatus(Resource):
         action = kwargs.get("action")
         status_and_message = StatusAndMessage(Status.FAILED, f"{action} not understood")
         if action == Actions.STATUS.value:
+            hyperion.log.LOGGER.debug(
+                f"Runner recieved status request - state of the runner object is: {self.runner.__dict__} - state of the RE is: {self.runner.RE.__dict__}"
+            )
             status_and_message = self.runner.current_status
         return asdict(status_and_message)
 
--- a/src/hyperion/system_tests/test_main_system.py
+++ b/src/hyperion/system_tests/test_main_system.py
@@ -26,6 +26,7 @@ from hyperion.__main__ import (
 )
 from hyperion.exceptions import WarningException
 from hyperion.experiment_plans.experiment_registry import PLAN_REGISTRY
+from hyperion.log import LOGGER
 from hyperion.parameters import external_parameters
 from hyperion.parameters.plan_specific.gridscan_internal_params import (
     GridscanInternalParameters,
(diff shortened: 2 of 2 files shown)


src/hyperion/system_tests/test_main_system.py
```diff
@@ -117,8 +121,9 @@ def mock_dict_values(d: dict):
 
 
 @pytest.fixture
-def test_env():
+def test_env(request):
```
DominicOram, line 124: :+1: I did not know you could do this, it's cool

src/hyperion/system_tests/test_main_system.py
```diff
@@ -117,8 +121,9 @@ def mock_dict_values(d: dict):
...
     mock_run_engine = MockRunEngine()
```
DominicOram, line 125: Nit: Better if the `MockRunEngine` requires a test name on init, then we don't accidentally end up with logs saying the test was called `test`

src/hyperion/__main__.py
```diff
@@ -221,6 +221,9 @@ def get(self, **kwargs):
         action = kwargs.get("action")
         status_and_message = StatusAndMessage(Status.FAILED, f"{action} not understood")
         if action == Actions.STATUS.value:
+            hyperion.log.LOGGER.debug(
+                f"Runner recieved status request - state of the runner object is: {self.runner.__dict__} - state of the RE is: {self.runner.RE.__dict__}"
+            )
```
DominicOram, line 226: Could: Not sure how useful this is in production code. I would remove it but if you think it's useful that's fine

DominicOram (approved):
Great, thank you. Minor comments in code, take them or leave them

src/hyperion/__main__.py
dperl-dls: I thought it might actually be handy some time to be able to make it dump the RE status

PR #962, opened 2023-11-08
951 activate callbacks based on plan

### LINKED ISSUE #951: Trigger callbacks based on the plan type
To allow subscribing to callbacks at the runengine level, we need to trigger the callbacks only if the plan it corresponds to is running, rather than on any start document, so we need to include some metadata in the start document to facilitate this and activate/deactivate persistent callback instances based on that metadata

dperl-dls:
Fixes #951 

Adds a class in between our callbacks and `CallbackBase` which adds functions gated by an activity flag which is set and unset on receiving appropriate documents from the `RunEngine`

### To test:
1. Run tests


src/hyperion/external_interaction/callbacks/plan_reactive_callback.py
```diff
@@ -0,0 +1,51 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Callable
+
+from bluesky.callbacks import CallbackBase
+from event_model.documents.event import Event
+
+if TYPE_CHECKING:
+    from event_model.documents.event import Event
```
olliesilvester, line 9: This is imported twice

src/hyperion/external_interaction/callbacks/plan_reactive_callback.py
```diff
@@ -0,0 +1,51 @@
...
+        return self.activity_gated_event(doc) if self.active else doc
+
+    def stop(self, doc: RunStop) -> RunStop | None:
+        do_stop = self.active
+        if doc.get("run_start") == self.activity_uid:
+            self.active = False
+            self.activity_uid = 0
+        return self.activity_gated_stop(doc) if do_stop else doc
+
+    def activity_gated_start(self, doc: RunStart):
```
olliesilvester, line 41: Do we want to enforce these classes being defined by any subclasses of `PlanReactiveCallback`?

dperl-dls: no, a lot of callbacks don't use all the methods

src/hyperion/external_interaction/callbacks/plan_reactive_callback.py
```diff
@@ -0,0 +1,51 @@
...
+from event_model.documents.event import Event
+
+if TYPE_CHECKING:
+    from event_model.documents.event import Event
+    from event_model.documents.event_descriptor import EventDescriptor
+    from event_model.documents.run_start import RunStart
+    from event_model.documents.run_stop import RunStop
+
+
+class PlanReactiveCallback(CallbackBase):
```
olliesilvester, line 15: Might be useful to have a brief comment explaining that this should be used when we only want callbacks while a plan is running

olliesilvester (approved):
This looks neat! There's a double import I've mentioned on the comments, but it looks like everything here should work okay.

src/hyperion/external_interaction/callbacks/plan_reactive_callback.py
dperl-dls: good catch, thanks

PR #971, opened 2023-11-09
(955) Implement parsing of simple lists in GDA beamline parameter files

### LINKED ISSUE #955: Reading I04's GDA beamline_parameters currently errors as it tries to set a list as a float
When running Hyperion on I04, `/hyperion/parameters/beamline_parameters.py` reads GDA's 
`/i04-config/scripts/beamlineParameters` and reads in some of the parameters as a float. This errors when these parameters are in a list format (eg the `flux_predict_polynomial_coefficients` parameters so we need a way to handle this.

rtuck99:
Fixes #955 

Hyperion should now be able to read configuration files with square-bracket delineated lists.
Parsing errors should now result in a WARNING level log message rather than causing an exception

### To test:
Verify that the unit tests exercise the code and the configuration files can be read


src/hyperion/parameters/beamline_parameters.py
```diff
@@ -30,22 +32,55 @@ def from_lines(cls, config_lines: list[str]):
...
-            elif value in BEAMLINE_PARAMETER_KEYWORDS:
-                pass
-            else:
-                config_pairs[i] = (config_pairs[i][0], float(config_pairs[i][1]))
+            try:
+                # BEAMLINE_PARAMETER_KEYWORDS effectively raw string but whitespace removed
+                if value not in BEAMLINE_PARAMETER_KEYWORDS:
+                    config_pairs[i] = (
+                        config_pairs[i][0],
+                        cls.parse_value(config_pairs[i][1]),
```
olliesilvester, line 40: NIT: maybe slightly easier to read if this is cls.parse_value(value)

dperl-dls: Do the tests pass for you locally @olliesilvester? for some reason everything is broken for me, only on this branch - but I see they have passed in CI.

olliesilvester: No, they're not passing for me either

dperl-dls: Ok thanks, I'll try to figure out what's going on - it's weird, seems unrelated to the code that was touched here but it really is specifically on this branch

dperl-dls: Ah, it was because the parameters hadn't been updated for the dodal fix which stopped noninteger dwell times - the test data has been fixed in main, but the error reporting is poor so I've made an issue for that https://github.com/DiamondLightSource/hyperion/issues/972

src/hyperion/parameters/beamline_parameters.py
```diff
@@ -30,22 +32,55 @@ def from_lines(cls, config_lines: list[str]):
...
+        remaining = value.strip()
+        i = 0
+        while (i := remaining.find(",")) != -1:
+            list_output.append(cls.parse_list_element(remaining[:i]))
+            remaining = remaining[i + 1 :].lstrip()
+        if (i := remaining.find("]")) != -1:
+            list_output.append(cls.parse_list_element(remaining[:i]))
+            remaining = remaining[i + 1 :].lstrip()
+        else:
+            raise Exception("Missing closing ']' in list expression")
```
dperl-dls, line 82: Could: it might be better to have a more specific exception here, e.g. `ValueError`, or define a new one - and catch it above, so that we don't throw away any other exceptions, but since it's on a per-line basis I'm not too worried about it

dperl-dls (approved):
Great, this is much nicer, thanks! One minor comment in code

src/hyperion/parameters/beamline_parameters.py
olliesilvester: Might be nice to have a test for this exception too

src/hyperion/parameters/beamline_parameters.py
```diff
@@ -30,22 +32,55 @@ def from_lines(cls, config_lines: list[str]):
...
+    def parse_list_element(cls, value: str):
+        if value == "Yes":
+            return True
+        elif value == "No":
+            return False
+        else:
+            return float(value)
+
+    @classmethod
+    def parse_list(cls, value: str):
```
olliesilvester, line 71:
```suggestion
    @classmethod
    def parse_list(cls, value: str):
        list_output = []

        if value[-1] != "]":
            raise Exception("Missing closing ']' in list expression")

        remaining = value[:-1]

        for i in remaining.split(","):
            i = i.strip()
            list_output.append(cls.parse_list_element(i))

        return list_output
```
I think this is just personal preference as I always get confused with indices, the logic you have is perfectly fine!

src/hyperion/parameters/tests/test_external_parameters.py
```diff
@@ -49,6 +49,39 @@ def test_beamline_parameters():
...
+    assert params["flux_predict_polynomial_coefficients_5"] == [
+        -0.0000707134131045123,
+        7.0205491504418,
+        -194299.6440518530,
+        1835805807.3974800,
+        -3280251055671.100,
+    ]
+
+
+def test_parse_exception_causes_warning():
```
olliesilvester, line 65: We can assert that the logger raises a warning using patch

```suggestion
@patch("hyperion.parameters.beamline_parameters.LOGGER")
def test_parse_exception_causes_warning(mock_logger):
    params = GDABeamlineParameters.from_file(
        "src/hyperion/parameters/tests/test_data/bad_beamlineParameters"
    )
    assert params["flux_predict_polynomial_coefficients_5"] == [
        -0.0000707134131045123,
        7.0205491504418,
        -194299.6440518530,
        1835805807.3974800,
        -3280251055671.100,
    ]
    mock_logger.warning.assert_called_once()

```

src/hyperion/parameters/beamline_parameters.py
rtuck99: I wrote it that way because I originally intended it to tokenise chunks of the remaining string incrementally, so that it could handle nested lists. However it turned out that due to not having sufficiently well defined syntax we don't have enough information to parse it in as nice a way as I wanted without adding a lot of complexity.

dperl-dls:
@rtuck99 Feel free to merge this yourself when you're happy with it - if someone has approved it with optional changes then we usually leave it to the submitter

src/hyperion/parameters/beamline_parameters.py
```diff
@@ -17,10 +18,11 @@ def __getitem__(self, item: str):
         return self.params[item]
 
     @classmethod
```
DominicOram, line 20: Nit: Generally it's not very pythonic to have a class with lots of static methods, it's preferred to just use bare functions, in this case we're just following what was there before and it's not too bad so not a problem

DominicOram (approved):
My comments were entirely FYIs, feel free to merge without addressing them. Also, if you want GH to link the commits to the issue you need a `#` before the number in the commit message e.g. `(#955) ...`

PR #978, opened 2023-11-15
878 grid scans should move dectector in and open shutter

### LINKED ISSUE #878: Grid scans should move detector in and open shutter
Currently detector motion is done GDA for grid scans. We should do this in Hyperion and wait on it properly

## Acceptance Criteria
* As early as possible Hyperion starts moving the detector and opening the shutter
* Just before data collection we wait on the movement to finish
* Spin off a new ticket to think about smaller plans still being callable on their own

rtuck99:
Fixes #878

Also adds some new testing infrastructure that allows lightweight integration test.

Link to dodal PR (if required):  [https://github.com/DiamondLightSource/dodal/pull/237]()

Please check:

* That I'm getting the detector_position from the right parameters
* That removing the `wait_for_det_to_finish` shouldn't break anything and that the `ready_for_data_collection` is waited at the correct location
* That the values for the various mocks in the unit tests are at least somewhat realistic / robust to future changes
* Is there any corresponding change in the caller to hyperion needed to disable their detector move?
* Whether this sort of test belongs in the unit tests or the integration tests?


src/hyperion/experiment_plans/tests/test_pin_centre_then_xray_centre_plan.py
```diff
@@ -60,3 +68,72 @@ def test_when_pin_centre_xray_centre_called_then_plan_runs_correctly(
...
+            and msg.kwargs["group"] == "ready_for_data_collection",
+        )
+    )
+    assert messages_from_where(
+        messages[2:],
+        lambda msg: msg.command == "open_run"
+        and msg.kwargs["subplan_name"] == "do_fgs",
+    )
+
+
```
dperl-dls, line 135: Nit: maybe it would be better for this function to live with the run engine simulator? it seems reusable

src/hyperion/experiment_plans/tests/run_engine_simulator.py
```diff
@@ -0,0 +1,161 @@
...
+    """Handlers to simulate a basic oav callback firing"""
+    sim.add_handler(
+        "set",
+        "oav_mxsc_enable_callbacks",
+        # XXX what are reasonable values for these?
+        lambda msg: sim.fire_callback(
+            "event",
+            {
+                "data": {
+                    "oav_snapshot_last_saved_path": "xxx",
```
dperl-dls, line 146: these can probably just be something like `"/tmp/image1.png"`, `"/tmp/image2.png"`, `"/tmp/image3.png"`

dperl-dls (changes requested):
This is great! This must be the most well-tested little function we've got. I look forward to using your simulator in other tests.

Although this test is complex and kind of like an integration test as far as components of hyperion are concerned, the distinction between unit/system tests here is basically whether we need to connect to real EPICS PVs (whether from a simulator or the real hardware ones) in order to run the test. If it can run self-sufficiently, it should be grouped with unit tests even if conceptually more complex.

src/hyperion/device_setup_plans/utils.py
```diff
@@ -3,19 +3,36 @@
...
+     Preparation consists of:
+     * Arming the Eiger
+     * Moving the detector to the specified position
+     * Opening the detect shutter
+     If the plan fails it will disarm the eiger.
     """
     yield from bps.abs_set(eiger.do_arm, 1, group=group)
 
+    yield from set_detector_z_position(detector_motion, detector_distance, group)
+    yield from set_shutter(detector_motion, ShutterState.OPEN, group)
```
olliesilvester, line 34: Should we wait for these moves to have completed before doing the next plan? We have the groups here but they're never waited for

rtuck99: The group is already waited for in flyscan_xray_centre_plan, the unit test calls all the subplans and asserts that this group is waited for before the main fast grid scan starts

olliesilvester: ah yes, thanks

dperl-dls (approved):
Brilliant, thanks!

src/hyperion/experiment_plans/tests/test_pin_centre_then_xray_centre_plan.py
```diff
@@ -60,3 +68,72 @@ def test_when_pin_centre_xray_centre_called_then_plan_runs_correctly(
...
+@patch(
+    "hyperion.experiment_plans.pin_centre_then_xray_centre_plan.pin_tip_centre_plan",
+    autospec=True,
+)
+@patch(
+    "hyperion.external_interaction.callbacks.xray_centre.zocalo_callback.XrayCentreZocaloCallback.wait_for_results",
+    lambda self, x: ([0, 0, 0], [1, 1, 1]),
+)
+def test_when_pin_centre_xray_centre_called_then_detector_positioned(
```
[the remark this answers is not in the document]
DominicOram: @rtuck99 - seconding this, it's really good, thank you!

PR #980, opened 2023-11-16
350 automatically add assignee when pr opened

### LINKED ISSUE #350: GH Actions: Automatically add assignee when PR opened
If I add a PR that references an issue that is not assigned to anyone the action will move the issue in to review but it will be there without an assignee. We should add whoever opened the PR as the assignee to the issue.

## Acceptance Criteria
* When a PR is submitted against an issue with no assignee the PR author is added as an assignee to the issue

subinsaji:
Fixes #350 

### To test:
1. Do thing x
2. Confirm thing y happens


.github/workflows/add_assignee_when_pr_opened.yml
```diff
@@ -0,0 +1,48 @@
...
+ pull_request:
+      types: [ready_for_review, opened, review_requested]
+jobs:
+  assign_pr_author_as_assignee:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get project data
+        # This will get information about the project in general and add it to the environment.
+        # * ISSUE_REFERENCED_ID - The ID for the Issue that was referenes at Fixes#999
+        # * AUTHOR_ID - The ID of the person who made a pull request so that the author can be assigned to the Fixes#999
```
DominicOram, line 13: Should: This comment is wrong, it doesn't output `ISSUE_REFERENCED_ID` but instead `ASSIGNABLE_ID`

.github/workflows/add_assignee_when_pr_opened.yml
```diff
@@ -0,0 +1,48 @@
...
+                    edges {
+                      node {
+                        id
+                      }
+                    }
+                  }
+                  author { ... on User {id} }
+                }
+              }
+            }' -f org=$ORGANIZATION -f repo=${REPO_ORG_NAME##*/} -F number=$PULL_REQUEST_NUMBER > project_data.json
```
DominicOram, line 33: Should: This is basically the same as `get_issue_from_pr.yml`. Could we just get that to return the author too and reuse the same job?

.github/workflows/add_assignee_when_pr_opened.yml
```diff
@@ -0,0 +1,48 @@
...
+          
+          echo "$(<project_data.json )"
```
DominicOram, line 35: Should: This was added in for testing, please remove

DominicOram (changes requested):
Must: This looks like it is now broken, I think that you still need the organization that you removed in your last commit.

Otherwise looks good, thank you! 

Slight edit to your instructions on how to test, you need to make a new PR from a branch that has come off this branch as the behaviour will not work on main

subinsaji (commented):
Changes requests are now complete

.github/workflows/get_issue_from_pr.yml
```diff
@@ -13,34 +13,40 @@ on:
...
           ORGANIZATION: ${{ github.repository_owner }}
           REPO_ORG_NAME: ${{ github.repository }} # Contains the repo_name in format org/repo_name
+          PULL_REQUEST_NUMBER: ${{ github.event.pull_request.number }}
         run: |
           gh api graphql -f query='
-            query($org: String!, $repo: String!){
+            query($org: String!, $repo: String!, $number: Int!){
               repository(owner: $org, name: $repo) {
-                pullRequest(number: ${{ inputs.pr_id }}) {
+                pullRequest(number: $number) {
```
DominicOram, line 38: Should: With this change `inputs.pr_id` is never used. We should either remove it and so this is hardcoded to always use the current PR or we should keep using it and pass it in on `add_assignee_when_pr_opened.yml`

DominicOram (changes requested):
Great, thank you. Change requested in code

DominicOram (approved):
Test failures are unrelated, thanks for all your work!

PR #981, opened 2023-11-16
698 refactor external params

dperl-dls:
Turns the external params into pydantic model and vastly simplifies the conversion from external to internal

Link to dodal PR branch: https://github.com/DiamondLightSource/dodal/tree/hyperion_698_refactor_parameters


DominicOram:
Some discussion between @dperl-dls and I. We do want to do a refactor like this but it requires a bit more thought. We should create a mini design document and think about how the design will handle:
* A new technique e.g. florescence
* Adding a new parameter into current techniques
* A new plan that doesn't need much data e.g. move_sample
* The proposed arch for Agamemnon 2.0

dperl-dls:
closing as stale

PR #991, opened 2023-11-21
Separate tests from source code

### LINKED ISSUE #970: Move tests into a separate folder
After discussion we have decided that we want the tests in a separate folder, not in folders next to the code. They should also be split into `tests/system_tests` and `tests/unit_tests`. The reasoning behind this is:
* It follows the DLS pattern of where tests live
* It allows us to more easily just run the unit tests from VScode

## Acceptance criteria
- All tests marked `@pytest.mark.s03` are separated at a top level directory from all tests without this mark

dperl-dls:
Fixes #970 

### To test:
1. Run the unit tests in vscode and see all pass


tests/test_data/scratch/README
```diff
@@ -0,0 +1 @@
+This directory exists for tests to temporarily write files. Anything in here except this readme can be safely deleted.
```
olliesilvester, line 1: I don't know if it really makes a difference, but we could use pytests `tmpdir`fixture for this instead? (https://docs.pytest.org/en/6.2.x/tmpdir.html)

olliesilvester (approved):
This looks good, and I personally prefer this structure for tests a lot more. I haven't manually gone through each file to check they're the same, but everything passes and the coverage is the same. For my comment about tmpdir, I will create a separate PR with this suggested change

PR #995, opened 2023-11-22
Use the tmpdir fixture for tests that create files

### LINKED ISSUE #994: Use pytest's tmpdir fixture for tests which require creating files
To me this feels nicer than having to manually create and remove test files each time

olliesilvester:
Fixes #994

### To test:
1. Run tests
2. Check if change is useful/makes sense

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,10 +3,6 @@ __pycache__/
 *.py[cod]
 *$py.class
 
-# Files written by tests - but not the README!
-tests/test_data/scratch/*
-!tests/test_data/scratch/README
-
 # C extensions
 *.so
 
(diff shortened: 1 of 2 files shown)


dperl-dls (changes requested):
I didn't know about this fixture, this does seem nice!

It doesn't seem like your `temp_directory fixture` is doing anything different to the builtin `tmpdir`? so I think we can just use that. Also in that case you can delete the `.../test_data/scratch` directory and `.gitignore` entries

olliesilvester:
thanks, didnt see it was in the gitignore

dperl-dls (approved):
Looks good, thanks!

DominicOram:
Nit: Can you tidy up the PR name to make more sense to a user please?

Issue #999, opened 2023-11-27
Unit test failing: test_given_started_when_stopped_and_started_again_then_runs


rtuck99:
The test fails with the following message

```
_________ test_given_started_when_stopped_and_started_again_then_runs __________

test_env = ClientAndRunEngine(client=<FlaskClient <Flask 'hyperion.__main__'>>, mock_run_engine=<tests.system_tests.hyperion.test_main_system.MockRunEngine object at 0x7fb887ba5a20>)

    def test_given_started_when_stopped_and_started_again_then_runs(
        test_env: ClientAndRunEngine,
    ):
        test_env.client.put(START_ENDPOINT, data=TEST_PARAMS)
        test_env.client.put(STOP_ENDPOINT)
        response = test_env.client.put(START_ENDPOINT, data=TEST_PARAMS)
        check_status_in_response(response, Status.SUCCESS)
        response = test_env.client.get(STATUS_ENDPOINT)
>       check_status_in_response(response, Status.BUSY)

tests/system_tests/hyperion/test_main_system.py:257: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

response_object = <WrapperTestResponse 38 bytes [200 OK]>
expected_result = <Status.BUSY: 'Busy'>

    def check_status_in_response(response_object, expected_result: Status):
        response_json = json.loads(response_object.data)
>       assert response_json["status"] == expected_result.value
E       AssertionError: assert 'Idle' == 'Busy'
E         - Busy
E         + Idle

tests/system_tests/hyperion/test_main_system.py:180: AssertionError
```

dperl-dls:
Fixed in https://github.com/DiamondLightSource/hyperion/pull/1086

Issue #998, opened 2023-11-27
Unit test failing: test_grid_detection_plan_runs_and_triggers_snapshots


rtuck99:
The unit test test_grid_detection_plan_runs_and_triggers_snapshots fails with the following message:

FAILED tests/unit_tests/experiment_plans/test_grid_detection_plan.py::test_grid_detection_plan_runs_and_triggers_snapshots - AssertionError: assert 0 == 6

dperl-dls:
This was resolved at some point

PR #1001, opened 2023-11-27
Add plan to align DCM and mirrors when changing energy

### LINKED ISSUE #986: Set energy: Apply lookup table value for pitch/roll
~~We should copy the functionality from GFA's `goToPeakGaussian2`in `beamLineSpecificEnergy`~~

~~There is probably ways to do this in bluesky already - peak statistics callback?~~

~~Acceptance criteria:~~

~~- [ ] Create a dodal function / bluesky plan to replicate the logic in this GDA function using the already-made bluesky functionality~~
~~- [ ] Make tests for each part of this logic~~

See https://github.com/DiamondLightSource/hyperion/issues/986#issuecomment-1834054866

rtuck99:
Fixes #986

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/243

### To test:
The code is intended to implement what currently happens in GDA in `align_beam_specific_worker()` between the 
`disableFeedbackLoop()` and `reenableFeedbackLoopIfNeeded()`

* The unit tests have some spurious failures, I've raised #998 and #999

There are a number of implementation issues/questions in the code marked with `###`, namely:
* Initial DCM peak auto adjustment search appears to be centred on around the current pitch position, but I'm not sure where/whether this is actually initialised anywhere?
* There's some code to set the attenuator transmission to 1 and restore it afterwards, I'm not sure if this is something that should be the responsibility of this code, it wasn't mentioned in the parent github issue, so I haven't implemented it, but do you think I should?
* There's a random wait for 2 seconds after the DCM pitch adjusts, but I'm not sure what this is waiting for, is it sufficient to just wait for the group to finish moving?
* HFM initial preset position and search start appear to differ slightly in GDA, but I don't know if this is a deliberate thing or not.
* I couldn't use bps.mv in the unit test for `_adjust_dcm_roll` because the fake EpicsMotor doesn't seem to have a RealPosition attribute, so had to use set (bluesky docs suggest that `mv` is equivalent to `set` with `wait=True`, but looking at the code, this isn't actually the case and `bps.mv` appears to try clever things), is this a known thing?
* The roll angle lookup table code that I wrote, currently does a very primitive peak find which finds the position of the maximum value in the scan. In GDA it does extrapolation from the LUT, the simple numpy linear interpolation implementation doesn't, will this be a problem? Also when I looked at the table in GDA deployed I03 the fit appeared to be completely flat i.e. generated a constant roll angle regardless of Bragg input.
* Originally I thought the gaussian fit wasn't actually called for I03, however as I write this, I now think that maybe it does, so perhaps will have to implement that in another PR. 
* Also I have just noticed in the gaussian code, there is this comment "special" code for I03/04 (depending on function you look at) :
`drive back past peak to start before moving to peak`. This sounds like it's for correcting backlash in mechanism and possibly a good idea, is this something I should do generally?
* Also, would like to try some of this at some point on actual beamline hardware if that's possible to see if it really works!


npat113:
In rough order:

- We don't do the DCM peak search any more, but if it were used again then it would be after the move to the LUT derived value.
- Final step of energy change is to wait for beam to be in position (so feedback on and trans=1). I suspect we already have a plan for this, so can probably use that.
- Wait for pitch is probablynot needed. The basic idea is set all the various motors to the LUT values and then let feedback handle the minor adjustments needed. So wait before feedback if needed.
- HFM shouldn't be used at all. We use DCM roll in feedback for this.
- Roll for I03 is a fixed value. Our beamline configuration is to set a roll value that delivers beam to the same position across the energy range. Minor adjustments handled by feedback.
- Unsure why backlash is being done in GDA and not sure it is needed. 

Energy change code in GDA is ancient. I suspect there's a lot of legacy crud in there that we really don't want to carry forward.

src/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
```diff
@@ -0,0 +1,167 @@
...
+
+HFM_FINE_PITCH_INITIAL_PRESET_MM = 0.017
+HFM_FINE_PITCH_ADJUST_PEAK_WIDTH_MM = 0.016
+HFM_FINE_PITCH_ADJUST_STEP_MM = 0.0006
+
+DCM_PITCH_ADJUST_PEAK_WIDTH = 0.075
+DCM_PITCH_ADJUST_PEAK_STEP = 0.002
+
+
+class DCMPitchRollMirrorAdjuster:
```
DominicOram, line 40: Should: We tend not to hold plans in a class like this, instead choosing to have a composite with the devices and bare generators for the plans (take a look at the other plans). We want to do this as:
* It integrates better with the way `blueapi` (the API package we want to move to) works
* General consistency

src/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
```diff
@@ -0,0 +1,167 @@
...
+        yield from self._adjust_to_peak(
+            self._dcm.pitch,
+            self._qbpm1.intensityC,
+            current_pitch,
+            DCM_PITCH_ADJUST_PEAK_WIDTH,
+            DCM_PITCH_ADJUST_PEAK_STEP,
+        )
+
+        # XXX GDA waits here for 2 seconds, not sure what we are waiting for here?
+        yield from bps.sleep(2)
```
DominicOram, line 84: Should: Our general pattern with these sleeps in GDA has been to remove them, work out what they're actually waiting on when we get to the beamline and then refactor to wait on it properly

DominicOram (changes requested):
Couple of additional points and obviously happy with all of Neil's points. I think it's worth moving it closer to what Neil suggests and then we go down and play with it on the beamline

src/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
rtuck99: I can move the detectors into a composite but I don't think the peak finder and estimator should be included as they aren't really value objects. The purpose of the class is really to separate the domain function (configure the optics) from the strategy (how to find peaks). I'm expecting that it would be invoked by a wrapper function that would instantiate whatever implementations of finders/estimators from a serialized form, that would presumably have some kind of composite. But I haven't written it as I don't know what the ultimate shape of that would be, whether it would consist of only this section of code or whether it would consist of other plan elements as well.

src/hyperion/device_setup_plans/adjuster.py
```diff
@@ -0,0 +1,38 @@
...
+    def adjust(self, group=None) -> Generator[Msg, None, None]:
+        pass
+
+
+class NullAdjuster(Adjuster):
+    def adjust(self, group=None) -> Generator[Msg, None, None]:
+        pass
+
+
+class LUTAdjuster(Adjuster):
```
DominicOram, line 25: Should: I think the structure of this adjuster isn't very pythonic. Unlike Java, python encourages a procedural approach first with an OOP approach taken when the problem calls for it. In this case it's not clear what other Adjusters we will need in the short/medium term so I do not think an OOP approach is required. In fact I would reduce this whole thing to a single function of something like:
```python
def adjust(lookup_table: LookupTableConverter, output_device: EpicsMotor, input):
```

src/hyperion/utils/lookup_table.py
```diff
@@ -0,0 +1,42 @@
...
+
+from hyperion.log import LOGGER
+
+
+class LookupTableConverter(ABC):
+    """Interface for generic lookup table functionality."""
+
+    @abstractmethod
+    def s_to_t(self, s):
+        pass
```
DominicOram, line 14: Could: As above it's not clear what other things will use this interface and so I would not write it yet.

DominicOram (changes requested):
Good work, thank you.

I think we should remove the `PeakFinder` code, it's now not used and I don't think we will need it in the short/medium term. Sorry.

As per slack chat with Neil it looks like we don't really need to wait between any of these sets (except I think we might do on the apply). Can you go through and remove the ones we don't need?

DominicOram:
Additionally, please could you fix the unit tests?

src/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
```diff
@@ -0,0 +1,132 @@
...
+            voltage_channel, required_voltages[i], group=MIRROR_VOLTAGE_GROUP
+        )
+
+    yield from bps.wait(group=MIRROR_VOLTAGE_GROUP)
+
+
+def adjust_mirror_stripe(
+    energy_kev, mirror: FocusingMirror, mirror_voltages: VFMMirrorVoltages
+):
+    # Transmission should be 100% and feedback should be OFF prior to entry
```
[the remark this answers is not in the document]
rtuck99: have changed it to a docstring. What I really wanted to do was put some kind of assert here but there didn't seem to be an easy way to do so without being overly heavyweight.
Agree about transmission, I mainly included it because all existing GDA code does this, and so does our existing feedback decorator.

src/hyperion/device_setup_plans/adjuster.py
dperl-dls: I am inclined to agree, I think it's better not to abstract until we know what all or many of the specific cases look like, and better to use loose functions than classes if they'll accomplish the same job. But I'm interested in hearing arguments for this approach?

src/hyperion/utils/lookup_table.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class LinearInterpolationLUTConverter(LookupTableConverter):
```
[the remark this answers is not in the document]
DominicOram: Written https://github.com/DiamondLightSource/dodal/issues/274 for this

tests/conftest.py
```diff
@@ -355,3 +408,148 @@ def fake_fgs_composite(smargon: Smargon, test_fgs_params: InternalParameters):
...
+
+
+class MessageHandler:
+    def __init__(self, p: Callable[[Msg], bool], r: Callable[[Msg], object]):
+        self.predicate = p
+        self.runnable = r
+
+
+@pytest.fixture
+def sim():
```
DominicOram, line 554: Sorry, can I raise this to a should after seeing it used a bit more? I think `sim_run_engine` would be clearer

DominicOram (changes requested):
Single minor comment, otherwise good. Thanks!

DominicOram (approved):
Great, thank you. Sorry I went through so many iterations on it.

PR #1003, opened 2023-11-28
Make settle time less than validity time in pin tip tests

olliesilvester:
Fixes a test so we can do a release


DominicOram:
Closing in favour of https://github.com/DiamondLightSource/hyperion/pull/1005

PR #1009, opened 2023-11-29
Use correct oav_centring_file address

### LINKED ISSUE #1006: Hotfixes on v8.2.0
When testing v8.2.0 we had a number of issues:
* Latest ophyd-async crashes on start-up https://github.com/bluesky/ophyd-async/issues/89 (should fix with a pin)
* Microns per pixel not initialised correctly - fixed with https://github.com/DiamondLightSource/dodal/pull/246
* JSON files still not pointing to the new `daq_configuration` area
* When manually triggering Hyperion if the stub offsets have already been applied from a previous run it will still try and apply them again anyway, this causes a timeout - https://github.com/DiamondLightSource/hyperion/issues/1007

olliesilvester:
Fixes #1006 along with https://github.com/DiamondLightSource/dodal/pull/248

Instructions for reviewer:
Confirm this new address is correct

### To test:
Run tests


DominicOram (approved):
Great, thank you!

PR #1014, opened 2023-11-30
Allow disabling stub offsets

### LINKED ISSUE #1010: Stub offsets: Provide GDA property to enabl/disable setting stub offsets in Hyperion
We have seen numerous issues with stub offsets to the point that Hyperion has had to be disabled because of it. We should allow the user to just enable/disable stub offset configuring with a GDA property.

## Acceptance Criteria
* There is a GDA property that will disable/enable stub offsets configuring in Hyperion

DominicOram:
Fixes #1010 

Dodal PR at https://github.com/DiamondLightSource/dodal/pull/251

### To test:
1. Confirm tests still pass

## Notes

Please also review https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/40665 at the same time

I'm not a big fan of passing this all the way down like this, especially as we have to then put it in dodal. But hopefully its a temporary measure, open to better suggestions though.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ install_requires =
     xarray
     doct
     databroker
-    dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@0d960bb3ebeb5006544c3195fcadd1b1521aeb14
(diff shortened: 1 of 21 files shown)


DominicOram:
Also made https://github.com/DiamondLightSource/hyperion/issues/1015 to do some tidying up around the tests

Issue #1017, opened 2023-12-02
ISPYB: Deposit robot load entry


DominicOram:
With Hyperion starting to do the robot load we should be putting the robot load entry into ispyb.

Additionally, we would like to do this with the new ispyb REST interface, [exp-eye](https://gitlab.diamond.ac.uk/expeye/core).

## Acceptance Criteria
* There is an ispyb callback to put the robot load data into ispyb
* There are sufficient tests for this

DominicOram:
From looking at current collections it looks like:
* `snapshot_after` is actually a snapshot of the webcam view after robot load
* `snapshot_before` is actually a snapshot after load but from the OAV view

DominicOram:
GDA actually starts the robot load entry, we just need GDA to pass in the id and we can finish it. See loads on https://ispyb.diamond.ac.uk/samples/sid/5054101

DominicOram:
Note that we're not actually going to use this yet, see https://github.com/DiamondLightSource/hyperion/issues/1047#issuecomment-1867666125

DominicOram:
FYI @gfrn. I started some work on this a month ago in the associated PR but dropped it for more urgent things. We would like to try again at some point in March but no fixed idea of when yet.

DominicOram:
Ignore the snapshots part for now, will be done as part of https://github.com/DiamondLightSource/hyperion/issues/1282

DominicOram:
Using existing python API we think S
Using new REST interface we think M

@DominicOram - work out which of these we use

DominicOram:
Discussed with @NKatti2011, we should use `exp-eye`.

PR #1040, opened 2023-12-12
Refactor ISPYB code for robot load: move common parts out and rename

DominicOram:
Part of #1017 

Robot load ispyb entries are very different to data collection ones. This code takes the common parts out and puts them in a utils class.

### To test:
1. Confirm behavior is broadly unchanged
2. Confirm unit tests still pass

### DIFF
--- a/src/hyperion/external_interaction/ispyb/ispyb_utils.py
+++ b/src/hyperion/external_interaction/ispyb/ispyb_utils.py
@@ -1,0 +1,35 @@
+import datetime
+import os
+import re
+
+from ispyb import NoResult
+from ispyb.connector.mysqlsp.main import ISPyBMySQLSPConnector as Connector
+from ispyb.sp.core import Core
+
+from hyperion.parameters.constants import (
+    SIM_ISPYB_CONFIG,
+)
+
+VISIT_PATH_REGEX = r".+/([a-zA-Z]{2}\d{4,5}-\d{1,3})(/?$)"
+
+
+def get_current_time_string():
+    now = datetime.datetime.now()
+    return now.strftime("%Y-%m-%d %H:%M:%S")
+
+
+def get_ispyb_config():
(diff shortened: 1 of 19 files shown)


src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
```diff
@@ -34,7 +36,6 @@
 
 I03_EIGER_DETECTOR = 78
 EIGER_FILE_SUFFIX = "h5"
-VISIT_PATH_REGEX = r".+/([a-zA-Z]{2}\d{4,5}-\d{1,3})(/?$)"
 
 
 class IspybIds(BaseModel):
```
dperl-dls, line 41: I can't comment on the exact location because it's not in the diff but `StoreInIspyb.experiment_type` is defined twice in its `__init__`

src/hyperion/external_interaction/ispyb/ispyb_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    match = re.search(VISIT_PATH_REGEX, path) if path else None
+    return str(match.group(1)) if match else None
+
+
+def get_session_id_from_visit(conn: Connector, visit: str):
+    try:
+        core: Core = conn.core
+        return core.retrieve_visit_id(visit)
+    except NoResult:
+        raise Exception(f"Not found - session ID for visit {visit}")
```
dperl-dls, line 35: Nit: Can this be a more specific exception type?

DominicOram:
Made a mess of this branch, apologies. See https://github.com/DiamondLightSource/hyperion/pull/1053 instead

dperl-dls (approved):
Looks good, couple tiny comments, I'm approving this anyway but please fix the `StoreInIspyb.__init__` one because mypy finds it an error

PR #1044, opened 2023-12-14
Fixed some type errors

DominicOram:
### To test:
1. Run `pyright .` on main and here
2. Confirm number of failures reduced
3. Confirm no behaviour change


dperl-dls (approved):
great, thanks

PR #1053, opened 2023-12-18
Move common parts out of store in ispyb and rename

DominicOram:
Part of #1017 

Robot load ispyb entries are very different to data collection ones. This code takes the common parts out and puts them in a utils class.

### To test:
1. Confirm behavior is broadly unchanged
2. Confirm unit tests still pass


dperl-dls (approved):
See comments here: https://github.com/DiamondLightSource/hyperion/pull/1040#pullrequestreview-1786933444

PR #1054, opened 2023-12-18
Speed up unit tests

### LINKED ISSUE #1050: Speed up unit tests
Our unit tests currently take about 30s to run through, we should improve the speed of them down to <20s. Particularly  tests in `test_grid_detection_plan.py` takes some time

DominicOram:
Fixes #1050

### To test:
1. Run `pytest -m "not s03" --durations=0` on main
2. Run on this branch, confirm tests still pass but are ~10s faster


olliesilvester (approved):
Looks good, units test run in ~15s here compared to ~26s in main

PR #1056, opened 2024-01-02
Remove using import-mode=importlib for tests and add required __init__ files

DominicOram:
Using the `import-mode=importlib` option hides a number of legitimate errors for me and for CI (though interestingly not for @rtuck99). In particular it hides the bad `IspybIds` import that this fixes. I'm not sure why we added this option in in the first place @dperl-dls as it seems to work w/o it?

### To test:
1. Confirm tests still pass


rtuck99 (approved):
I can confirm that this passes on my machine

PR #1060, opened 2024-01-08
Add PandA grid scan and setup plans

olliesilvester:
Fixes #https://github.com/DiamondLightSource/hyperion/issues/870

After this is merged, there are some follow-up issues to add a cleaner way to switch between regular fast grid scans and panda grid scans. The dodal change for this PR pins ophyd-async to a branch which works with the panda and with hyperion

Link to dodal PR (if required): #[279](https://github.com/DiamondLightSource/dodal/pull/279)

### To test:
1. Check that use_panda property makes sense and that Hyperion runs as it would do ordinarily with this set to false.
2. Check new unit tests


olliesilvester:
To get this in a state where we can properly test it with the rest of Hyperion, the `pin_tip_centre_then_xray_centre` will use an extra `using_panda` parameter, in which case it calls `panda_flyscan_xray_centre` instead of the regular `flyscan_xray_centre`. 

Long term solution is https://github.com/DiamondLightSource/hyperion/issues/1067

src/hyperion/experiment_plans/grid_detect_then_xray_centre_plan.py
```diff
@@ -161,11 +189,17 @@ def run_grid_detection_plan(
...
-        parameters, grid_params
-    )
+        flyscan_xray_centre_parameters = (
+            create_parameters_for_panda_flyscan_xray_centre(parameters, grid_params)
+        )
+    else:
+        grid_params = grid_params_callback.get_grid_parameters()
+        flyscan_xray_centre_parameters = create_parameters_for_flyscan_xray_centre(
+            parameters, grid_params
+        )
```
DominicOram, line 202: Nit: There's no reason you can't make these parameters later and just have the one `if` statement for `use_panda`

src/hyperion/experiment_plans/grid_detect_then_xray_centre_plan.py
```diff
@@ -101,6 +118,17 @@ def create_parameters_for_flyscan_xray_centre(
     return flyscan_xray_centre_parameters
 
 
+def create_parameters_for_panda_flyscan_xray_centre(
```
DominicOram, line 121: Could: I think with some generics stuff you can merge this and `create_parameters_for_flyscan_xray_centre`

DominicOram (changes requested):
Great, thanks for getting started with this. I've highlighted a few places where we can tidy it up a bit but we also need a new issue that does it properly. If you like we could make the use of generics into a new issue.

I think that we should also make a new issue to make this more generic to other bits of hardware. For example my immediate thought is that `MM_TO_ENCODER_COUNTS` is a property of the motor so we should be getting that from the motor record somehow. `DEADTIME_S` depends on the detector so should be stored somewhere in the detector data structures etc.

DominicOram:
Have you updated this plan now that the way we interact with Zocalo has changed? We should also add the change made in https://github.com/DiamondLightSource/hyperion/pull/1082

olliesilvester:
The zocalo change is in the panda branch (but untested). I'll also put in #1082

olliesilvester:
The CI failing is due to copied-over warnings from the original `gridscan_internal_params`. I'll make an issue for this but I think it probably makes sense to address it when it comes to properly refactoring the panda code in https://github.com/DiamondLightSource/hyperion/issues/1067

src/hyperion/parameters/plan_specific/panda/panda_gridscan_internal_params.py
```diff
@@ -0,0 +1,139 @@
...
+
+    class Config:
+        arbitrary_types_allowed = True
+        json_encoders = {
+            **DetectorParams.Config.json_encoders,
+            **GridscanIspybParams.Config.json_encoders,
+        }
+
+
+class PandaGridscanInternalParameters(InternalParameters):
```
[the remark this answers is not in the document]
olliesilvester: This wasn't as easy as I'd hoped, but I think it's doable. See https://github.com/DiamondLightSource/hyperion/issues/1085

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def disarm_panda_for_gridscan(panda, group="disarm_panda_gridscan") -> MsgGenerator:
+    yield from bps.abs_set(panda.seq[1].enable, "ZERO", group=group)
+    yield from bps.abs_set(
+        panda.clock[1].enable, "ZERO", group=group
+    )  # While disarming the clock shouldn't be necessery,
+    # it will stop the eiger continuing to trigger if something in the sequencer table goes wrong
+    yield from bps.abs_set(panda.pulse[1].enable, "ZERO", group=group)
+    yield from bps.wait(group="disarm_panda_gridscan", timeout=GENERAL_TIMEOUT)
```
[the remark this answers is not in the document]
olliesilvester: Do you mean in ophyd async? I can make a ticket to add constants / enums for `ZERO` and `ONE`

src/hyperion/experiment_plans/panda_flyscan_xray_centre_plan.py
```diff
@@ -0,0 +1,402 @@
...
+
+    time_between_x_steps_ms = (
+        DEADTIME_S + parameters.hyperion_params.detector_params.exposure_time
+    )
+
+    # Check smargon speed. Exposure time in s, x_step_size in mm. TODO: discuss best place to do this
+    smargon_speed = (
+        parameters.experiment_params.x_step_size * 1e3 / time_between_x_steps_ms
+    )
+    if smargon_speed > 10:
```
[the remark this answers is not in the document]
olliesilvester: The smargon speed is set during the motion script which won't have started yet, so we won't be able to check the speed via a PV at this stage. We could handle this entire check at some other point or in EPICS instead though

src/hyperion/experiment_plans/grid_detect_then_xray_centre_plan.py
olliesilvester: I'll do this when I get around to doing the other generic stuff

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -0,0 +1,147 @@
...
+    # The trigger width should last the same length as the exposure time
+    yield from bps.abs_set(
+        panda.pulse[1].width, 1e-8
+    )  # TODO at some point, thinnk about what constant this shoudl be
+
+    table = get_seq_table(parameters, time_between_x_steps_ms, exposure_time_s)
+
+    LOGGER.info(f"Setting Panda sequencer values: {str(table)}")
+
+    yield from bps.abs_set(panda.seq[1].table, table)
```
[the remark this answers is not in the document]
olliesilvester: Should be able to do both at once

src/hyperion/experiment_plans/panda_flyscan_xray_centre_plan.py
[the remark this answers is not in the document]
olliesilvester: I realised you meant the speed limit, which looks like we can access from BL03I-MO-SGON-01:X.VMAX . I'll add that in

tests/test_data/parameter_json_files/panda_test_parameters.json
```diff
@@ -0,0 +1,75 @@
+{
```
DominicOram, line 1: Could: As I said in the previous review, would be nice to have a test that this can correctly trigger a panda scan

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -0,0 +1,153 @@
...
+
+    row6 = SeqTableRow(
+        trigger=SeqTrigger.POSA_LT,
+        position=parameters.x_start * MM_TO_ENCODER_COUNTS
+        - (MM_TO_ENCODER_COUNTS * (parameters.x_step_size / 2))
+        + (sample_velocity_mm_per_s * exposure_time_s * MM_TO_ENCODER_COUNTS),
+        time2=1,
+    )
+
+    rows = [row1, row2, row3, row4, row5, row6]
```
DominicOram, line 96: Nit: I think this is more readable than before but the use of `rowX` throughout isn't that great. Could you either name each row something related to its function or just do successive `row.append` calls? you could probably also improve readability by making some intermediate variables e.g. `x_start_counts`, `x_end_counts`, `safe_distance`

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -0,0 +1,153 @@
...
+        panda.inenc[1].setp, initial_x * MM_TO_ENCODER_COUNTS, wait=True  # type: ignore
+    )
+    LOGGER.info(
+        f"Initialising panda to {initial_x} mm, {initial_x * MM_TO_ENCODER_COUNTS} counts"
+    )
+
+    yield from bps.abs_set(panda.clock[1].period, time_between_x_steps_ms)  # type: ignore
+
+    # The trigger width should last the same length as the exposure time
+    yield from bps.abs_set(panda.pulse[1].width, DETECTOR_TRIGGER_WIDTH)
```
DominicOram, line 126: Should: Is it my misunderstanding or does the comment say something different to what's actually happening?

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -0,0 +1,153 @@
...
+    """This should load a 'base' panda-flyscan yaml file, then set the
+    specific grid parameters, then adjust the PandAsequencer table to match this new grid
+    """
+    yield from load_device(panda, config_yaml_path)
+
+    yield from bps.abs_set(
+        panda.inenc[1].setp, initial_x * MM_TO_ENCODER_COUNTS, wait=True  # type: ignore
+    )
+    LOGGER.info(
+        f"Initialising panda to {initial_x} mm, {initial_x * MM_TO_ENCODER_COUNTS} counts"
```
DominicOram, line 120: Should: Initialising what? The `x_encoder`? A comment about why this is necessary would be good.

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def setup_panda_for_flyscan(
+    panda: PandA,
+    config_yaml_path: str,
+    parameters: PandAGridScanParams,
+    initial_x: float,
+    exposure_time_s: float,
+    time_between_x_steps_ms: float,
+) -> MsgGenerator:
```
DominicOram, line 110: Should: I think docstrings on some of these args would be good.

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -0,0 +1,153 @@
...
+
+    table = get_seq_table(parameters, time_between_x_steps_ms, exposure_time_s)
+
+    LOGGER.info(f"Setting Panda sequencer values: {str(table)}")
+
+    yield from bps.abs_set(panda.seq[1].table, table, group="panda-config")
+
+    yield from arm_panda_for_gridscan(
+        panda,
+    )
```
DominicOram, line 136: Must: If we're able to do this at the same time as the config we should pass that group in here so that we wait on it properly arming in the next line

src/hyperion/experiment_plans/grid_detect_then_xray_centre_plan.py
```diff
@@ -177,26 +196,64 @@ def run_grid_detection_plan(
...
+            s4_slit_gaps=composite.s4_slit_gaps,
+            smargon=composite.smargon,
+            undulator=composite.undulator,
+            synchrotron=composite.synchrotron,
+            xbpm_feedback=composite.xbpm_feedback,
+            zebra=composite.zebra,
+            zocalo=composite.zocalo,
+            panda=composite.panda,
+            panda_fast_grid_scan=composite.panda_fast_grid_scan,
+        )
```
DominicOram, line 251: Should: What is the difference between this composite and the one above?

DominicOram (changes requested):
Thanks, if you want to get this in quicker you could just address the must and pull other things into a new issue. Could you fix the merge conflicts too please?

src/hyperion/device_setup_plans/setup_panda.py
olliesilvester: I think I'll just remove this logging message - it was made when the panda encoder was getting zero'd at the smargon starting point, but now we zero it based off the absolute smargon position.

src/hyperion/device_setup_plans/setup_panda.py
olliesilvester: Thanks - old comment, it's wrong now

src/hyperion/device_setup_plans/setup_panda.py
DominicOram: I'm not too fussed about logging here but if we're not we should add a comment to the line above as it's not immediately clear why it's needed

src/hyperion/device_setup_plans/setup_panda.py
olliesilvester: Good idea, it's much more readable with the intermediate variables

src/hyperion/device_setup_plans/setup_panda.py
olliesilvester: Actually I think we should wait for the table to be set before arming the PandA - if we arm before the table is properly set, there could be situations where we randomly send trigger signals

tests/test_data/parameter_json_files/panda_test_parameters.json
olliesilvester: The tests in `test_panda_flyscan_xray_centre_plan` now use these parameters

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+    )
+    rows.append(
+        SeqTableRow(
+            position=end_of_grid_x_counts + safe_distance_x_counts,
+            trigger=SeqTrigger.POSA_GT,
+            time2=1,
+        )
+    )
+    parameters.x_step_size
```
DominicOram, line 82: Nit: This line does nothing?

DominicOram (approved):
One very minor nit-pick otherwise this is really great work that you should be very proud of.

PR #1061, opened 2024-01-08
1036 set energy integrate set energy functions into hyperion plans

### LINKED ISSUE #1036: Set energy: integrate set energy functions into Hyperion plans
Once all of the required functions for changing energy have been made in dodal, we should integrate this into Hyperion plan and ensure Hyperion does this instead of GDA

**Acceptance criteria:**

- When Hyperion is being used, GDA no longer does the set energy
- Hyperion does set energy (probably) during the robot load stage
- Sometimes also need to set energy after centring or after one collection
- GDA UI displays the energy that was applied by Hyperion, even if GDA isn't used itself to set the energy directly
- We can still choose to control energy via the old mechanism in GDA via a configuration option.

rtuck99:
Fixes #1036

Link to dodal PR (if required): DiamondLightSource/dodal#280

### To test:
1. If `requested_energy_kev` is specified in the `wait_for_robot_load_then_centre` experiment parameters then the energy will be set on the undulator and the dcm and mirror configuration will be adjusted to suit in addition to waiting for robot load.
2. There is now also a `set_energy` plan which will set the energy and adjust dcm/mirrors without waiting for robot load


src/hyperion/experiment_plans/set_energy_plan.py
dperl-dls, line 1: What is the rationale for set_energy_plan being an experiment plan rather than device setup? In my mind, it's not really taking any experimental measurements, so my intuition says it belongs with the latter, but I'm open to arguments.

If it's just because we want to be able to call it directly from curl or GDA, I think we could also just import and re-export it from `...experiment_plans.__init__.py` even if it otherwise lives with the setup plans, or change the blueAPI context to use `hyperion` as the plan module and export all the plans we want from there

src/hyperion/parameters/plan_specific/set_energy_internal_params.py
dperl-dls, line 1: As with above comment, this seems overkill - we're hardly using anything in here, energy changes don't get their own ISPyB deposition,  etc.

dperl-dls (changes requested):
Sorry I didn't notice this yesterday when we were looking at the param parsing, tunnel vision...

src/hyperion/experiment_plans/set_energy_plan.py
```diff
@@ -0,0 +1,76 @@
...
+
+def set_energy_plan(
+    energy_kev,
+    composite: SetEnergyComposite,
+):
+    yield from transmission_and_xbpm_feedback_for_collection_wrapper(
+        _set_energy_plan(energy_kev, composite),
+        composite.xbpm_feedback,
+        composite.attenuator,
+        0.1,
```
olliesilvester, line 66: Might be nicer to move this to a constant

src/hyperion/experiment_plans/set_energy_plan.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+
+
+def set_energy(
```
olliesilvester, line 70: Is this function used anywhere? It looks a bit strange to me since we're calling a generator but not returning or yielding it

src/hyperion/parameters/plan_specific/set_energy_internal_params.py
DominicOram: Sorry, this might be my bad. I wasn't sure how much of this boilerplate we needed to get a basic set_energy endpoint working so suggested copying this all as a start. If we can get away with less to do a simple call then I agree we should.

dperl-dls: well, it will make the `InternalParameters` type annotation in `__main__.py` be wrong... but that's already not really correct. otherwise, I think we could use any param type to put in the registry as long as it implements `from_json()`

src/hyperion/experiment_plans/set_energy_plan.py
rtuck99: just missing a yield from

src/hyperion/parameters/plan_specific/set_energy_internal_params.py
rtuck99: I think I agree with @dperl-dls here, I think what we need is a second category of endpoint where we can expose basic device setup without requiring all the experimental context.

dperl-dls (approved):
Looks good! I took the liberty of fixing a few tests - hope that's okay - please check my changes before merging

PR #1074, opened 2024-01-16
1071 replace current energy from gda with requested energy and actual energy

### LINKED ISSUE #1071: Replace current_energy specified from GDA with requested_energy and actual energy from PV readback as appropriate
Currently hyperion logs the current_energy sent from GDA in the request as the energy in ispyb and populates the detector params with it as until now hyperion hasn't set the energy itself. When hyperion sets the energy the we should record the readback ourselves and populate ispyb with it. The detector parameters should use the requested energy.

rtuck99:
Fixes #1071 

Link to dodal PR (if required): DiamondLightSource/dodal#293

### To test:
1. `current_energy_ev` is now ignored if specified when invoking wait for robot load expt. and is removed from the detector params, in favour of `expected_energy_ev`
2. The energy is now determined by the DCM energy readback during robot load (after energy changes if requested energy was specified) and ispyb param `current_energy_ev` is populated from this
3. If `requested_energy_kev` is specified in the experiment params, the detector params `expected_energy_ev` is populated from this, otherwise the readback energy is used.

### DIFF
--- a/src/hyperion/experiment_plans/wait_for_robot_load_then_centre_plan.py
+++ b/src/hyperion/experiment_plans/wait_for_robot_load_then_centre_plan.py
@@ -124,6 +124,11 @@ def wait_for_robot_load_then_centre_plan(
             set_energy_composite,
         )
 
+    actual_energy_ev = 1000 * (yield from bps.rd(composite.dcm.energy_in_kev))
+    parameters.hyperion_params.ispyb_params.current_energy_ev = actual_energy_ev
+    if not parameters.experiment_params.requested_energy_kev:
+        parameters.hyperion_params.detector_params.expected_energy_ev = actual_energy_ev
+
     yield from wait_for_smargon_not_disabled(composite.smargon)
 
     params_json = json.loads(parameters.json())
--- a/src/hyperion/experiment_plans/wait_for_robot_load_then_centre_plan.py
+++ b/src/hyperion/experiment_plans/wait_for_robot_load_then_centre_plan.py
@@ -124,11 +124,6 @@ def wait_for_robot_load_then_centre_plan(
             set_energy_composite,
         )
 
-    actual_energy_ev = 1000 * (yield from bps.rd(composite.dcm.energy_in_kev))
-    parameters.hyperion_params.ispyb_params.current_energy_ev = actual_energy_ev
-    if not parameters.experiment_params.requested_energy_kev:
-        parameters.hyperion_params.detector_params.expected_energy_ev = actual_energy_ev
-
     yield from wait_for_smargon_not_disabled(composite.smargon)
 
     params_json = json.loads(parameters.json())
@@ -165,9 +160,14 @@ def wait_for_robot_load_then_centre(
 ) -> MsgGenerator:
     eiger: EigerDetector = composite.eiger
 
+    actual_energy_ev = 1000 * (yield from bps.rd(composite.dcm.energy_in_kev))
+    parameters.hyperion_params.ispyb_params.current_energy_ev = actual_energy_ev
+    if not parameters.experiment_params.requested_energy_kev:
+        parameters.hyperion_params.detector_params.expected_energy_ev = actual_energy_ev
+
     eiger.set_detector_parameters(parameters.hyperion_params.detector_params)
 
-    return start_preparing_data_collection_then_do_plan(
+    yield from start_preparing_data_collection_then_do_plan(
         eiger,
         composite.detector_motion,
         parameters.experiment_params.detector_distance,
--- a/src/hyperion/experiment_plans/wait_for_robot_load_then_centre_plan.py
+++ b/src/hyperion/experiment_plans/wait_for_robot_load_then_centre_plan.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import json
+from typing import cast
 
 import bluesky.plan_stubs as bps
 from blueapi.core import BlueskyContext, MsgGenerator
@@ -38,6 +39,7 @@ from hyperion.experiment_plans.pin_centre_then_xray_centre_plan import (
 )
 from hyperion.experiment_plans.set_energy_plan import (
     SetEnergyComposite,
+    read_energy,
     set_energy_plan,
 )
 from hyperion.log import LOGGER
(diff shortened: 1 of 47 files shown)


src/hyperion/parameters/schemas/detector_parameters_schema.json
```diff
@@ -21,5 +18,5 @@
             "type": "string"
         }
     },
-    "minProperties": 6
+    "minProperties": 5
```
DominicOram, line 21: Should: If `expected_energy_ev` is a valid thing to set it should be in the list above. Then if it's optional you should ennumerate the required ones here, have a look round some of the other schemas, they do this.

src/hyperion/parameters/schemas/full_external_parameters_schema.json
```diff
@@ -3,7 +3,7 @@
     "type": "object",
     "properties": {
         "params_version": {
-            "const": "4.0.2"
+            "const": "4.0.3"
```
DominicOram, line 6: Must: We will need a corresponding GDA change that updates to have this as the expected param version and changes all hyperion calls to use `expected_energy` too

src/hyperion/parameters/plan_specific/wait_for_robot_load_then_center_params.py
```diff
@@ -96,6 +96,11 @@ def _preprocess_hyperion_params(
         all_params["num_images_per_trigger"] = 1
         all_params["trigger_mode"] = TriggerMode.FREE_RUN
         all_params["upper_left"] = np.zeros(3, dtype=np.int32)
+        all_params["expected_energy_ev"] = (
+            experiment_params.requested_energy_kev * 1000
+            if experiment_params.requested_energy_kev
+            else None
+        )
```
DominicOram, line 103: Should: Let's have a discussion (probably with @dperl-dls) to work out if we can come up with a nicer way of using the real energy if one is not requested. Having the logic split between here and the plan isn't great

src/hyperion/experiment_plans/wait_for_robot_load_then_centre_plan.py
```diff
@@ -118,6 +118,11 @@ def wait_for_robot_load_then_centre_plan(
             set_energy_composite,
         )
 
+    actual_energy_ev = 1000 * (yield from bps.rd(composite.dcm.energy_in_kev))
+    parameters.hyperion_params.ispyb_params.current_energy_ev = actual_energy_ev
+    if not parameters.experiment_params.requested_energy_kev:
+        parameters.hyperion_params.detector_params.expected_energy_ev = actual_energy_ev
```
DominicOram, line 124: Must: The energy inside `DetectorParams` must be valid before we do `eiger.set_detector_parameters` below.

src/hyperion/experiment_plans/wait_for_robot_load_then_centre_plan.py
```diff
@@ -118,6 +118,11 @@ def wait_for_robot_load_then_centre_plan(
             set_energy_composite,
         )
 
+    actual_energy_ev = 1000 * (yield from bps.rd(composite.dcm.energy_in_kev))
+    parameters.hyperion_params.ispyb_params.current_energy_ev = actual_energy_ev
```
DominicOram, line 122: Should: A better way to do this is just to read it when we read all the ispyb parameters during the data collection inside `read_hardware_for_ispyb_during_collection`

DominicOram (changes requested):
Good start, I think this is on the way there but we probably need a bit of a wider discussion about some of the edge cases. Some issues to discuss:
* The energy inside detector params must be valid before we trigger the detector
* We want to trigger the detector as early as possible
* If a plan provides an energy it's easy, we just give that to the detector
* If it doesn't we will need to read the current energy
* Does this mean that all plans must provide an energy or that all plans must read the energy?
* Writing to ispyb is a separate issue that we can just do by reading energy during collection

tests/unit_tests/experiment_plans/test_wait_for_robot_load_then_centre.py
```diff
@@ -58,6 +62,11 @@ def test_when_plan_run_then_centring_plan_run_with_expected_parameters(
     wait_for_robot_load_then_centre_params: WaitForRobotLoadThenCentreInternalParameters,
 ):
     RE = RunEngine()
+
+    # async def do_nothing_and_return_current_energy(_):
+    #     return 11.105
+
+    # RE.register_command("set_energy_plan", do_nothing_and_return_current_energy)
```
DominicOram, line 69: Should: can you remove commented out code please?

src/hyperion/experiment_plans/wait_for_robot_load_then_centre_plan.py
rtuck99: I have questions about the witchcraft that happens in that method!

tests/unit_tests/experiment_plans/test_flyscan_xray_centre_plan.py
```diff
@@ -137,7 +137,14 @@ def test_read_hardware_for_ispyb_updates_from_ophyd_devices(
...
         transmission_test_value = 0.01
-        fake_fgs_composite.attenuator.actual_transmission.sim_put(transmission_test_value)  # type: ignore
+        fake_fgs_composite.attenuator.actual_transmission.sim_put(
+            transmission_test_value
+        )  # type: ignore
+
+        current_energy_ev_test_value = 12.05
+        fake_fgs_composite.dcm.energy_in_kev.user_readback.sim_put(
+            current_energy_ev_test_value
+        )  # type: ignore
```
olliesilvester, line 147: CI is failing because this type ignore needs to be on line 145

olliesilvester (commented):
This looks good to me now, but I guess we can't merge until there's a GDA change to update the parameter version. I don't see any parameters to turn the set energy feature on and off like we do with the panda and robot load. Are we testing robot load and set energy together?

rtuck99:
> This looks good to me now, but I guess we can't merge until there's a GDA change to update the parameter version. I don't see any parameters to turn the set energy feature on and off like we do with the panda and robot load. Are we testing robot load and set energy together?

Set energy is controlled by the present of `requested_energy_kev` in wait for robot load parameters. If it is present then we set energy otherwise we read the current energy from the dcm and use that as the detector threshold and record it in ispyb.

There is a corresponding gerrit change referenced above that will also need to go in.

rtuck99:
> Good start, I think this is on the way there but we probably need a bit of a wider discussion about some of the edge cases. Some issues to discuss:
> 
>     * The energy inside detector params must be valid before we stage the detector
> 
>     * We want to stage the detector as early as possible
> 
>     * If a plan provides an energy it's easy, we just give that to the detector
> 
>     * If it doesn't we will need to read the current energy
> 
>     * Does this mean that all plans must provide an energy or that all plans must read the energy?
> 
>     * Maybe we can push some of this logic into `ophyd`? You always read the requested energy from there but it could be initialised to the current energy if it's not been changed?
> 
>     * Writing to ispyb is a separate issue that we can just do by reading energy during collection

I think this is now done?

DominicOram (approved):
Thank you! This looks broadly good to me. I think we can improve a little bit in https://github.com/DiamondLightSource/hyperion/issues/1102 (but this also probably feeds into parameter discussion so worth just getting this merged)

PR #1079, opened 2024-01-16
1073 make external callbacks optional

### LINKED ISSUE #1073: Make running callbacks in a process optional
I was too trigger happy merging https://github.com/DiamondLightSource/hyperion/pull/1065. Given it causes us risk we should have gated it behind a property of some sort. Ideally it would be a GDA property so that the beamline scientist can revert it if there are issues. If we do sufficient beamline testing with this turned on we can remove the property.

## Acceptance Criteria
* There is a GDA property that we can use to switch callbacks running in the same process or not

dperl-dls:
Fixes #1073

Requires GDA change https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/40900 to take effect, but shouldn't break anything on its own

### To test:
1. Run tests
2. Run `hyperion_restart()` from GDA (with the above change applied) and see that hyperion is started with the external callbacks turned off

### DIFF
--- a/run_hyperion.sh
+++ b/run_hyperion.sh
@@ -4,6 +4,8 @@ STOP=0
 START=1
 SKIP_STARTUP_CONNECTION=false
 VERBOSE_EVENT_LOGGING=false
+IN_DEV=false
+EXTERNAL_CALLBACK_SERVICE=false
 LOGGING_LEVEL="INFO"
 
 for option in "$@"; do
@@ -27,9 +29,12 @@ for option in "$@"; do
         --verbose-event-logging)
             VERBOSE_EVENT_LOGGING=true
             ;;
+        --external-callbacks)
+            EXTERNAL_CALLBACK_SERVICE=true
+            ;;
         --logging-level=*)
-        LOGGING_LEVEL="${option#*=}"
-        ;;
+            LOGGING_LEVEL="${option#*=}"
+            ;;
 
         --help|--info|--h)
         
@@ -41,7 +46,6 @@ for option in "$@"; do
             echo "Operations"
             echo "  --stop                  Used to stop a currently running instance of Hyperion. Will override any other operations"
             echo "                          options"
-
             echo "  --no-start              Used to specify that the script should be run without starting the server."
             echo " "
             echo "By default this script will start an Hyperion server unless the --no-start flag is specified."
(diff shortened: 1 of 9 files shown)


run_hyperion.sh
```diff
@@ -124,20 +128,37 @@ if [[ $START == 1 ]]; then
...
 
     #Add future arguments here
-    declare -A args=( ["IN_DEV"]="$IN_DEV" ["SKIP_STARTUP_CONNECTION"]="$SKIP_STARTUP_CONNECTION" ["VERBOSE_EVENT_LOGGING"]="$VERBOSE_EVENT_LOGGING"
-                    ["LOGGING_LEVEL"]="$LOGGING_LEVEL")
-    declare -A arg_strings=( ["IN_DEV"]="--dev" ["SKIP_STARTUP_CONNECTION"]="--skip-startup-connection" ["VERBOSE_EVENT_LOGGING"]="--verbose-event-logging"
-                            ["LOGGING_LEVEL"]="--logging-level=$LOGGING_LEVEL")
-
-    commands=()
-    for i in "${!args[@]}"
+    declare -A h_only_args=(        ["SKIP_STARTUP_CONNECTION"]="$SKIP_STARTUP_CONNECTION"
```
DominicOram, line 131: Could: Having just `h` here (and in the file as a whole) is not so obvious as to what you mean. Maybe `hyperion_main_only_args`?

src/hyperion/experiment_plans/experiment_registry.py
```diff
@@ -54,8 +56,24 @@ def do_nothing():
...
+            GridscanInternalParameters
+            | GridScanWithEdgeDetectInternalParameters
+            | RotationInternalParameters
+            | PinCentreThenXrayCentreInternalParameters
+            | SteppedGridScanInternalParameters
+            | WaitForRobotLoadThenCentreInternalParameters
+        ]
+    )
+    experiment_param_type: type[AbstractExperimentParameterBase]
+    callback_collection_type: type[AbstractPlanCallbackCollection]
```
DominicOram, line 72: ❤️

DominicOram (changes requested):
Love all the additional typing, thanks! Could you add a couple of unit tests that cover the different behavior of `BlueskyRunner` depending on if the `use_external_callbacks` is set?

DominicOram (dismissed):

olliesilvester (approved):
Looks good - that extra test seems trustworthy

PR #1083, opened 2024-01-17
(#1046) Fix eiger not disarming if detector setup fails

### LINKED ISSUE #1046: Eiger not cleaned up if rest of data collection prep fails
If the detector z moves or the detector shutter open fails then the eiger is started arming but does not cleaned up. It should do.

rtuck99:
Fixes #1046

Link to dodal PR (if required):N/A

### To test:
1. Eiger should now disarm if detector move fails or if the shutter fails to open. Note that the detector position and shutter state are not restored.

### DIFF
--- a/tests/unit_tests/device_setup_plans/test_utils.py
+++ b/tests/unit_tests/device_setup_plans/test_utils.py
@@ -3,7 +3,6 @@ from unittest.mock import MagicMock
 import pytest
 from bluesky import FailedStatus
 from bluesky import plan_stubs as bps
-from bluesky.run_engine import RunEngine
 from dodal.beamlines import i03
 from ophyd.status import Status
 
@@ -26,7 +25,7 @@ class TestException(Exception):
 
 
 def test_given_plan_raises_when_exception_raised_then_eiger_disarmed_and_correct_exception_returned(
-    mock_eiger,
+    mock_eiger, RE
 ):
     def my_plan():
         yield from bps.null()
@@ -34,8 +33,6 @@ def test_given_plan_raises_when_exception_raised_then_eiger_disarmed_and_correct
 
     eiger = mock_eiger
     detector_motion = MagicMock()
-
-    RE = RunEngine()
 
     with pytest.raises(TestException):
         RE(
@@ -56,10 +53,9 @@ def null_plan():
 
 
 def test_given_shutter_open_fails_then_eiger_disarmed_and_correct_exception_returned(
-    mock_eiger, null_plan
+    mock_eiger, null_plan, RE
 ):
     detector_motion = MagicMock()
-    RE = RunEngine()
     status = Status(done=True, success=False)
     detector_motion.z.set = MagicMock(return_value=status)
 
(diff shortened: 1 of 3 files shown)


tests/unit_tests/device_setup_plans/test_utils.py
```diff
@@ -37,3 +48,48 @@ def my_plan():
     eiger.async_stage.assert_called_once()
 
     eiger.disarm_detector.assert_called_once()
+
+
+@pytest.fixture()
+def null_plan():
+    yield from bps.null()
```
DominicOram, line 55: Should: This isn't doing what it looks like. Because you can create fixtures with generators this is actually returning a `Msg` so when you use it in the plans below you're passing a `Msg` in to `start_preparing_data_collection_then_do_plan` rather than a generator. This isn't currently causing any errors because you're throwing an exception before hitting that but it might cause confusion down the line. You could fix it here with:
```
@pytest.fixture()
def null_plan():
    return bps.null()
```

tests/unit_tests/device_setup_plans/test_utils.py
```diff
@@ -37,3 +48,48 @@ def my_plan():
...
+    detector_motion = MagicMock()
+    RE = RunEngine()
+    detector_motion.z.set = MagicMock(return_value=Status(done=True, success=False))
+
+    with pytest.raises(FailedStatus):
+        RE(
+            start_preparing_data_collection_then_do_plan(
+                mock_eiger, detector_motion, 100, null_plan
+            )
+        )
```
DominicOram, line 70: Could: It would be nice to check more explicitly that the exception is the one expected:
```python
    failing_status = Status(done=True, success=False)
    detector_motion.z.set = MagicMock(return_value=failing_status)

    with pytest.raises(FailedStatus) as e:
        RE(
            start_preparing_data_collection_then_do_plan(
                mock_eiger, detector_motion, 100, null_plan
            )
        )
   assert e.value == failing_status
```

DominicOram (changes requested):
Thank you. Couple of comments which aren't show stoppers but nice to do I think

tests/unit_tests/device_setup_plans/test_utils.py
rtuck99: well spotted!

tests/unit_tests/device_setup_plans/test_utils.py
```diff
@@ -1,27 +1,38 @@
...
         raise TestException()
 
-    eiger = i03.eiger(fake_with_ophyd_sim=True)
-    eiger.detector_params = MagicMock()
-    eiger.async_stage = MagicMock()
-    eiger.disarm_detector = MagicMock()
+    eiger = mock_eiger
     detector_motion = MagicMock()
 
     RE = RunEngine()
```
olliesilvester, line 38: To make tests cleanup more reliably I think it's best if we always use the `RunEngine` fixture in `tests/conftest.py`

tests/unit_tests/device_setup_plans/test_utils.py
```diff
@@ -37,3 +48,50 @@ def my_plan():
...
+        return_value=Status(done=True, success=False)
+    )
+
+    with pytest.raises(FailedStatus):
+        RE(
+            start_preparing_data_collection_then_do_plan(
+                mock_eiger, detector_motion, 100, null_plan
+            )
+        )
+
```
olliesilvester, line 94: Might as well have the `assert e.value.args[0] is status` part in this test as well

olliesilvester (approved):
thanks!

PR #1084, opened 2024-01-18
add unstage for zocalo

dperl-dls:
Required for https://github.com/DiamondLightSource/dodal/pull/300

### To test:
1. Run tests against version of dodal in linked PR

### DIFF
--- a/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
+++ b/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
@@ -25,6 +25,7 @@ from dodal.devices.xbpm_feedback import XBPMFeedback
 from dodal.devices.zebra import Zebra
 from dodal.devices.zocalo import (
     ZOCALO_READING_PLAN_NAME,
+    ZOCALO_STAGE_GROUP,
     ZocaloResults,
     get_processing_result,
 )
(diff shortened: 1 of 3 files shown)


src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -239,8 +238,13 @@ def run_gridscan_and_move(
     yield from setup_zebra_for_gridscan(fgs_composite.zebra)
 
     LOGGER.info("Starting grid scan")
-
+    yield from bps.stage(
+        fgs_composite.zocalo
+    )  # connect to zocalo and make sure the queue is clear
```
DominicOram, line 243: Should: I'm a bit worried that we're not explicitly waiting on the queue being cleared. If we run this without specifying `wait` I think the behaviour looks to be that it will just not wait so we end up assuming the 2s queue clear is less than the first zocalo message we actually want. Probably a reasonable assumption but I would prefer if we put this in a group that we do wait on at some point, not sure when we would though?

DominicOram (changes requested):
Great thanks, just one comment

DominicOram (approved):
Great, thanks @dperl-dls!

PR #1086, opened 2024-01-19
Fix flaky main system test

dperl-dls:
- Properly destroys the `RunEngine` after each test
- stops some classes holding state they shouldn't
 

Maybe fixes that flaky test, who knows. I ran the tests a little over 50 times with

```bash
for i in {0..50}; do 
    pytest --random-order -m "not s03";
    if [ $? -ne 0 ]; then 
        break; 
    fi; 
done
```

and didn't get any failures.

### DIFF
--- a/src/hyperion/__main__.py
+++ b/src/hyperion/__main__.py
@@ -58,17 +58,15 @@ class ErrorStatusAndMessage(StatusAndMessage):
 
 
 class BlueskyRunner:
-    command_queue: "Queue[Command]" = Queue()
-    current_status: StatusAndMessage = StatusAndMessage(Status.IDLE)
-    last_run_aborted: bool = False
-    aperture_change_callback = ApertureChangeCallback()
-    RE: RunEngine
-    skip_startup_connection: bool
-    context: BlueskyContext
-
     def __init__(
         self, RE: RunEngine, context: BlueskyContext, skip_startup_connection=False
     ) -> None:
+        self.command_queue: Queue[Command] = Queue()
+        self.current_status: StatusAndMessage = StatusAndMessage(Status.IDLE)
+        self.last_run_aborted: bool = False
+        self.aperture_change_callback = ApertureChangeCallback()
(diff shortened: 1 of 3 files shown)


src/hyperion/__main__.py
```diff
@@ -58,17 +58,15 @@ def __init__(self, exception: Exception) -> None:
...
 
 class BlueskyRunner:
-    command_queue: "Queue[Command]" = Queue()
-    current_status: StatusAndMessage = StatusAndMessage(Status.IDLE)
-    last_run_aborted: bool = False
-    aperture_change_callback = ApertureChangeCallback()
-    RE: RunEngine
-    skip_startup_connection: bool
-    context: BlueskyContext
-
```
olliesilvester, line 68: Were these originally class attributes because we assume that there would never be more than one `BlueskyRunner`?

tests/conftest.py
```diff
@@ -115,7 +115,12 @@ def RE():
     RE.subscribe(
         VerbosePlanExecutionLoggingCallback()
     )  # log all events at INFO for easier debugging
-    return RE
+    yield RE
+    try:
+        RE.halt()
```
olliesilvester, line 120: It might be better to use [RE.abort()](https://blueskyproject.io/bluesky/generated/bluesky.run_engine.RunEngine.abort.html#bluesky.run_engine.RunEngine.abort) here as I think this would still do cleanup

olliesilvester (approved):
Looks good, that test appears to be working now, I have one optional comment

src/hyperion/__main__.py
dperl-dls: yes, I guess, or we just didn't know how python worked. we still think there's only one BlueskyRunner in real life, but in tests they seem to not always get garbage collected quickly enough

tests/conftest.py
dperl-dls: I explicitly don't want to do cleanup :)

PR #1093, opened 2024-01-22
920 Combine rotations in ispyb

### LINKED ISSUE #920: Rotation Scan: Combine multiple chi sweeps into one dc group
When multiple rotation scans are done at different chi GDA will put them both in a single data collection group. We should do the same based on whether the last collection had the same sample ID

To avoid making things worse, this will first require:

- [x] #947

dperl-dls:
Fixes #920 

`RotationIspybCallback` now holds on to the previous experiment's sampleID, and if the next one matches, reuses the DCGID for the ISPYB deposition

### To test:
1. Run tests

### DIFF
--- a/src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
+++ b/src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
@@ -17,6 +17,8 @@ from hyperion.parameters.plan_specific.rotation_scan_internal_params import (
 
 if TYPE_CHECKING:
     from event_model.documents.event import Event
+    from event_model.documents.run_start import RunStart
+    from event_model.documents.run_stop import RunStop
 
 
 class RotationISPyBCallback(BaseISPyBCallback):
@@ -44,7 +46,7 @@ class RotationISPyBCallback(BaseISPyBCallback):
         assert isinstance(self.ispyb_ids.data_collection_ids, int)
         self._append_to_comment(self.ispyb_ids.data_collection_ids, comment)
 
-    def activity_gated_start(self, doc: dict):
+    def activity_gated_start(self, doc: RunStart):
         if doc.get("subplan_name") == ROTATION_OUTER_PLAN:
             ISPYB_LOGGER.info(
                 "ISPyB callback recieved start document with experiment parameters."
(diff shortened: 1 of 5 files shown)


src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
```diff
@@ -42,18 +51,26 @@ def activity_gated_start(self, doc: dict):
...
         ISPYB_LOGGER.info("ISPYB handler received start document.")
         if doc.get("subplan_name") == ROTATION_PLAN_MAIN:
             self.uid_to_finalize_on = doc.get("uid")
 
-    def activity_gated_event(self, doc: dict):
+    def activity_gated_event(self, doc: Event):
         super().activity_gated_event(doc)
         set_dcgid_tag(self.ispyb_ids.data_collection_group_id)
 
     def activity_gated_stop(self, doc: dict):
```
olliesilvester, line 73: Should the type hint for doc also be Event?

src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
```diff
@@ -42,18 +51,26 @@ def activity_gated_start(self, doc: dict):
             )
             json_params = doc.get("hyperion_internal_parameters")
             self.params = RotationInternalParameters.from_json(json_params)
-            self.ispyb: StoreRotationInIspyb = StoreRotationInIspyb(
-                self.ispyb_config, self.params
+            dcgid = (
```
olliesilvester, line 54: nit: Personally I prefer `dcgid` to be called the longer `data_collection_group_id`

tests/unit_tests/external_interaction/callbacks/test_rotation_callbacks.py
```diff
@@ -316,3 +321,52 @@ def after_main_do(callbacks: RotationCallbackCollection):
         autospec=True,
     ):
         RE(fake_rotation_scan(params, cb, after_open_do, after_main_do))
+
+
+def random_ids() -> IspybIds:
+    rand_100 = partial(randrange, 100)
+    return IspybIds(
```
olliesilvester, line 328: I think we should avoid having unit tests use random numbers

olliesilvester (changes requested):
This looks good, I think probably just remove the usage of random numbers in tests so that they are properly repeatable (maybe up for debate, but I've had previous reviews saying this)

src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
dperl-dls: In general I agree, but it depends on the scope - this name only lasts for one statement and is never accessed elsewhere. I changed it in `StoreRotationInIspyb` based on this comment though

olliesilvester (approved):
Looks good

PR #1112, opened 2024-02-02
#1101 Wait for moves before FGS kickoff

### LINKED ISSUE #1101: We don't know how long to wait between setting FGS params and being able to hit `PROC`
In the changes between v8.5.0 and v.8.7.0, we have reduced the timing between setting the FGS params from ~250 ms to ~180 ms (eg [v8.7.0](https://graylog2.diamond.ac.uk/streams/631745613d8af00a49bfee55/search?rangetype=absolute&fields=message%2Csource&width=1500&highlightMessage=&from=2024-01-26T13%3A55%3A30.000Z&to=2024-01-26T13%3A57%3A00.000Z&q=) vs [v8.5.0](https://graylog2.diamond.ac.uk/streams/631745613d8af00a49bfee55/search?rangetype=absolute&fields=message%2Csource&width=1500&highlightMessage=&from=2024-01-26T04%3A09%3A02.000Z&to=2024-01-26T04%3A10%3A00.000Z&q=)) due to shuffling around of where callbacks are triggered. The has exposed a race condition where setting `:PROC` (in [FGS.kickoff()](https://github.com/DiamondLightSource/dodal/blob/0c0a5e4f475cf88e947be2177e72c1d16d2a392d/src/dodal/devices/fast_grid_scan.py#L280)) sometimes never completes.

This can be patched over with a sleep, but ultimately we will want confirmation from EPICS of whether we can write or not. Ideally we should be able to trust the `SCAN_VALID` PV and write to the `PROC` PV as soon as we receive the callback that the position counter has been set to 0.

### Acceptance criteria:
We could run a plan with `fast_grid_scan.set_fast_grid_scan_params()` and `FastGridScan.kickoff()` immediately following, and the kickoff would be accepted.

dperl-dls:
Fixes #1101 

Link to dodal PR (if required): [#317](https://github.com/DiamondLightSource/dodal/pull/317)

Has already been tested on the beamline.

### To test:
1. See tests aren't failing.

### DIFF
--- a/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
+++ b/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
@@ -214,8 +214,9 @@ def run_gridscan(
             total_exposure,
             30.0,
         )
+        LOGGER.info("Wait for all moves with no assigned group")
+        yield from bps.wait()
         LOGGER.info("kicking off FGS")
-        yield from bps.wait()  # Wait for all moves with no assigned group
         yield from bps.kickoff(fgs_motors)
         LOGGER.info("Waiting for Zocalo device queue to have been cleared...")
         yield from bps.wait(
--- a/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
+++ b/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
@@ -215,6 +215,7 @@ def run_gridscan(
             30.0,
         )
         LOGGER.info("kicking off FGS")
+        yield from bps.wait()  # Wait for all moves with no assigned group
         yield from bps.kickoff(fgs_motors)
         LOGGER.info("Waiting for Zocalo device queue to have been cleared...")
         yield from bps.wait(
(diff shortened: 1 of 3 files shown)


src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -215,6 +215,7 @@ def do_fgs():
             30.0,
         )
         LOGGER.info("kicking off FGS")
+        yield from bps.wait()  # Wait for all moves with no assigned group
```
DominicOram, line 218: Should: The ordering of the logs here is deceptive. I think we should have two logs, one before the wait that logs basically the same as the comment then we log saying we're kicking off and do the kick off

DominicOram (changes requested):
Thanks, minor comment

olliesilvester:
Can we apply the same change to `panda_flyscan_xray_centre_plan` I know it's really annoying having to do everything twice, I'll try and get them merged soon!

DominicOram (approved):

PR #1124, opened 2024-02-05
Retry writing to the zebra if it fails

### LINKED ISSUE #1117: Zebra PVs occassionally failing to PV callback
Yesterday we saw two times where we failed to set PVs on the zebra [here](https://graylog2.diamond.ac.uk/streams/631745613d8af00a49bfee55/search?rangetype=absolute&fields=message%2Csource&width=1500&highlightMessage=&from=2024-02-01T19%3A00%3A00.000Z&to=2024-02-01T19%3A03%3A00.000Z&q=) and [here](https://graylog2.diamond.ac.uk/streams/631745613d8af00a49bfee55/search?rangetype=absolute&fields=message%2Csource&width=3440&highlightMessage=&from=2024-02-01T21%3A03%3A23.000Z&to=2024-02-01T21%3A05%3A35.000Z&q=). Given this is a failure in the EPICS layer it will be hard to debug but we should:
* Swallow any put complete errors when setting up the zebra
* Log when these errors do occur
* Keep an eye on their frequency

DominicOram:
Fixes #1117

This is quite a generic decorator and so maybe it makes sense in `dodal` but I really don't like that we have to do this so I would rather not encourage it elsewhere. Happy to be argued with though

### To test:
1. Confirm new tests are good and pass

### DIFF
--- a/src/hyperion/device_setup_plans/setup_zebra.py
+++ b/src/hyperion/device_setup_plans/setup_zebra.py
@@ -1,4 +1,7 @@
+from typing import Callable
+
 import bluesky.plan_stubs as bps
+import bluesky.preprocessors as bpp
 from dodal.devices.zebra import (
     DISCONNECT,
     IN1_TTL,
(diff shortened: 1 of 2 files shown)


src/hyperion/device_setup_plans/setup_zebra.py
```diff
@@ -20,6 +23,24 @@
 from hyperion.log import LOGGER
 
 
+def bluesky_retry():
```
olliesilvester, line 26: nit: Might be useful to have a comment saying that this is a decorator used to retry a bluesky plan upon exception

olliesilvester (approved):
Looks good! Personally I find decorators more intuitive like:

```
def retry_plan_decorator(func_to_decorate):
   def new_func():
      ...
   return new_func
```
Rather than the extra function with no parameters. Might just be me though

PR #1126, opened 2024-02-05
Panda grid scans: Add half a Y step to both sides of grid given an odd number of Y steps

### LINKED ISSUE #1118: Panda grid scans: position compare invalid after scanning through an odd number of rows
The [panda fast grid scan motion script](https://gitlab.diamond.ac.uk/controls/motion/bl03i/-/blob/master/Settings/BL03I-MO-PMAC-03/PLCs/12_panda_grid_scan.pmc?ref_type=heads) currently resets the smargon back to its start conditions after the omega rotation. Meanwhile, the panda's position compare expects the snake trajectory to continue through the rotation. This means that the position compare values will be incorrect if we scan through an odd number of rows after its rotation. 

The simplest solution here is to enforce that our grid scans always have an even number of rows for the first 2d grid scan (after rotation it doesn't matter). During grid detect, we should add one to the number of rows in the first grid if the original value is odd. This shouldn't cost us any time since the motion is the same either way.

### Acceptance criteria

-  Hyperion's grid detection plan forces number of rows on the first grid to be even (if odd then +1)
- Unit tests

olliesilvester:
Fixes #1118 

### To test:
1. Make sure changes are sensible
2. Check new tests and changes to old ones

### DIFF
--- a/src/hyperion/parameters/plan_specific/gridscan_internal_params.py
+++ b/src/hyperion/parameters/plan_specific/gridscan_internal_params.py
@@ -21,8 +21,9 @@ from hyperion.parameters.internal_parameters import (
 )
 
 
-class YStepOddNumberException(Exception):
+class OddYStepsException(Exception):
     pass
+
 
 class GridscanHyperionParameters(HyperionParameters):
     ispyb_params: GridscanIspybParams = GridscanIspybParams(
--- a/src/hyperion/parameters/plan_specific/gridscan_internal_params.py
+++ b/src/hyperion/parameters/plan_specific/gridscan_internal_params.py
@@ -13,6 +13,7 @@ from hyperion.external_interaction.ispyb.ispyb_dataclass import (
     GRIDSCAN_ISPYB_PARAM_DEFAULTS,
     GridscanIspybParams,
 )
+from hyperion.log import LOGGER
 from hyperion.parameters.internal_parameters import (
     HyperionParameters,
     InternalParameters,
@@ -59,6 +60,12 @@ class GridscanInternalParameters(InternalParameters):
         cls,
         experiment_params: dict[str, Any],
     ):
+
+        #Force first grid scan to have even number of rows so next row starts at same point
+        if experiment_params['y_steps']%2:
+            experiment_params['y_steps'] += 1
+            LOGGER.debug(f"Making Y steps an even number by increasing it to {experiment_params['y_steps']}")
+
         return GridScanParams(
             **extract_experiment_params_from_flat_dict(
                 GridScanParams, experiment_params
--- a/src/hyperion/parameters/plan_specific/gridscan_internal_params.py
+++ b/src/hyperion/parameters/plan_specific/gridscan_internal_params.py
@@ -63,11 +63,6 @@ class GridscanInternalParameters(InternalParameters):
         experiment_params: dict[str, Any],
     ):
 
-        #FGS motion script moves sample to initial X position for start of second grid, so we need an even number of
-        #rows during the first grid to ensure that any position tracking done in the software is consistent
-        if experiment_params['y_steps']%2:
-            raise YStepOddNumberException("The number of Y steps must be even")
-
         return GridScanParams(
             **extract_experiment_params_from_flat_dict(
                 GridScanParams, experiment_params
--- a/src/hyperion/parameters/plan_specific/gridscan_internal_params.py
+++ b/src/hyperion/parameters/plan_specific/gridscan_internal_params.py
@@ -21,7 +21,7 @@ from hyperion.parameters.internal_parameters import (
 )
 
 
-class OddNumberException(Exception):
+class YStepOddNumberException(Exception):
     pass
 
 class GridscanHyperionParameters(HyperionParameters):
@@ -66,7 +66,7 @@ class GridscanInternalParameters(InternalParameters):
         #FGS motion script moves sample to initial X position for start of second grid, so we need an even number of
         #rows during the first grid to ensure that any position tracking done in the software is consistent
         if experiment_params['y_steps']%2:
-            raise OddNumberException("The number of Y steps must be even")
+            raise YStepOddNumberException("The number of Y steps must be even")
 
         return GridScanParams(
             **extract_experiment_params_from_flat_dict(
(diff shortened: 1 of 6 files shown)


olliesilvester:
We have the option of only having this in panda scans, or doing it in regular FGS too. Not sure which is preferable, but I've included in both for now.

src/hyperion/parameters/plan_specific/gridscan_internal_params.py
```diff
@@ -59,6 +60,12 @@ def _preprocess_experiment_params(
         cls,
         experiment_params: dict[str, Any],
     ):
+
+        #Force first grid scan to have even number of rows so next grid starts at same point
+        if experiment_params['y_steps']%2:
```
DominicOram, line 65: Should: This has different whitespace than the one below. I feel like the one below is more correct

src/hyperion/parameters/plan_specific/gridscan_internal_params.py
```diff
@@ -59,6 +60,12 @@ def _preprocess_experiment_params(
         cls,
         experiment_params: dict[str, Any],
     ):
+
+        #Force first grid scan to have even number of rows so next grid starts at same point
```
DominicOram, line 64: Could: I think maybe a bit more explanation here as to why we need to do it would be good.

DominicOram (changes requested):
Unfortunately I don't think this is going to work because the grid detection will draw a smaller grid on the OAV image. I think that we need to push this check to be earlier, whilst we're doing grid detection.

tests/unit_tests/experiment_plans/test_grid_detection_plan.py
```diff
@@ -254,7 +254,7 @@ def test_when_grid_detection_plan_run_then_grid_detection_callback_gets_correct_
     assert my_grid_params.y_step_size == pytest.approx(0.02)
     assert my_grid_params.z_step_size == pytest.approx(0.02)
     assert my_grid_params.x_steps == pytest.approx(9)
-    assert my_grid_params.y_steps == pytest.approx(1)
+    assert my_grid_params.y_steps == pytest.approx(2)
```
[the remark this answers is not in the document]
DominicOram: Should: Yes, it probably does but I think it would benefit from an explicit test too

DominicOram (changes requested):
This is good, thanks. Could you update https://github.com/DiamondLightSource/hyperion/wiki/PandA-constant%E2%80%90motion-scanning with some mention of having to be an even number in the first grid and why. 

For correcting it in the grid detection I think it might be worth being a bit smarter in the long run. What this is currently doing is just adding a line at the bottom, that line is very unlikely to hit the crystal. Instead we should probably shift the whole grid up by 1/2 a box when we add the extra box. Happy to do this in a new issue if you like? Might need a chat with @neil-i03 to make sure there's not a cleaner way.

olliesilvester:
> For correcting it in the grid detection I think it might be worth being a bit smarter in the long run. What this is currently doing is just adding a line at the bottom, that line is very unlikely to hit the crystal. Instead we should probably shift the whole grid up by 1/2 a box when we add the extra box. Happy to do this in a new issue if you like? Might need a chat with @neil-i03 to make sure there's not a cleaner way.

Good idea - I think we might as well do this in the same PR

src/hyperion/parameters/plan_specific/gridscan_internal_params.py
olliesilvester: Good idea, no point throwing an error for the regular FGS here.

tests/unit_tests/experiment_plans/test_grid_detection_plan.py
```diff
@@ -265,3 +276,92 @@ def test_when_grid_detection_plan_run_then_grid_detection_callback_gets_correct_
...
+    "odd",
+    [(True), (False)],
+)
+@patch("dodal.beamlines.beamline_utils.active_device_is_same_type", lambda a, b: True)
+@patch("bluesky.plan_stubs.sleep", new=MagicMock())
+@patch(
+    "hyperion.experiment_plans.oav_grid_detection_plan.wait_for_tip_to_be_found_ad_mxsc"
+)
+@patch("hyperion.experiment_plans.oav_grid_detection_plan.LOGGER")
+def test_when_detected_grid_has_odd_y_steps_then_add_a_y_step_and_shift_grid(
```
olliesilvester, line 291: I don't know how I feel about this test, it looks horrible to read and is very long

src/hyperion/parameters/plan_specific/panda/panda_gridscan_internal_params.py
```diff
@@ -48,6 +49,12 @@ def _preprocess_experiment_params(
         cls,
         experiment_params: dict[str, Any],
     ):
+
+        # Panda not configured to run a half complete snake so enforce even rows on first grid
+        # See https://github.com/DiamondLightSource/hyperion/wiki/PandA-constant%E2%80%90motion-scanning#motion-program-summary
+        if experiment_params["y_steps"] % 2:
```
dperl-dls, line 55: Is this only important for a 3D grid scan? If so we should check that too.

src/hyperion/parameters/plan_specific/gridscan_internal_params.py
```diff
@@ -21,6 +21,9 @@
 )
 
 
+class YStepOddNumberException(Exception):
```
dperl-dls, line 24: Nit: could just call this `OddYSteps` or something

tests/unit_tests/experiment_plans/test_grid_detection_plan.py
dperl-dls: I think it's okay. It's a lot of setup but that seems unavoidable for where the stuff is you're testing - in an extremely long function (this is why splitting stuff up into little functions makes it more testable). I don't think it's too bad to read. All of our tests for plans are a bit like this.

dperl-dls (approved):
This looks good! I would like it if `PandAGridScanParams` would also validate that `y_steps` must be even if `z_steps > 1`, since it is a requirement of the device and not really of hyperion's experiment. Maybe make a separate issue for that?

src/hyperion/parameters/plan_specific/panda/panda_gridscan_internal_params.py
olliesilvester: Good catch thanks

olliesilvester:
> This looks good! I would like it if `PandAGridScanParams` would also validate that `y_steps` must be even if `z_steps > 1`, since it is a requirement of the device and not really of hyperion's experiment. Maybe make a separate issue for that?

Yeah that's a good point, I'll make a new issue for that. The device would only be invalid if y_steps are odd AND z_steps > 0 though, right?

PR #1134, opened 2024-02-07
#1133 don't bundle screening collections with rotation scan

### LINKED ISSUE #1133: Don't put screening collections in DCG with rotation scans
On discussion with beamline staff, a check that number of images < 200 is sufficient for now, more complicated logic for bundling plans can be implemented when UDC does not rely on GDA

dperl-dls:
Fixes #1133

### To test:
1. Check tests aren't broken

### DIFF
--- a/src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
+++ b/src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
@@ -74,7 +74,6 @@ class RotationISPyBCallback(BaseISPyBCallback):
                 )
                 self.last_sample_id = self.params.hyperion_params.ispyb_params.sample_id
             self.ispyb = StoreRotationInIspyb(self.ispyb_config, self.params, dcgid)
-            self.last_sample_id = self.params.hyperion_params.ispyb_params.sample_id
         self.ispyb_ids: IspybIds = IspybIds()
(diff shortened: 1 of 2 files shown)


src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
```diff
@@ -61,6 +61,8 @@ def activity_gated_start(self, doc: RunStart):
                 )
                 else None
             )
+            if self.params.experiment_params.get_num_images() < 200:
```
DominicOram, line 64: Should: Could you pull this into a named constant please?

src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
```diff
@@ -61,6 +61,8 @@ def activity_gated_start(self, doc: RunStart):
...
+                dcgid = None
```
DominicOram, line 65: Should: I agree with codecov that this should have a test. If nothing else so that it's more obvious what the use case is

DominicOram (changes requested):
Ta. A little bit of tidying up in code. Also, could you make an issue for thinking about the "proper fix"

DominicOram (approved):
Great, thank you!

PR #1135, opened 2024-02-07
Use dodal PulseOutput device for the Zebra output panel

noemifrisina:
Small device added to the Zebra in dodal, updating hyperion to use it.

Link to dodal PR (if required): [#229](https://github.com/DiamondLightSource/dodal/pull/299)

### To test:
1. Check correct device used
2. Check tests still pass


dperl-dls (changes requested):
Can the fields on the dodal device please be named `input`, `delay`, and `width` rather than the abbreviated versions? otherwise looks good.

dperl-dls (approved):
Thanks!

PR #1136, opened 2024-02-08
1114 ispyb callback should make deposition on start

### LINKED ISSUE #1114: Ispyb Callback should make deposition on start
When errors occur on Hyperion we're often getting `ISPyBDepositionNotMade('ispyb was not initialised at run start')` reported. This is because the scan has failed and when we try to write the failure to ispyb we have not yet made the initial deposition. We want the errors to be in ispyb correctly so we should make an initial ispyb deposition on the `start` of the callback, even if we don't have all the information yet. When we get the information in subsequent events ispyb should be updated.

## Acceptance Criteria
* When the ispybCallback is started it makes an initial ispyb deposition
* When an error happens at any point during the scan that the IspybCallback is listening to it gets put in ispyb
* Other information that we read during the plan gets put into ispyb as expected

rtuck99:
Fixes #1114

Would appreciate some feedback on the changes.

I'm not sure I'm happy with the unit tests in the `store_datacollection_in_ispyb` classes - it feels like there is a lot of testing of internals and not much testing of actual ispyb interaction which is presumably what is more interesting.

I'm now in the process of splitting the them out into separate files to make navigation easier, also would like to rename them to `<scan_type>_ispyb_store` as prefer classes named as nouns rather than verbs. Opinions please.

As a result the UT changes here so far are a bit scrappy and mostly focused on just keeping them all working.

**Edit:**

I have pushed an additional set of changes which separate out the classes into separate files. For the changes up to that point see [3a3e8dd](https://github.com/DiamondLightSource/hyperion/pull/1136/commits/3a3e8dda9ef0452d8fc584d7e2791a8b1a0818a2)

Additional refactoring will be in another PR


src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
```diff
@@ -173,22 +178,34 @@ def _store_data_collection_group_table(self, conn: Connector) -> int:
...
+        return self._upsert_data_collection_group(conn, params)
+
+    @staticmethod
+    def _upsert_data_collection_group(
+        conn: Connector, params: StrictOrderedDict
+    ) -> int:
+        return conn.mx_acquisition.upsert_data_collection_group(list(params.values()))
+
+    @staticmethod
+    def _upsert_data_collection(conn: Connector, params: StrictOrderedDict) -> int:
```
olliesilvester, line 190: What's the difference between these two functions? Could we combine them?

src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
```diff
@@ -286,19 +303,37 @@ def _mutate_data_collection_params_for_experiment(
...
+        self._store_data_collection_table(conn, self.data_collection_group_id)
+        self._store_position_table(conn, self.data_collection_id)
 
-        return data_collection_id, self.data_collection_group_id
+        return self.data_collection_id, self.data_collection_group_id
 
     def begin_deposition(self) -> IspybIds:
+        # prevent pyright + black fighting
+        # fmt: off
+        with ispyb.open(self.ISPYB_CONFIG_PATH) as conn:
```
olliesilvester, line 321: Should: Have the `assert conn is not None` like in `update_deposition`

src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
```diff
@@ -286,19 +303,37 @@ def _mutate_data_collection_params_for_experiment(
...
+            if not self.data_collection_group_id:
+                self.data_collection_group_id = self._store_data_collection_group_table(
+                    conn  # type: ignore
+                )
+            if not self.data_collection_id:
```
olliesilvester, line 326: Should this always be none at the time we reach this function? If so, maybe throw an error or have some kind of warning if it isn't none

src/hyperion/external_interaction/callbacks/ispyb_callback_base.py
```diff
@@ -106,7 +106,7 @@ def activity_gated_event(self, doc: Event):
             )
 
             ISPYB_LOGGER.info("Creating ispyb entry.")
```
olliesilvester, line 108: Change to `updating ispyb entry?`

src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
```diff
@@ -286,19 +303,37 @@ def _mutate_data_collection_params_for_experiment(
...
+            self.data_collection_id
+        ), "Attempted to store scan data without a collection"
+        self._store_data_collection_group_table(conn)
+        self._store_data_collection_table(conn, self.data_collection_group_id)
+        self._store_position_table(conn, self.data_collection_id)
 
-        return data_collection_id, self.data_collection_group_id
+        return self.data_collection_id, self.data_collection_group_id
 
     def begin_deposition(self) -> IspybIds:
```
olliesilvester, line 318: Since we always want to do this at the run document, it might be worth calling it in `ispyb_callback_base`'s `activity_gated_start`

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -58,9 +58,11 @@ def activity_gated_start(self, doc: dict):
             self.params = GridscanInternalParameters.from_json(json_params)
             self.ispyb = (
                 Store3DGridscanInIspyb(self.ispyb_config, self.params)
+                # XXX Does this parameter even exist any more?
```
olliesilvester, line 61: It looks like it yeah, but it's a property of the FGS device rather than an input parameter

src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
```diff
@@ -173,22 +178,34 @@ def _store_data_collection_group_table(self, conn: Connector) -> int:
...
-        self, conn: Connector, data_collection_group_id: int
+        self,
+        conn: Connector,
+        data_collection_group_id: int,
+        data_collection_id: Optional[int] = None,
     ) -> int:
-        assert (
-            self.ispyb_params is not None
-            and self.detector_params is not None
-            and self.xtal_snapshots is not None
```
olliesilvester, line 185: Why do we no longer need to do the assert  on `self.xtal_snapshots`?

src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
```diff
@@ -486,15 +551,40 @@ class Store2DGridscanInIspyb(StoreGridscanInIspyb):
...
+        # fmt: on
+
     def _store_scan_data(self, conn: Connector):
-        data_collection_group_id = self._store_data_collection_group_table(conn)
+        assert (
+            self.data_collection_group_id
+        ), "Attempted to store scan data without a collection group"
+        assert (
+            self.data_collection_ids
+        ), "Attempted to store scan data without a collection"
```
olliesilvester, line 578:
```suggestion
        ), "Attempted to store scan data without at least one collection"
```

src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
rtuck99: DataCollection and DataCollectionGroup are two different tables in ispyb - we probably want to keep them separate as sometimes we are doing an update and sometimes we are doing an insert, depending on whether we supply the row id.

There are a few bugs I have fixed in my workspace where we currently aren't supplying these on the `update_deposition()`

src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
rtuck99: Maybe - although TBH I would be surprised if it didn't raise an exception in the actual call to open if there was a problem

src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
rtuck99: I'm assuming that we won't have crystal snapshots at `begin_deposition` but perhaps I'm wrong? Are these fields supplied by GDA or is this something hyperion does?

olliesilvester: Not sure actually, I think GDA takes snapshots of a sample once it gets loaded, but I'm not sure which snapshots these are referring to. @dperl-dls will know!

DominicOram: Hyperion takes the snapshots. As we currently stand we will have them at the start of deposition but I think it's reasonable to assume we might not always

src/hyperion/external_interaction/ispyb/store_datacollection_in_ispyb.py
DominicOram: Yh, it will but by asserting it's not `None` here we also can remove some of the `type: ignore` later

src/hyperion/external_interaction/callbacks/xray_centre/zocalo_callback.py
```diff
@@ -14,7 +14,7 @@
     GridscanISPyBCallback,
 )
 from hyperion.external_interaction.exceptions import ISPyBDepositionNotMade
-from hyperion.external_interaction.ispyb.store_datacollection_in_ispyb import IspybIds
+from hyperion.external_interaction.ispyb.ispyb_store import IspybIds
```
dperl-dls, line 17: nit: I think we could export some of these from the ispyb module

tests/system_tests/external_interaction/conftest.py
```diff
@@ -10,8 +10,10 @@
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 
-from hyperion.external_interaction.ispyb.store_datacollection_in_ispyb import (
+from hyperion.external_interaction.ispyb.gridscan_ispyb_store_2d import (
```
dperl-dls, line 13: as above re: import/export

dperl-dls (approved):
Although some of this is kind of in a weird state, this doesn't really make anything worse than it already is, so I am willing to approve it as-is considering you are already starting on the subsequent refactor, you could address my comments here or there.

PR #1140, opened 2024-02-08
Use proper beamline name in nexus

### LINKED ISSUE #1139: Write the beamline name properly in the nexus file
Using hyperion we write `"DIAMOND BEAMLINE BL03I"` into the `/entry/instrument/name` in the Nexus file. We should instead write `"DIAMOND BEAMLINE I03"` as GDA does.

DominicOram:
Fixes #1139

### To test:
1. Confirm new test fails on main and passes here
2. Confirm there are no type errors in `test_write_nexus.py` and functionality unchanged other than new test


noemifrisina (approved):
Thanks

PR #1145, opened 2024-02-13
1114 Further ispyb store refactoring

rtuck99:
Further adventures  in ispyb store refactoring.

A summary of changes:
* Removal of member variables, in favour of passing them as parameters to methods
* Coalesce the various variables into `DataCollectionInfo` which captures data destined for `DataCollection` rows, and `GridScanInfo` which captures the y/z-axis specific information for abstracting 2D+3D grid scans
* Remove the `_mutate_data_collection_params_for_experiment` and `__prepare_second_scan_params` methods and instead have `populate_xy_collection_info` for both 2+3D gridscan, and `populate_xz_collection_info` for 3D gridscan
* Methods now return IspybIds instead of a random tuple of numbers
* `experiment_type` is now an immutable property

The classes still have rowid state but I feel we can probably remove this and push into callbacks
The params variables are still internal state potentially mutated by external magic but this is for future refactoring


dperl-dls (approved):
LGTM!

PR #1149, opened 2024-02-14
Fix failing CI tests due to DAQ config files

### LINKED ISSUE #1148: Fix failing tests in CI due to missing DAQ config files.
Tests are failing in CI when instantiating GDABeamlineParameters as they rely on the DAQ configuration file which live on the diamond file system. We could use the mock configs from dodal.

noemifrisina:
Fixes #1148 

### To test:
1. Run tests
2. Confirm tests pass in CI


dperl-dls (approved):
Thanks!

PR #1151, opened 2024-02-14
Make ophyd pin tip detection more robust to false negatives

### LINKED ISSUE #1069: Ophyd pin tip: Trigger and wait for the device to find a pin
As of https://github.com/DiamondLightSource/hyperion/pull/938, if we don't find a pin from the ophyd pin detection we wait a second and try again. Instead we should trigger the pin tip detection and wait for it to find a valid pin with a timeout, this is more like how the AD pin tip detection works.

## Acceptance Criteria
* The ophyd pin tip detection allows triggering
* When it is triggered it will only return when a valid pin is found or a timeout is reached
* The subsequent read will return this valid position

DominicOram:
Fixes #1069

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/336

Notes (feel free to disagree):
* I believe the test removed here is covered in `dodal` by `test_given_find_tip_fails_twice_when_triggered_then_tip_invalid_and_tried_twice` trying to add it here would mean a lot of getting in the guts of `ophyd-async`, which I feel we should be somewhat removed from
* This does mean that we no longer log the edges of the pin when no tip is found. I thought that was worth it for the cleaner code. We will also get even better debugging with https://github.com/DiamondLightSource/dodal/issues/290 at some point anyway

### To test:
1. Confirm tests still pass and code cleaner

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     xarray
     doct
     databroker
-    dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@328330036a23cae8748e84a26eb9b77ad113f8a5
+    dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@bede031f673c3468fe57ede1ef21905bbb74e6d5
     pydantic<2.0 # See https://github.com/DiamondLightSource/hyperion/issues/774
     scipy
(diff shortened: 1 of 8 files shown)


olliesilvester (approved):
Looks good, I agree that the test is covered by the one in dodal. I'm not sure how useful the removed logging was but the code is much more clean now.

Issue #1159, opened 2024-02-16
Investigate why Hyperion tries to connect to devices after already being connected


olliesilvester:
Hyperion raised a timeout error when an ophyd_async device failed to connect in time. After looking at https://graylog2.diamond.ac.uk/messages/logs_8652/cd356670-cc46-11ee-a034-1866dafadc6c, it looks like Hyperion was running its setup plans when it had already been running for a while. This part of the code should only be running once, when Hyperion is first started

DominicOram:
Linked PR stops us doing the reconnect but breaks Hyperion if `skip_startup_connection=True`. A proper fix needs some discussion with `ophyd_async` as ideally they would match the behaviour of `ophyd` or at least let us see if a device is connected then decide not to connect it. If the behaviour in `ophyd_async` does not change we will need to pass around `skip_startup_connection` and use that to decide whether we should be calling the connect or not

PR #1165, opened 2024-02-19
Do not reconnect ophyd_async devices on every plan call

DominicOram:
Mitigates (but does not properly fix, see https://github.com/DiamondLightSource/hyperion/issues/1159#issuecomment-1952669379) #1159 

### To test:
1. Confirm matches what's in production and makes sense


olliesilvester (approved):
Looks good for now. I think checking that devices are still connected is a good feature in the long term though, but we need to wait until we find out what was causing the reconnect issues

PR #1172, opened 2024-02-19
Use the bit depth from the detector to work out what VDS data type to use

### LINKED ISSUE #1162: Rotation scan data VDS is 16-bit
When doing a rotation scan in Hyperion the data appears to be 16-bit (https://ispyb.diamond.ac.uk/dc/visit/mx31353-115/id/12985950) whereas it should actually be 32-bit (https://ispyb.diamond.ac.uk/dc/visit/mx31353-115/id/12985506) (you can easily tell this as the Hyperion data has yellow tile gaps).

## Acceptance Criteria
* Rotation data has correct bit depth

DominicOram:
Fixes #1162

### To test:
1. Confirm new tests pass and correctly check the VDS type


olliesilvester (approved):
Looks like a good solution! I think it would be nice if we did it this way for xray centring as well at some point

EDIT: Just seen you've already mentioned that in linked issue

PR #1177, opened 2024-02-21
1146 the great ispyb interaction refactor - part 3

### LINKED ISSUE #1146: The great ispyb interaction refactor
The way we store data in ispyb is currently very messy and not very extensible. After a discussion between @dperl-dls, @rtuck99 and @DominicOram we decided on a new design with the following principles:
* The `StoreInIspyb` classes should hold no state, it should just put relevant data into the database and forget about it
* `Callbacks` and `Store` classes should be as agnostic as possible to changes in when data is read from devices and what devices are called. If there is an open run then the latest data read from the device should just overwrite the correct field in ispyb
* More data should be moved to being read from ophyd devices and less to the metadata supplied in the start document

The first steps in doing this are to:
1. Move all state out of `StoreInIspyb`
2. Split the parameters of what the callbacks take and what `StoreInIspyb` requires

rtuck99:
Fixes #1146

This is part 3 of the refactoring.
Previous episodes:
#1145 
#1136 

This set of changes:
* Removes all references to `params`, `ispyb_params`, `detector_params` from `ispyb_store` modules
* Introduces `data_model.py` module in `ispyb` package which contains the data model interface to ispyb stores as `dataclass` types
* vastly simplifies `ispyb_store` hierarchy, guts gridscan 2d/3d leaf classes
* moves work of populating the data model into free functions in callback layer, in `ispyb_mapping` modules

Still to come in part 4:
* Removal of all id state from `ispyb_store` classes
* further rationalisation of those classes


DominicOram (changes requested):
Definitely going in a great direction. I think we can simplify a bit see https://github.com/DiamondLightSource/hyperion/pull/1187

PR #1179, opened 2024-02-22
1146 the great ispyb interaction refactor part 4

### LINKED ISSUE #1146: The great ispyb interaction refactor

rtuck99:
The final installment of ispyb refactoring

Fixes #1146

In this PR:
* Removes all management of data collection IDs from the ispyb_store
* All subclasses of `StoreInIspyb` have been removed
* Fixes for S03 tests of ispyb that were previously broken (untested)
* A fix for issue introduced in prior refactorings where visit id wasn't correctly populated
* IspybIds `grid_ids` and `data_collection_ids` are no longer of type `int | tuple[int] | None` and are now always just `tuple[int]`

Previous episodes:
#1177
#1145 
#1136 

### To test:

* Ispyb depositions should work normally as before for rotation, 3d + 2d gridscans
*


DominicOram (changes requested):
Could you fix the tests please? I think they need a dodal update but with the dodal update you get https://github.com/DiamondLightSource/dodal/pull/346/files, which I think breaks a few others.

PR #1180, opened 2024-02-22
Panda grid scans: Set up position capture

### LINKED ISSUE #1178: Create Panda configuration for capturing data
For diagnostics and eventually for data analysis, we need to have the option to use the panda's position capture (PCAP) within Hyperion's grid scans. Note that until https://github.com/bluesky/ophyd-async/pull/90 is merged, we can only capture data manually.

### Acceptance Criteria

- A separate panda config yaml file is created in `/dls_sw/i03/software/daq_configuration/panda_configs`, which wires up the PCAP so that: Its trigger and gate inputs are wired to the CLOCK1 output, and the trigger is set to the rising edge. This will allow us to get the MIN/MAX/MEAN values of the X value for each trigger point. 
- This file should specify that we capture MIN/MAX/MEAN of the INENC1, the time at which the PCAP is triggered, as well as TTLIN1 (This is the signal sent by PPMAC when we start a new row)
- If the PCAP config is selected, Hyperion should enable and arm the PCAP block during panda setup, and disarm and disable it during tidy-up. Config choice can be hardcoded for now since we can only capture data manually

olliesilvester:
Fixes #1178

### To test:
1. Check changes in hyperion arm, disarm, and trigger the PCAP block at the correct points
2. Confirm that the pcap.gate and pcap.trigger settings in https://gitlab.diamond.ac.uk/mx/i03-config/-/blob/main/panda_configs/flyscan_with_pcap.yaml?ref_type=heads are consistent with what the linked issue describes, and confirm that these are sensible

### DIFF
--- a/src/hyperion/device_setup_plans/setup_panda.py
+++ b/src/hyperion/device_setup_plans/setup_panda.py
@@ -24,6 +24,11 @@ class Enabled(Enum):
     DISABLED = "ZERO"
 
 
+class PcapArm(Enum):
+    ARMED = "Arm"
+    DISARMED = "Disarm"
+
+
 def get_seq_table(
     parameters: PandAGridScanParams,
     exposure_distance_mm,
@@ -154,6 +159,9 @@ def setup_panda_for_flyscan(
 
     yield from bps.abs_set(panda.seq[1].table, table, group="panda-config")
 
+    # Wait here since we need PCAP to be enabled before armed
+    yield from bps.abs_set(panda.pcap.enable, Enabled.ENABLED.value, wait=True)  # type: ignore
+
     yield from arm_panda_for_gridscan(panda, group="panda-config")
 
     yield from bps.wait(group="panda-config", timeout=GENERAL_TIMEOUT)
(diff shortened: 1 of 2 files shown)


DominicOram (changes requested):
Looks good. As I aske on slack, would be good if I could open the new `yaml` to confirm it. If you can document how to do that or point to the documentation for it that would be great. Additionally:
* Tests are failing because this is pointing at an old `dodal` - can you bring them forward and see if that fixes it?
* Do we need to update `PANDA_SETUP_PATH` with this PR?

olliesilvester:
While we can only save `PCAP` data manually, I think leave `PANDA_SETUP_PATH` as is, then change it manually upon testing. Doesn't really matter, but otherwise we'd be arming and triggering PCAP loads without the data going anywhere, which seems weird

olliesilvester:
Yep, the PCAP PV's exist in both setups, but the PV's are set according to:
```
pcap.arm: Disarm
pcap.bits0_capture: 'No'
pcap.bits1_capture: 'No'
pcap.bits2_capture: 'No'
pcap.bits3_capture: 'No'
pcap.enable: ZERO
```

PR #1183, opened 2024-02-23
Allow the ispyb experiment_type field to be populated by GDA

### LINKED ISSUE #1182: Rotation scans should put correct experiment kind into ispyb
The experiment kind in ispyb is used by analysis to work out what to do with the data. Currently all rotation experiments go in as SAD but actually some may be of a different type. This type is known by agamemnon and GDA so needs passing through to Hyperion.

## Acceptance Criteria
* Experiment kind for rotations is passed through from GDA to hyperion and put in the correct ispyb field

rtuck99:
Fixes #1182

### To test:
* ispyb experiment_type field should now be populated via the `ispyb_params.ispyb_experiment_type` field as specified by the call from GDA (for rotation scans)
* For gridscans it will be left as "mesh" or "Mesh3D" as appropriate, and if unspecified for rotation scans it will be "SAD"

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     xarray
     doct
     databroker
-    dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@8d25746e3407c8331357e8ce159e89c5d3bfee92
+    dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@00ef53ddf55c53fd000990c25ad6093e88f39fe4
     pydantic<2.0 # See https://github.com/DiamondLightSource/hyperion/issues/774
     scipy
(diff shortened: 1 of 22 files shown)


DominicOram (approved):
Thans @rtuck99, this seems good to me. If I was being nitpicky I think `kind` is maybe used more widely for this than `type` but not too fussed.

rtuck99:
> Thans @rtuck99, this seems good to me. If I was being nitpicky I think `kind` is maybe used more widely for this than `type` but not too fussed.

I'm just going by the nomenclature I see in IspyB and in GDA - I haven't seen anywhere in the code that references _kind

DominicOram:
> I'm just going by the nomenclature I see in IspyB and in GDA - I haven't seen anywhere in the code that references _kind

Fair enough, it looked like GDA had `kind` where i saw it. But I'm happy with type

Issue #1091, opened 2024-01-22
Send zocalo information on the acquisition id and number of images


DominicOram:
For real time analysis the GPU will need to know the sequence ID of the next collection (which we can read from `BL03I-EA-EIGER-01:CAM:SequenceId`) and the range of images over which the collection is being taken (e.g. the size of the 2D grid and the subsequent 2D grid). This should be sent as parameters to zocalo on `ZocaloTrigger.run_start`.

DominicOram:
@ndevenish, anything I've missed here? Do we need to use a different recipe too?

DominicOram:
After discussion with @ndevenish, it might not be sequenceID they want. Hold off on doing anythign on this until we have a better idea of what's needed

DominicOram:
They instead need acquisition ID (which should just be filename).

PR #1185, opened 2024-02-26
Refactor rotation and grids to use the same Zocalo callback

DominicOram:
Lays some groundwork useful for #1091

To do this I have:
* Moved the kickoff and complete gridscan parts into a common bit of code between panda and FGS
* Moved zocalo triggering to the inner rotation plan, this was needed so that we still made sure that we triggered zocalo after ispyb but wanting to do it on the start document rather than the end (like grids)
* Provide the `zocalo_env` directly in the start metadata, means the zocalo callback doesn't need to know about wider parameters classes
* Moved the calculation of processing time into the ispyb callback rather than having this weird interchange of info between the zocalo and ispyb callbacks, that isn't needed in rotations
* Store the data collections as a tuple always, now its just off length 0 when no storage has been done and 1 for the rotations

### To test:
1. Confirm no major behavior change, other than listed above
2. Confirm tests barely changed and still pass


src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -163,6 +165,48 @@ def tidy_up_plans(fgs_composite: FlyScanXRayCentreComposite):
...
+        }
+    )
+    @bpp.contingency_decorator(
+        except_plan=lambda e: (yield from bps.stop(eiger)),
+        else_plan=lambda: (yield from bps.unstage(eiger)),
+    )
+    def do_fgs():
+        # Check topup gate
+        expected_images = yield from bps.rd(gridscan.expected_images)
+        exposure_sec_per_image = yield from bps.rd(eiger.cam.acquire_time)
```
olliesilvester, line 189: It would be nice to have the `acquire_time` signal in dodal be renamed `acquire_time_s`

dperl-dls (approved):
This is really nice, thanks. I think we can further eliminate the need for two separate callback instances by using `emit` from the ispyb callbacks and appending the IDs to the doc - then we can both clean up the metadata in the plan as well as use it to tell zocalo when to trigger - I'll make a new issue for it

DominicOram:
> I think we can further eliminate the need for two separate callback instances by using emit from the ispyb callbacks and appending the IDs to the doc

I was looking for a way to do this but couldn't easily find out how. Will be great if we can.

src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
DominicOram: Agree but I think cleaner to do in it's own PR. Have made https://github.com/DiamondLightSource/hyperion/issues/1188

PR #1187, opened 2024-02-27
The great ispyb interaction refactor doms suggestion

DominicOram:
Moves more info into `GridScanInfo` and so simplifies a lot of other code, the new `GridScanInfo` is also a realistic packet of data that we could put into a Blueksy event so prepares us for that. Tests are broken in the same way as they are in base PR.

### DIFF
--- a/tests/unit_tests/external_interaction/ispyb/test_gridscan_ispyb_store_2d.py
+++ b/tests/unit_tests/external_interaction/ispyb/test_gridscan_ispyb_store_2d.py
@@ -11,7 +11,7 @@ from hyperion.external_interaction.callbacks.common.ispyb_mapping import (
 )
 from hyperion.external_interaction.callbacks.xray_centre.ispyb_mapping import (
     construct_comment_for_gridscan,
-    populate_xy_data_collection_info,
+    populate_data_collection_info,
 )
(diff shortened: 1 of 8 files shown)


tests/unit_tests/external_interaction/ispyb/test_gridscan_ispyb_store_2d.py
```diff
@@ -5,13 +5,13 @@
 from ispyb.sp.mxacquisition import MXAcquisition
```
rtuck99, line 5: Actually I think one of the things that would be a fair comment on the refactoring is that I think these test modules should be rewritten so that rather than construct the inputs by essentially duplicating the logic of the callback layer (since the callback layer tests are effectively copied from here and end-to-end to the ispyb layer), the fixtures should just directly populate the raw *Info dataclasses that are passed in.

DominicOram: Sure, as long as we still have the callback tests covering this then that sounds great

rtuck99 (approved):
Good changes I think this considerably simplifies a lot of the mapping logic

PR #1190, opened 2024-02-28
Correctly set run up distance

### LINKED ISSUE #1189: Panda grid scans: Correctly set run-up distance
Hyperion's entry points from GDA's full data collections do not accept run-up distance as a parameter. This means the run-up distance has been fixed to its default value, which seems to not be enough for fast grid scans at 435hz or greater. We should properly add run-up distance to other parameters.

### Acceptance Criteria

- `run_up_distance` added as optional parameters to all appropriate entry points. I believe these are `grid_detect_then_xray_centre`, `pin_tip_centre_then_xray_centre`, and `wait_for_robot_load_then_centre`

olliesilvester:
Fixes #1189 

Link to dodal PR (if required): [#348](https://github.com/DiamondLightSource/dodal/pull/348)

### To test:
1. Check changes are sensible
2. Tests pass


DominicOram (approved):
Great, thanks!

PR #1198, opened 2024-02-29
Pass the ISPyB DCIDs from the ISPyB callback to the Zocalo one using the emitted document

### LINKED ISSUE #1192: Pass Zocalo callback ISPyB IDs through documents
Following https://github.com/DiamondLightSource/hyperion/pull/1185 it would be nice to only have to create a single instance of the `ZocaloCallback`. In order to achieve this, instead of initialising the zocalo callback with an ispyb callback instance, the zocalo callback should be given to the `emit` argument of the ispyb callbacks, and not subscribed to the plan independently. the IDs can then be appended to the document of the method where they are generated, and the zocalo callback will recieve them. the point of where to trigger zocalo can then also be included in the plan metadata

dperl-dls:
Fixes #1192 
Tidies up the implementation of emit in the `PlanReactiveCallback` and uses it to forward documents from the ISPyB callback to the Zocalo callback, after appending the DCIDS to it.

### To test:
1. Run tests


src/hyperion/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -115,7 +120,7 @@ def activity_gated_event(self, doc: Event):
             )
 
         set_dcgid_tag(self.ispyb_ids.data_collection_group_id)
-        return doc
+        return self._tag_doc(doc)
```
DominicOram, line 123: Should: We've already called `super().activity_gated_event(doc)` here, so we already have the tag in there?

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -75,6 +79,7 @@ def activity_gated_start(self, doc: RunStart):
                 else Store2DGridscanInIspyb(self.ispyb_config, self.params)
             )
             self.ispyb_ids = self.ispyb.begin_deposition()
+        return self._tag_doc(doc)
```
DominicOram, line 82: Could: Feels like we can just call the parent `start` here and let that do the tagging?

src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
```diff
@@ -38,8 +38,12 @@ class RotationISPyBCallback(BaseISPyBCallback):
     Usually used as part of a RotationCallbackCollection.
     """
 
-    def __init__(self) -> None:
-        super().__init__()
+    def __init__(
```
DominicOram, line 41: Should: As with the centring callback I think we can call the base methods and let them do the tagging

DominicOram (changes requested):
Some small comments in the code. I generally like the idea of this but some general comments that I think we could discuss:
* Would it be cleaner to follow a pattern more like the [LiveDispatcher](https://nsls-ii.github.io/bluesky/callbacks.html?highlight=livedispatcher#bluesky.callbacks.stream.LiveDispatcher), where we subscribe the zocalo callback to the ispyb one?
* Do we even need a separate run to trigger zocalo on. If the only reason we needed to do that was that we needed a valid ispyb deposition before we could trigger zocalo then just the fact we have an ispyb_id in the doc is enough for zocalo to be triggered? In fact it feels like this might mean we can entirely remove the inner runs and simplify the plans somewhat?
* A bit nit-picky but the original issue says that we would now be able to use just one zocalo callback. We're not, we're still using two instances. I think that's fine but worth mentioning

dperl-dls:
> Do we even need a separate run to trigger zocalo on. If the only reason we needed to do that was that we needed a valid ispyb deposition before we could trigger zocalo then just the fact we have an ispyb_id in the doc is enough for zocalo to be triggered? In fact it feels like this might mean we can entirely remove the inner runs and simplify the plans somewhat?

I think so, but can we investigate this in a different issue? This one is already getting kind of big, but I think it allows us to tidy up in a few different places, like #1199

src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
dperl-dls: good point, except for the `stop` method which we can only use once, I've made it so that we only use the base class for tagging

DominicOram:
>Maybe, yes. Do you mean that we change the ISPyB callback to be a LiveDispatcher? It seems like a big job but it looks like it should work. The way I envisaged this heading is that, if I've understood correctly eventually there will be some message bus for documents which we should use instead anyway, to which everything can subscribe and emit back if needed, and this is preparatory in that at least all the callbacks are only getting and putting information into documents.

Yes, I think the ISPyB it could be a `LiveDispatcher`. Though it looks like `LiveDispatcher` only processes events, not start/stop docs so it may have to be something that just looks like a `LiveDispatcher`. Yes, I think we're envisaging the same thing. It just seemed a bit cleaner to me that rather than having to pass zocalo in as an `emit`, which seems like a special case, you think of it as the ispyb callback creating a new event stream that has ids on it, and you can later subscribe to it.

>No, in src/hyperion/external_interaction/callbacks/__main__.py...

My bad, sorry.

> I think so, but can we investigate this in a different issue?

Sure, https://github.com/DiamondLightSource/hyperion/issues/1204

dperl-dls:
I think it would only really be cleaner if we had more things to use it with than just this one case, which is the only one in the whole application. It's a lot of extra boilerplate to keep a list of subscriptions, manage dispatching to them, etc.

DominicOram:
Not if we push the boilerplate into `bluesky`. But I think you've convinced me that it's overkill for now, worth keeping in the back of our minds as a possibility though. I will re-review this PR this afternoon

DominicOram (approved):
Looks good, thanks. I've added one place where I think you missed calling the parent.

PR #1206, opened 2024-03-01
Move rotation scan calculations to one function

### LINKED ISSUE #853: Refactor rotation scan maths
Currently the rotation scan plan does:
1. Work out the offsets
2. Set up the zebra
3. Work out and send start position
4. Work out and send end position

This means the maths of how to do the scan is a bit spread out. It would be more readable and more testable if we had a function to call at the start of the plan to work out all the motion profile then the subsequent plan could be more simple.

dperl-dls:
Fixes #853
Does all the calculation for the rotation scan motion in one place

What little system tests there were became redundant - issue for this in https://github.com/DiamondLightSource/hyperion/issues/1207

### To test:
1. Run tests


noemifrisina (approved):
lgtm

PR #1209, opened 2024-03-01
Add an endpoint to write `DEBUG` level logs

### LINKED ISSUE #1201: Provide endpoint to flush debug logs
As of https://github.com/DiamondLightSource/dodal/pull/352 we now store the debug logs in a circular buffer that gets flushed to disk on error. However, as mentioned in https://github.com/DiamondLightSource/hyperion/issues/1191, there may be cases where we want to manually flush the buffer to disk e.g. if Hyperion hangs. We should make an endpoint for this

## Acceptance Criteria
* I can write to an endpoint that will write the contents of the debug log buffer to disk
* This works when Hyperion has hung
* This is documented

### LINKED ISSUE #1208: Readme is out of date and wrong
It still refers to various options which are not available such as `logging-level` and the wording can be improved.

dperl-dls:
Fixes #1201, fixes #1208

### To test:
1. Run dev hyperion and execute: `curl -X PUT http://127.0.0.1:5005/flush_debug_log` and see that this works
2. Check readme changes make sense

### DIFF
--- a/src/hyperion/log.py
+++ b/src/hyperion/log.py
@@ -5,6 +5,8 @@ from typing import Optional
 
 from dodal.log import (
     ERROR_LOG_BUFFER_LINES,
+    CircularMemoryHandler,
+    DodalLogHandlers,
     integrate_bluesky_and_ophyd_logging,
     set_up_all_logging_handlers,
 )
@@ -13,6 +15,7 @@ from dodal.log import LOGGER as dodal_logger
 LOGGER = logging.getLogger("Hyperion")
 LOGGER.setLevel("DEBUG")
 LOGGER.parent = dodal_logger
+__logger_handlers: Optional[DodalLogHandlers] = None
 
 ISPYB_LOGGER = logging.getLogger("Hyperion ISPyB and Zocalo callbacks")
 ISPYB_LOGGER.setLevel(logging.DEBUG)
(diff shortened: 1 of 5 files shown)


src/hyperion/log.py
```diff
@@ -52,6 +55,16 @@ def do_default_logging_setup(dev_mode=False):
     integrate_bluesky_and_ophyd_logging(dodal_logger, handlers)
     handlers["graylog_handler"].addFilter(dc_group_id_filter)
 
+    global __logger_handlers
+    __logger_handlers = handlers
+
+
+def get_memory_handler() -> CircularMemoryHandler:
```
DominicOram, line 62: Nit: I think `get_debug_handler` would be slightly clearer as it's logging the debug not the memory

DominicOram (approved):
Thanks, works as expected some minor things, take them or leave them.

Could: If I delete the `hyperion.log` debug file then call the flush it doesn't make a new one, I think this is an edge case though so up to you if you want to address it.

Could: Ideally I think exposing the handlers out of `log.py` is a bit messy and would rather just expose a `flush_debug` function but I like that we return the filename in the status so maybe `flush_debug` should return the filename it flushes too? Not sure, up to you if you think it's cleaner or not

PR #1210, opened 2024-03-01
Tag log messages with the UID of the current outer run

### LINKED ISSUE #1132: External callbacks break DCGID log filter
The DCGID log filter will only be applied to messages from the callback process, as this is where the deposition is created. It should be communicated back to the main process somehow. One candidate is the zocalo device, but that currently doesn't have access to the DC(G)ID until the zocalo results come in. Perhaps we could update it earlier, or find another way, or live with only tagging the messages about the results.

dperl-dls:
Fixes #1132

Adds a callback to set a tag on log messages for the current run based on a metadata field, using the bluesky UID for the outer run of each experiment

### To test:
1. Run tests


noemifrisina (approved):
Looks okay, tests pass.

DominicOram (commented):
Do we really need `SET_LOG_UID_TAG`? Can't we in the callback log with the outermost id?
```python
class LogUidTaggingCallback(CallbackBase):
    def __init__(self) -> None:
        self.run_uid = None

    def start(self, doc: RunStart):
        if self.run_uid is None:
            self.run_uid = doc.get("uid")
            run_uid_filter.run_uid = self.run_uid

    def stop(self, doc: RunStop):
        if doc.get("run_start") == self.run_uid:
            self.run_uid = None
            run_uid_filter.run_uid = None
```
I don't like that the plans now have to have knowledge of what the callbacks are doing re logging

dperl-dls:
@DominicOram Doh, of course. New PR here: https://github.com/DiamondLightSource/hyperion/pull/1233

DominicOram:
Thank you!

PR #1215, opened 2024-03-05
Speed up the unit tests

rtuck99:
Unit tests were being slowed down, mainly by a series of exceptions thrown after completion in `test_optimise_attenuation_plan`, caused by a Status timeout. This fixes the issue by initialising the fake PV to the expected value, causing the dodal device set not to time out.

It is still unclear what is the primary cause of the delay, but there seems to be a ~0.5s penalty for every exception reported on the RunEngine, and several other cases where exceptions are thrown after a test completes successfully, however none as significant as this.

On my machine, the following speedup was obtained:
Before fix, tests tooks
66.19, 65.47, 65.60 seconds

After fix:
29.38, 29.61, 29.60 seconds

The other fix here is to ensure that the RunEngine event loop is properly shut down after use. Unfortunately this doesn't seem to have any significant impact on performance.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import sys
+import threading
 from functools import partial
 from os import environ, getenv
 from typing import Callable, Generator, Optional, Sequence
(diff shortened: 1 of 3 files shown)


dperl-dls (approved):
Looks good, thanks!

PR #1221, opened 2024-03-05
Provide zocalo with more info for GPU analysis

### LINKED ISSUE #1091: Send zocalo information on the acquisition id and number of images

DominicOram:
Fixes #1091

Link to dodal PR: https://github.com/DiamondLightSource/dodal/pull/359

### To test:
1. Confirm tests pass

### DIFF
--- a/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
+++ b/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import argparse
 import dataclasses
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, List
 
 import bluesky.plan_stubs as bps
 import bluesky.preprocessors as bpp
@@ -72,6 +72,7 @@ if TYPE_CHECKING:
     )
 
     PandaOrZebraGridscan = FastGridScan | PandAFastGridScan
+    from scanspec.core import AxesPoints, Axis
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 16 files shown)


dperl-dls (commented):
~~Could you update this to match the dodal branch? Tests are broken since merging main on dodal~~ Never mind, was looking at the wrong branch

dperl-dls (approved):
Looks good, thanks!

DominicOram:
Merge was non-trivial, could you re-review @dperl-dls?

src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -177,13 +179,15 @@ def kickoff_and_complete_gridscan(
...
     zocalo_environment: str,
+    scan_points: List[AxesPoints[Axis]],
 ):
     @TRACER.start_as_current_span(DO_FGS)
     @bpp.set_run_key_decorator(DO_FGS)
     @bpp.run_decorator(
         md={
             "subplan_name": DO_FGS,
             "zocalo_environment": zocalo_environment,
+            "scan_points": scan_points,
```
dperl-dls, line 190: On second thought I actually don't like this in the plan, I think we should do it in the callbacks instead, but I'm happy to leave it as is because I think it fits nicely in https://github.com/DiamondLightSource/hyperion/issues/1217

dperl-dls (approved):
lgtm

src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
DominicOram: Do what in the callbacks? Create the points out of the parameters?

dperl-dls: yeah, or out of the grid scan info, whichever

PR #1229, opened 2024-03-06
1228 tidy up aperture reading

### LINKED ISSUE #1228: Tidy up aperture reading code
Following https://github.com/DiamondLightSource/hyperion/issues/1032 we could do with some tidying up.

## Must
* Need to update GDA to pick up the human readable aperture names now
* Reading the `aperture_device` won't work if we haven't set it in Hyperion yet, we should instead read it based on the current state of the motors

## Should
* We should instead set using a `SingleAperturePosition` rather than the underlying motor positions as that's a cleaner interface
* The tolerance of being in position should be read from the EPICS motors

## Could
* I'm not sure `ApertureFiveDimensionalLocation` gives us much over just using the `aperture_x` etc. inside `SingleAperturePosition`

DominicOram:
Fixes #1228

Link to dodal PR: https://github.com/DiamondLightSource/dodal/pull/363

Note: I added GDA name in Hyperion rather than convert in GDA as it's simpler and the code will get removed when we get full grid+rotation in anyway

### To test:
1. Confirm tests pass


dperl-dls (approved):
LGTM, thanks

PR #1232, opened 2024-03-07
Organise util scripts

### LINKED ISSUE #1231: utility scripts are cluttered

dperl-dls:
Fixes #1231 

The wiki will need to be updated to reference the new location of the deploy script.

### To test:
1. Look at how hyperion takes up four less lines in the vscode explorer, and you can see more of dodal
2. Run the `dls_dev_env` script in and out of the hyperion dir (in a new deployment if you want)
3. Test that the deploy script still works (except it won't find the `dls_dev_env`, since it's in a different place now - confirm the location the deploy script looks will be correct)


utility_scripts/deploy/deploy_hyperion.py
```diff
@@ -128,13 +131,17 @@ def get_hyperion_release_dir_from_args(repo: repo) -> str:
 Only do so if you have informed the beamline scientist and you're sure Hyperion is not running.
 """
     )
-    # Creates symlink: software/bluesky/hyperion_version -> software/bluesky/hyperion
+    # Creates symlinks: software/bluesky/hyperion_latest -> software/bluesky/hyperion_{version}/hyperion
+    #                   software/bluesky/hyperion -> software/bluesky/hyperion
     if move_symlink == "y":
```
DominicOram, line 141: Should: I wonder if we should have another step here that says something like "Previous live version was X, do you want to set this is the stable version?" Then move it to `hyperion_stable`?

DominicOram (changes requested):
Thanks, I tidied up a failing relative path thing. Couple of other comments in code. Also, can you update https://github.com/DiamondLightSource/hyperion/wiki/Creating-a-release-and-deploying-it?

dperl-dls:
For testing you will need to change line 122 to
`            hyperion_repo.deploy_location, "dls_dev_env.sh"`
because this path changes between this branch and main

DominicOram (approved):
Great thank you!

DominicOram:
I've updated the wiki

PR #1233, opened 2024-03-07
#1132 Remove need for metadata tag for logs

dperl-dls:
### To test:
1. See that `test_messages_are_tagged_with_run_uid` still passes even without the plan metadata


DominicOram (changes requested):
Thanks, this looks cleaner. But whilst we're here:
* What is `create_and_set_new` doing on the `RunUIDFilter`? I don't think it's used. 
* Could: I wonder whether we can combine the `DCGIDFilter` and `RunUIDFilter` classes into some kind of `AppendMetaDataFilter` which takes the name of the parameter to append as on `__init__` but then still have separate instances for `run_uid` and `dc_group`?

DominicOram (approved):
Great, thanks!

PR #1237, opened 2024-03-08
Use VMAX for resetting omega before and after rotation scans

### LINKED ISSUE #685: Get max velo for moving to start of rotation
We currently do `yield from bps.abs_set(axis.velocity, 120, wait=True)` to go to the start of the rotation scan, we should be able to just get the `.VMAX` of the motor and use that. This has the advantage of being slightly more readable and will update if the motor ever becomes faster/slower.

Note that `EpicsMotor` doesn't have `VMAX` at the moment so we will need to add it in

dperl-dls:
Fixes #685
Uses the `max_velocity` from the `ExtendedEpicsMotor` to reset velocity for rotation scans

Link to dodal PR (if required): [367](https://github.com/DiamondLightSource/dodal/pull/367)

### To test:
1. Run tests


src/hyperion/experiment_plans/rotation_scan_plan.py
```diff
@@ -268,11 +266,23 @@ def rotation_scan(composite: RotationScanComposite, parameters: Any) -> MsgGener
     def rotation_scan_plan_with_stage_and_cleanup(
         params: RotationInternalParameters,
     ):
+        motor_time_to_speed = yield from bps.rd(composite.smargon.omega.acceleration)
+        max_vel = (
+            yield from bps.rd(composite.smargon.omega.max_velocity)
+            or DEFAULT_MAX_VELOCITY
```
DominicOram, line 272: Should: Why do we have a default in here? In case the PV doesn't exist - we have bigger problems. For the unit tests - we should just be setting up the mock device properly

DominicOram (changes requested):
Looks good, one minor query

src/hyperion/experiment_plans/rotation_scan_plan.py
DominicOram: I think that's very unlikely but I'm not that bothered

DominicOram (approved):
Great, ta!

Issue #1227, opened 2024-03-06
Grids sometimes to the edge of image


dperl-dls:
e.g. https://ispyb.diamond.ac.uk/dc/visit/in29352-55/dcg/11425683

DominicOram:
We should look back and count how many times the grid was large recently (maybe just do a histogram of grid heights?). We should also then log the edges, either through logging the array or taking an image.

dperl-dls:
![Figure_1](https://github.com/DiamondLightSource/hyperion/assets/115003895/48bf0a00-891b-4a71-a45e-0a5c1da72f95)
Extracted y_steps from the last two days of logs

DominicOram:
Looks like there are -1's in the edge detection for both top/bottom https://graylog2.diamond.ac.uk/messages/logs_8684/a440f231-dd61-11ee-a034-1866dafadc6c. Maybe filter these out rather than the 0s?

PR #1238, opened 2024-03-08
#1227 add edge logging

dperl-dls:
Adds some logs to help us diagnose edge detection issues


DominicOram (approved):
Thanks, this does make the logs a bit dense. I wonder if we want to remove one of the sets of logs in the long run but this is good for now.

PR #1256, opened 2024-03-13
Speed up unit tests and fix some typing issues

### LINKED ISSUE #1220: Investigate slowness of unit tests
Following on from #1215 
There are still a number of unit tests that seem to be slowing down unit test execution. This seems to be related to exceptions that are raised during successful execution of a test, and these seem to cause a delay. The reason why these exceptions cause the delay is unknown. 

e.g. in `test_given_no_tip_found_ever_when_get_tip_into_view_then_smargon_moved_positive_and_exception_thrown()` a number of exceptions similar to the following are thrown and they each cause ~0.7s of delay:
```
[2024-03-05 13:50:21,235] ophyd status ERROR: An error was raised on a background thread while running the callback <bound method PinTipDetect.log_tips_and_statistics of FakePinTipDetect(prefix='BL03S-DI-OAV-01:MXSC:', name='oav_mxsc_pin_tip', parent='oav_mxsc', read_attrs=['tip_x', 'tip_y', 'triggered_tip', 'validity_timeout', 'settle_time_s'], configuration_attrs=[])>(StableSubscriptionStatus(device=oav_mxsc_pin_tip_tip_x, done=True, success=True)).
Traceback (most recent call last):
  File "/scratch/ws/hyperion/.venv/lib/python3.10/site-packages/ophyd/status.py", line 269, in _run_callbacks
    cb(self)
  File "/scratch/ws/dodal/src/dodal/devices/areadetector/plugins/MXSC.py", line 56, in log_tips_and_statistics
    median, standard_deviation = statistics_of_positions(self.tip_positions)
  File "/scratch/ws/dodal/src/dodal/devices/areadetector/plugins/MXSC.py", line 27, in statistics_of_positions
    x_coords, y_coords = np.array(positions).T
ValueError: not enough values to unpack (expected 2, got 0)
[2024-03-05 13:50:21,989] Hyperion pin_tip_centring_plan INFO: Pin tip found at (-1, -1)
```

It would be good if we could speed up the slow tests by eliminating the exceptions, ideally we would also want to:
* Making it easier to identify when this is happening to prevent further occurrences 
* Figure out why this slows down the tests and remove the slowdown.

In investigating the latter, one possible avenue may be to temporarily re-instate the exceptions in the optimise_attenuation test and run the profiler against it (may need to tweak the timeout so that the exceptions happen before the test completes), possibly that may make the cause more obvious.

DominicOram:
Fixes #1220

* Removes logging by default, as discussed on issue
* Also speeds up a few other tests

### To test:
1. Run the tests with `pytest -m "not s03"`, confirm they all pass in ~30s
2. Confirm the tests in CI are still getting all the logs written to terminal

### DIFF
--- a/tests/unit_tests/external_interaction/test_write_rotation_nexus.py
+++ b/tests/unit_tests/external_interaction/test_write_rotation_nexus.py
@@ -75,6 +75,8 @@ def test_rotation_scan_nexus_output_compared_to_existing_file(
     nexus_filename = f"{tmpdir}/{TEST_FILENAME}_{run_number}.nxs"
     master_filename = f"{tmpdir}/{TEST_FILENAME}_{run_number}_master.h5"
 
+    eiger.bit_depth.sim_put(32)  # type: ignore
+
     RE = RunEngine({})
 
     with patch(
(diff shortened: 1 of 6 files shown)


tests/unit_tests/external_interaction/test_write_rotation_nexus.py
dperl-dls, line 1: This feels unrelated?

dperl-dls: But it's fine, I just updated the title to match

dperl-dls (approved):
Nice, thanks!

PR #1258, opened 2024-03-14
Remove callback collections

### LINKED ISSUE #1199: Remove `CallbackCollection`s
Following https://github.com/DiamondLightSource/hyperion/issues/1192, the `CallbackCollection` classes are but withered husks of their former selves, and can be uprooted and discarded.

dperl-dls:
Fixes #1199

Replaces callback collections with some simple util functions

### To test:
1. Check tests haven't broken


src/hyperion/__main__.py
```diff
@@ -45,7 +46,7 @@ class Command:
     devices: Optional[Any] = None
     experiment: Optional[Callable[[Any, Any], MsgGenerator]] = None
     parameters: Optional[InternalParameters] = None
-    callbacks: Optional[type[AbstractPlanCallbackCollection]] = None
+    callbacks: Optional[Callable[[], Tuple[CallbackBase, CallbackBase]]] = None
```
DominicOram, line 49: Could: This type is getting a bit unwieldy, maybe create do a:
```python
CallbackFactory = Callable[[], Tuple[CallbackBase, CallbackBase]]
```
somewhere and use that?

DominicOram (changes requested):
Great, thanks. In general I'm not a fan of indexing into tuples if we can avoid it as I think it makes things less readable and harder to deal with if the tuple order changes, we're doing this quite a lot in tests but I think we could instead do `nexus_callback, ispyb_callback = get_callbacks()` and improve readability. I've pointed out a couple of these places but if you could go through and do it throughout the tests that will be great

DominicOram (approved):
Great, thank you!

PR #1261, opened 2024-03-14
Provide index to zocalo for GPU

### LINKED ISSUE #1260: When triggering zocalo give an index
We have a race condition where when we trigger zocalo with the two grids the GPU can get the 2nd request before the first 1. We don't want to have to wait for all the data taken for the first grid to start on the second. To fix this we can provide an index to say this is 0th or 1st so that the GPU can wait for the first message to start analysis.

## Acceptance Criteria
* An index is sent with every zocalo start

DominicOram:
Fixes #1260

Link to dodal PR: https://github.com/DiamondLightSource/dodal/pull/383

### To test:
1. Confirm tests still pass

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     xarray
     doct
     databroker
-    dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git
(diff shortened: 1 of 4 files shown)


dperl-dls (approved):
lgtm!

PR #1262, opened 2024-03-15
Use event data for nexus writing

### LINKED ISSUE #1173: Use more event data in nexus writing
Following https://github.com/DiamondLightSource/hyperion/pull/1172 we should refactor the nexus writing to use more of the event data from the plan. Specifically:

* The two nexus callbacks should have a common parent
* The XRC callback should also use the bit depth to work out datatype
* Both should use the event data to replace any references to `ispy_params` in the writer

DominicOram:
Fixes #1173

Removes any dependency that Nexus writing has on `ispyb_params`.

Note that this does mean that the nexus file is only written when all this data comes in, which is at a different point for rotations/grids. I think if we do https://github.com/DiamondLightSource/hyperion/issues/1245 then we're pretty close to being able to have one nexus callback for both

### To test:
1. Confirm no reference to ispyb_params in nexus callbacks
2. Confirm tests are comprehensive and pass

### DIFF
--- a/src/hyperion/experiment_plans/rotation_scan_plan.py
+++ b/src/hyperion/experiment_plans/rotation_scan_plan.py
@@ -208,8 +208,6 @@ def rotation_scan_plan(
         yield from bps.wait("setup_zebra")
 
         # get some information for the ispyb deposition and trigger the callback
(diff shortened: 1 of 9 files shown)


dperl-dls (approved):
Nice, thanks!

PR #1269, opened 2024-03-18
Remove MXSC (edge detection) AD plugin device

### LINKED ISSUE #1219: Tidy from change to Ophyd-async edge detection
After https://github.com/DiamondLightSource/dodal/pull/358 and https://github.com/DiamondLightSource/hyperion/pull/1214, the AreaDetector plugin for edge detection is no longer needed; we should remove it from Dodal and any plans which relate to it from Hyperion
- Rename the `PinTipDetection` to something like `EdgeDetection`
- Remove optional MXSC use from hyperion plans
- Remove MXSC from dodal, changing the built in utils which use it
- Remove the parameter that turns this on from Hyperion/GDA

dperl-dls:
Fixes #1219

Dodal PR: https://github.com/DiamondLightSource/dodal/pull/390

Updates the tests to work with the new pin tip device, and removes the ability to choose between the two devices.

Bumps param version to 4.0.5 - Will require a GDA change to bump param version and remove the `use_ophyd_edge_detect` param. 

### To test:
1. Run tests


olliesilvester (approved):
This looks good, and everything seems to work. There is a comment on the dodal PR where I think a few extra things can be removed. 
Also, we need to remember to not merge until the GDA param versions have been updated

PR #1271, opened 2024-03-18
Clarify which transmission is which

### LINKED ISSUE #1033: Be clear about desired vs actual transmission
Currently, inside xray centring, we take the transmission specified in the ispyb section of the supplied parameters to mean the desired transmission and set it on the device. We then read this back to find the actual transmission and put that back in the ispyb params object to put in ispyb.

We should instead specify the desired transmission as part of the experiment parameters, then use that to set on the hardware and read the actual back from there to put in ispyb. Note that the removal of the putting the readback into `ispyb_params` will be done as the last part of https://github.com/DiamondLightSource/hyperion/issues/1218. So this issue mostly boils down to move desired_transmission into the experiment params (not in the base class as not used by `PinTipCentring`)

## Acceptance Criteria
* The desired transmission is set in the experiment params
* Plans no longer make reference to `ispyb_params` for this

dperl-dls:
Fixes #1033 
Transmission is now an experiment param; only the readback value is stored in the ispyb params, which will be resolved later.

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/394/files

GDA change in  https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/41468

### To test:
1. Run tests


olliesilvester (approved):
Looks good! I like that the experiment parameters now all inherit from `BaseModel`, but do we need separate `AbstractExperimentParameterBase` and  `AbstractExperimentWithBeamParams`? Won't everything always need a `transmission_fraction`?

olliesilvester:
Also I guess we shouldn't merge until we make the GDA change which puts `transmission fraction` in `experiment params`

PR #1279, opened 2024-03-22
1217 Use event from grid detect to do ispyb deposition

### LINKED ISSUE #1217: Use event from grid detect to do ispyb deposition
Following https://github.com/DiamondLightSource/hyperion/issues/1146 we have a `GridScanInfo` object that we could easily create from an event at the grid detect stage e.g. do grid detect at 0 omega and produce one doc with that info then rotate and produce the next. To do this we would need to:
* Move ispyb deposition to before grid detect (at least)
* Emit these events
* Use them in the callback
* Remove the grid related stuff in ispyb_params

rtuck99:
Fixes #1217

Link to dodal PR (if required):N/A

### To test:
1. Unit + System tests pass
2. Grid scan details in ispyb and snapshots in syncweb should continue to populate


src/hyperion/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -43,6 +42,21 @@
...
+            "activate_callbacks": ["GridscanISPyBCallback"],
+            "subplan_name": CONST.PLAN.GRID_DETECT_AND_DO_GRIDSCAN,
+            "hyperion_internal_parameters": parameters.json(),
+        }
+    )
+    @wraps(plan_generator)
+    def wrapped():
+        return plan_generator
+
+    yield from wrapped()
```
DominicOram, line 57: Should: 
```suggestion
def ispyb_activation_wrapper(parameters, plan_generator):
    return bpp.run_wrapper(
        plan_generator,
        md={
            "activate_callbacks": ["GridscanISPyBCallback"],
            "subplan_name": CONST.PLAN.GRID_DETECT_AND_DO_GRIDSCAN,
            "hyperion_internal_parameters": parameters.json(),
        },
    )
```

DominicOram: Nit: I think you could then switch the `parameters` and `plan_generator` to more closely match the `run_wrapper` syntax

src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
```diff
@@ -129,14 +130,12 @@ def activity_gated_start(self, doc: RunStart):
...
+        ), "Expect an existing DataCollection to update"
         params = cast(RotationInternalParameters, params)
-        initial_collection_info = populate_data_collection_info_for_rotation(
-            params.hyperion_params.ispyb_params,
-            params.hyperion_params.detector_params,
-            params,
-        )
-        initial_collection_info = replace(
-            initial_collection_info,
+        data_collection_info = replace(
```
DominicOram, line 137: Should: Is this now taking an empty data collection and recreating it from the existing `event_sourced_data_collection_info`? Why can't we just use the `event_sourced_data_collection_info` itself?

src/hyperion/external_interaction/callbacks/ispyb_callback_base.py
```diff
@@ -90,60 +103,130 @@ def activity_gated_event(self, doc: Event) -> Event:
...
+            x_steps=data["oav_snapshot_num_boxes_x"],
+            y_steps=data["oav_snapshot_num_boxes_y"],
+            # convert pixels back to mm
+            x_step_size_mm=data["oav_snapshot_box_width"]
+            * self.params.hyperion_params.ispyb_params.microns_per_pixel_x
+            / 1000,
+            y_step_size_mm=data["oav_snapshot_box_width"]
+            * self.params.hyperion_params.ispyb_params.microns_per_pixel_y
+            / 1000,
+        )
```
DominicOram, line 170: Should: Is the only reason we need the grid scan info for constructing the comment? It seems messy to make this object here, use it once, then use the same parameters to make a `DataCollectionGridInfo` straight after. Could we change `construct_comment_for_gridscan` to take a `DataCollectionGridInfo` or to just take the parameters it needs then we can get rid of `GridScanInfo`?

src/hyperion/external_interaction/ispyb/ispyb_store.py
```diff
@@ -51,51 +51,61 @@ def experiment_type(self) -> str:
...
         scan_data_infos: Sequence[ScanDataInfo],
-    ):
+    ) -> IspybIds:
         assert (
             ispyb_ids.data_collection_group_id
         ), "Attempted to store scan data without a collection group"
         assert (
             ispyb_ids.data_collection_ids
         ), "Attempted to store scan data without a collection"
+        print(f"OBJECT PASSED IN IS {ispyb_ids}")
```
DominicOram, line 78: Should: I assume this is for debugging? Can you remove please? Same with other places in this file.

src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
DominicOram: Similar in the `ispyb_callback` for xray centring.

src/hyperion/external_interaction/callbacks/ispyb_callback_base.py
```diff
@@ -90,60 +103,130 @@ def activity_gated_event(self, doc: Event) -> Event:
...
+        scan_data_info = ScanDataInfo(
+            data_collection_info=data_collection_info,
+            data_collection_id=data_collection_id,
+            data_collection_grid_info=data_collection_grid_info,
+        )
+        self.ispyb_ids = self.ispyb.update_deposition(
+            self.ispyb_ids, None, [scan_data_info]
+        )
+        self._oav_snapshot_event_idx += 1
+        pass
```
DominicOram, line 204: Should: What is this `pass` doing?

src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
rtuck99: yes, I should have spotted this!

src/hyperion/external_interaction/callbacks/ispyb_callback_base.py
rtuck99: I will get rid of it

src/hyperion/external_interaction/ispyb/ispyb_store.py
rtuck99: Doh

src/hyperion/external_interaction/callbacks/ispyb_callback_base.py
```diff
@@ -90,60 +103,130 @@ def activity_gated_event(self, doc: Event) -> Event:
...
-            if doc["data"]["attenuator_actual_transmission"]:
-                # Ispyb wants the transmission in a percentage, we use fractions
-                self._event_driven_data_collection_info.transmission = (
-                    doc["data"]["attenuator_actual_transmission"] * 100
+        match event_descriptor.get("name"):
+            case CONST.DESCRIPTORS.ISPYB_HARDWARE_READ:
+                self._handle_ispyb_hardware_read(doc)
+            case CONST.DESCRIPTORS.OAV_SNAPSHOT_TRIGGERED:
+                self._handle_oav_snapshot_triggered(doc)
+            case CONST.DESCRIPTORS.ISPYB_TRANSMISSION_FLUX_READ:
```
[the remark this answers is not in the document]
rtuck99: historical reasons really, that's the original underlying behaviour that has always been there.

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_mapping.py
```diff
@@ -1,101 +1,68 @@
...
 
-from hyperion.external_interaction.callbacks.common.ispyb_mapping import GridScanInfo
 from hyperion.external_interaction.ispyb.data_model import (
     DataCollectionGridInfo,
     DataCollectionInfo,
 )
-from hyperion.external_interaction.ispyb.ispyb_dataclass import Orientation
 
 
 def populate_xz_data_collection_info(
```
DominicOram, line 12: Should: Why does this need `full_params`?

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_mapping.py
```diff
@@ -1,101 +1,68 @@
...
-    ispyb_params,
     detector_params,
 ) -> DataCollectionInfo:
     assert (
         detector_params.omega_start is not None
         and detector_params.run_number is not None
-        and ispyb_params is not None
         and full_params is not None
     ), "StoreGridscanInIspyb failed to get parameters"
     omega_start = detector_params.omega_start + 90
```
DominicOram, line 21: Should: We should be able to read the omega from the smargon when we do the pin tip detection, can we make an issue for that?

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -97,35 +105,38 @@ def activity_gated_start(self, doc: RunStart):
...
-                    None,
-                    populate_xy_data_collection_info(
-                        grid_scan_info,
-                        self.params,
-                        self.params.hyperion_params.ispyb_params,
+            scan_data_infos = [
+                ScanDataInfo(
+                    data_collection_info=populate_remaining_data_collection_info(
+                        None,
+                        None,
```
DominicOram, line 113: Should: Shouldn't the `dcgid` here be the group made above?

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -97,35 +105,38 @@ def activity_gated_start(self, doc: RunStart):
...
-
-            scan_data_info = ScanDataInfo(
-                data_collection_info=populate_remaining_data_collection_info(
-                    constructor,
-                    None,
-                    populate_xy_data_collection_info(
-                        grid_scan_info,
-                        self.params,
-                        self.params.hyperion_params.ispyb_params,
+            scan_data_infos = [
```
DominicOram, line 109: Should: Rather than do anything specific upfront could we get away with doing all the specific stuff in `_handle_oav_snapshot_triggered`? e.g. this could be something like:
```python
number_of_collections = 2 if self.is_3d_gridscan() else 1

scan_data_infos = [ScanDataInfo(data_collection_info=populate_remaining_data_collection_info(
                            None,
                            None,
                            DataCollectionInfo(
                                axis_range=0,
                                axis_end=omega_start,
                            ),
                            self.params.hyperion_params.detector_params,
                            self.params.hyperion_params.ispyb_params,
                        ) for range in (number_of_collections)]
```

Then handle `run_number` and `omega_start` in `_handle_oav_snapshot_triggered`?

DominicOram (changes requested):
Apologies, some more comments in code.

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_mapping.py
rtuck99: Have made #1327 but is pin tip detection the right place or should it be done during `read_hardware_for_ispyb_pre_collection`?

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_callback.py
rtuck99: we haven't made it yet, we've only populated the `DataCollectionGroupInfo` but it hasn't been posted to the database so there's no id, leaving it as none will cause it to be populate during the call to `begin_deposition`

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_callback.py
rtuck99: I guess this is a wider question than for this one particular change but given that we are planning to move to ispyb-eye is it better to defer some of our internal changes until that point? At least until it's clearer how start / event documents might map to ispyb-eye calls? Unless you were proposing this with that already in mind.

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_mapping.py
DominicOram: It can't be done during `read_hardware_for_ispyb_pre_collection` because we want the omega for each grid. If we read omega at pre_collection we only get the first grid's omega. We would then have to add the 90 deg in the callback again, which I was trying to avoid. Basically I want the logic of what we're doing to be in the plan, the callback should just be blindly funneling it into ispyb

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_callback.py
DominicOram: Sure, happy to park this :)

DominicOram (approved):
Great, thank you! I think this all looks a lot cleaner now

PR #1281, opened 2024-03-25
Prepare for robot load

### LINKED ISSUE #1265: Prepare for robot load
To be able to do the robot load as in https://github.com/DiamondLightSource/hyperion/pull/1250 we must be in a position where the robot is allowed to unload/load. Specifically we need to at least:

* Move the aperture/scatterguard to robot load
* Reset the stub offsets in the smargon to robot load
* Move the smargon to the robot load position (which is all 0s)

## Acceptance Criteria

* We can reliably get into a position that the robot can load from

DominicOram:
Fixes #1265

### To test:
1. Confirm new unit test passes

### DIFF
--- a/src/hyperion/experiment_plans/robot_load_then_centre_plan.py
+++ b/src/hyperion/experiment_plans/robot_load_then_centre_plan.py
@@ -6,7 +6,7 @@ from typing import cast
 
 import bluesky.plan_stubs as bps
 from blueapi.core import BlueskyContext, MsgGenerator
-from dodal.devices.aperturescatterguard import ApertureScatterguard
+from dodal.devices.aperturescatterguard import AperturePositions, ApertureScatterguard
 from dodal.devices.attenuator import Attenuator
 from dodal.devices.backlight import Backlight
 from dodal.devices.DCM import DCM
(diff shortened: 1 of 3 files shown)


src/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -113,10 +113,34 @@ def wait_for_smargon_not_disabled(smargon: Smargon, timeout=60):
...
+def prepare_for_robot_load(composite: RobotLoadThenCentreComposite):
+    yield from bps.abs_set(
+        composite.aperture_scatterguard,
+        AperturePositions.ROBOT_LOAD,
+        group="prepare_robot_load",
+    )
+
+    yield from bps.mv(composite.smargon.stub_offsets, StubPosition.RESET_TO_ROBOT_LOAD)
+
+    # fmt: off
```
olliesilvester, line 125: What is fmt here?

olliesilvester (approved):
Nice!

src/hyperion/experiment_plans/robot_load_then_centre_plan.py
DominicOram: Turns `black` off. If you let `black` reformat the line it becomes less readable (I think) as it's not obvious which value corresponds to which axis.

PR #1284, opened 2024-03-26
New external parameter model

### LINKED ISSUE #698: improve external parameters and make pydantic models for them
to replace the JSON schema for validation
as part of this, the external parameter structure should be improved, to remove references to ISPYB etc
possibly depends on getting some values from hardware

dperl-dls:
Fixes #698

Introduces a new set of pydantic classes to replace the old external params. Eventually these will replace the internal params too.

Associated GDA changes at https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/41445

### To test:
1. Run tests


src/hyperion/parameters/gridscan.py
```diff
@@ -0,0 +1,263 @@
...
+            "detector_distance": self.detector_distance_mm,
+            "omega_start": self.omega_start_deg,
+            "omega_increment": 0,
+            "num_images_per_trigger": 1,
+            "num_triggers": self.num_images,
+            "use_roi_mode": self.use_roi_mode,
+            "det_dist_to_beam_converter_path": self.det_dist_to_beam_converter_path,
+            "run_number": self.run_number,
+            "trigger_mode": self.trigger_mode,
+        }
```
DominicOram, line 69: Should: Is this dict conversion required so the `DetectorParams` validators get run? If so can we add a comment that says specifically which validators this is needed for? If not can we just make it with the usual constructor?

src/hyperion/parameters/gridscan.py
```diff
@@ -0,0 +1,263 @@
...
+        return DetectorParams(**detector_params)
+
+    @property
+    def ispyb_params(
+        self,
```
DominicOram, line 74: Nit: Is `black` only putting this on a new line because of the comma? I think remove the comma...

src/hyperion/parameters/gridscan.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class PinTipCentreThenXrayCentre(GridCommon):
+    tip_offset_um: float = 0
+
+
+class RobotLoadThenCentre(GridCommon, WithSample): ...
+
+
+class SpecifiedGridScan(GridCommon, XyzStarts, WithScan, WithSample):
```
DominicOram, line 50: Should: Can you add some docstrings around what specified means in this context?

src/hyperion/parameters/gridscan.py
```diff
@@ -0,0 +1,263 @@
...
+
+# Doesn't yet exist but will look something like this
+class DoOneUDC(GridCommon):
+    """Diffraction data is for grids at start"""
+
+    rotation_exposure_s: float
+    rotation_axis: RotationAxis = RotationAxis.OMEGA
+    rotation_angle_deg: float
+    rotation_increment_deg: float
+    rotation_direction: RotationDirection
```
DominicOram, line 263: Should: I don't think we should put this in yet as it's not used

src/hyperion/parameters/gridscan.py
```diff
@@ -0,0 +1,263 @@
...
+    def _validate_axis_2(cls, axis_2: XyzAxis, values) -> XyzAxis:
+        if axis_2 == values["axis_1"]:
+            raise ValueError(
+                f"Axis 1 ({values['axis_1']}) and axis 2 ({axis_2}) cannot be equal!"
+            )
+        return axis_2
+
+    @property
+    def normal_axis(self) -> XyzAxis:
+        return ({XyzAxis.X, XyzAxis.Y, XyzAxis.Z} ^ {self.axis_1, self.axis_2}).pop()
```
DominicOram, line 132: Should: I think a docstring here would be useful

src/hyperion/parameters/gridscan.py
```diff
@@ -0,0 +1,263 @@
...
+    def scan_2(self) -> TwoDGridScan:
+        values = self.dict()
+        values["y_start_um"] = self.y2_start_um
+        values["z_start_um"] = self.z2_start_um
+        values["axis_1"] = XyzAxis.X
+        values["axis_2"] = XyzAxis.Z
+        values["axis_1_steps"] = self.x_steps
+        values["axis_2_steps"] = self.z_steps
+        self._delete_excess(values)
+        return TwoDGridScan(**values)
```
DominicOram, line 225: Should: This all feels like a messy way to represent the geometry of the system. Is it possible to leverage scanspec better to represent the geometry then use that to create the `GridScanParams`?

src/hyperion/parameters/rotation.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class RotationScan(
+    DiffractionExperiment,
+    WithScan,
+    OptionalGonioAngleStarts,
+    OptionalXyzStarts,
+    WithSample,
+):
+    omega_start_deg: float = 0  # type: ignore
```
DominicOram, line 32: Nit: Why is this not `Field(default=0)`?

src/hyperion/parameters/rotation.py
```diff
@@ -0,0 +1,101 @@
...
+            "undulator_gap": self.ispyb_extras.undulator_gap,
+            "synchrotron_mode": self.ispyb_extras.synchrotron_mode,
+            "slit_gap_size_x": self.ispyb_extras.slit_gap_size_x,
+            "slit_gap_size_y": self.ispyb_extras.slit_gap_size_y,
+            "xtal_snapshots_omega_start": self.ispyb_extras.xtal_snapshots_omega_start,
+            "xtal_snapshots_omega_end": self.ispyb_extras.xtal_snapshots_omega_end,
+            "ispyb_experiment_type": "SAD",
+            "upper_left": self.ispyb_extras.upper_left,
+        }
+        return RotationIspybParams(**ispyb_params)
```
DominicOram, line 86: Should: Why are we converting from a dict here and not in gridscan? My preference is just usual instantiation as you get better type hints etc.

src/hyperion/parameters/rotation.py
```diff
@@ -0,0 +1,101 @@
...
+            "detector_distance": self.detector_distance_mm,
+            "omega_start": self.omega_start_deg,
+            "omega_increment": self.rotation_increment_deg,
+            "num_images_per_trigger": self.num_images,
+            "num_triggers": 1,
+            "use_roi_mode": False,
+            "det_dist_to_beam_converter_path": self.det_dist_to_beam_converter_path
+            or CONST.PARAM.DETECTOR.BEAM_XY_LUT_PATH,
+            "run_number": self.run_number,
+        }
```
DominicOram, line 57: Should: As above, why is this a dict rather than usual instantiation?

DominicOram (changes requested):
Thanks, I think this is getting towards the right direction. Some comments in code. We can have a chat about the geometry thing, not sure what a better representation is off the top of my head.

src/hyperion/parameters/gridscan.py
DominicOram: Agreed that 3D and 2D grid scans are unrelated and don't need to refer to each other. Additionally, `scan_points` is the whole scan (including rotation in 3D case), `num_images` can just be derived from this but we then need a new property which is where to index the Nexus files to.

src/hyperion/parameters/gridscan.py
```diff
@@ -0,0 +1,259 @@
...
+        scan_2_z = Line("sam_z", self.z2_start_um, z2_end, self.z_steps)
+        return scan_2_x.zip(scan_2_y).zip(scan_2_omega) * ~scan_2_z
+
+    @property
+    def scan_indices(self):
+        """The first index of each gridscan"""
+        return [0, len(ScanPath(self.scan_1.calculate()).consume().midpoints["sam_x"])]
+
+    @property
+    def scan_spec(self):
```
olliesilvester, line 236: A one line doc string on all these properties would be useful

olliesilvester (commented):
I really like this, this new parameter model is much more intuitive. I think a section on the wiki summarising how it works and what various keywords mean would be very useful for onboarding too.

Also, it would be good to have some tests for the `gridscan.py` properties, like `scan_points` and `scan_indicies`

src/hyperion/experiment_plans/experiment_registry.py
```diff
@@ -69,37 +65,37 @@ class ExperimentRegistryEntry(TypedDict):
...
         "setup": pin_centre_then_xray_centre_plan.create_devices,
-        "internal_param_type": PinCentreThenXrayCentreInternalParameters,
+        "internal_param_type": PinTipCentreThenXrayCentre,
         "experiment_param_type": PinCentreThenXrayCentreParams,
         "callbacks_factory": create_gridscan_callbacks,
     },
     "wait_for_robot_load_then_centre": {
         "setup": wait_for_robot_load_then_centre_plan.create_devices,
-        "internal_param_type": WaitForRobotLoadThenCentreInternalParameters,
+        "internal_param_type": RobotLoadThenCentre,
```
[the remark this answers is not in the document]
DominicOram: They sort of are. In this case `centre` means pin tip centre centre then xray centre. I think it's ok to basically assume "centre" means do the full centring. But I appreciate it's not obvious

src/hyperion/parameters/rotation.py
```diff
@@ -0,0 +1,155 @@
...
+    shutter_opening_time_s: float = Field(default=CONST.I03.SHUTTER_TIME_S)
+    rotation_angle_deg: float
+    rotation_increment_deg: float
+    rotation_direction: RotationDirection
+    transmission_frac: float
+    ispyb_extras: TemporaryIspybExtras
+
+    @property
+    def directory(self):
+        directory = str(self.visit_directory / "auto" / str(self.sample_id))
```
DominicOram, line 54: Must: This certainly isn't right. From looking round on the file system it looks like some use sample name some use something else. We may have to provide the `prefix` as in http://agamemnon.diamond.ac.uk/example/collect for now then discuss with SIMS/Analysis how that is actually built.

DominicOram: We could just redefine it for UDC and say it is always the sample_id but we should shout a bit louder before we do that

src/hyperion/parameters/rotation.py
```diff
@@ -0,0 +1,155 @@
...
+    DiffractionExperiment,
+    WithScan,
+    OptionalGonioAngleStarts,
+    OptionalXyzStarts,
+    WithSample,
+):
+    omega_start_deg: float = Field(default=0)  # type: ignore
+    rotation_axis: RotationAxis = Field(default=RotationAxis.OMEGA)
+    shutter_opening_time_s: float = Field(default=CONST.I03.SHUTTER_TIME_S)
+    rotation_angle_deg: float
```
DominicOram, line 46: Should: `rotation_angle_deg` wasn't immediately obvious to me when I looked at it. Could you add a docstring to say what it is?

src/hyperion/parameters/rotation.py
```diff
@@ -0,0 +1,155 @@
...
+            ispyb_experiment_type="SAD",
+            flux=self.ispyb_extras.flux,
+        )
+
+    @property
+    def scan_points(self) -> AxesPoints:
+        scan_spec = Line(
+            axis="omega",
+            start=self.omega_start_deg,
+            stop=(self.rotation_angle_deg + self.omega_start_deg),
```
DominicOram, line 118: Nit: The addition being the other way round makes more sense to me but whatever

DominicOram (changes requested):
Some more comments, sorry. I think we are getting there though...

DominicOram:
Additionally, can we update https://github.com/DiamondLightSource/hyperion/wiki/Parameter-model a bit?

src/hyperion/parameters/gridscan.py
```diff
@@ -0,0 +1,389 @@
...
+            self.x_start_um, self.y_start_um, self.z_start_um
+        )
+
+    @property
+    def axis_1_end_um(self) -> float:
+        return self.axis_1_start_um + self.axis_1_step_size_um * self.axis_1_steps
+
+    @property
+    def axis_2_end_um(self) -> float:
+        return self.axis_2_start_um + self.axis_2_step_size_um * self.axis_2_steps
```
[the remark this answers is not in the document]
dperl-dls: On discussion removed 2D grid scan params entirely, we will know what it should look like when we need to add it back

src/hyperion/parameters/rotation.py
dperl-dls: I renamed it to scan_width_deg

DominicOram (changes requested):
Thanks! Please update the docs at https://github.com/DiamondLightSource/hyperion/wiki/Parameter-model to something close to where we want to be with this and then happy to approve.

DominicOram (approved):
Good docs, ta!

PR #1291, opened 2024-04-04
Add ispyb entries for robot load

### LINKED ISSUE #1017: ISPYB: Deposit robot load entry

DominicOram:
Fixes #1017

### To test:
1. Confirm tests pass (including system test, which you will have to manually confirm)

### DIFF
--- a/src/hyperion/experiment_plans/experiment_registry.py
+++ b/src/hyperion/experiment_plans/experiment_registry.py
@@ -17,6 +17,7 @@ from hyperion.experiment_plans import (
 from hyperion.external_interaction.callbacks.common.callback_util import (
     CallbacksFactory,
     create_gridscan_callbacks,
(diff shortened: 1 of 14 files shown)


dperl-dls (approved):
Looks good, tests pass etc. Exp eye class/methods could use some docstrings maybe

DominicOram:
> Looks good, tests pass etc. Exp eye class/methods could use some docstrings maybe

I agree, will add some before merging

PR #1294, opened 2024-04-05
Make common FGS and convert to ophyd async

olliesilvester:
Fixes dodal  #[413](https://github.com/DiamondLightSource/dodal/issues/413)

Link to dodal PR (if required): #[422](https://github.com/DiamondLightSource/dodal/pull/422)

After this is merged, we should be able to combine the `flyscan_xray_centre` and `panda_flyscan_xray_centre` plans

### To test:
1. Confirm functionality is unchanged


tests/conftest.py
```diff
@@ -589,8 +590,12 @@ def fake_fgs_composite(
         False
     )
```
rtuck99, line 592: I assume this is to be invoked separately from future tests and this commented out code can be removed?

rtuck99 (approved):
All ok - apart from comment re commented out code

DominicOram (changes requested):
Just the one comment, thanks!

DominicOram (approved):
Great, thank you!

PR #1300, opened 2024-04-09
(#1299) Add regex filter to include only build tags in the deployment script

### LINKED ISSUE #1299: deployment script should filter out only tags which correspond to actual releases
Running `deploy_hyperion.py` on my local dev environment ran into issues because it picked up other local tags that weren't related to releases.

rtuck99:
Fixes #1299

### To test:
1. deploying a release from dev workspace as per the wiki should now work even if you have other tags in your git repository

### DIFF
--- a/utility_scripts/deploy/deploy_hyperion.py
+++ b/utility_scripts/deploy/deploy_hyperion.py
@@ -5,11 +5,11 @@ from subprocess import PIPE, CalledProcessError, Popen
 from uuid import uuid1
 
 from git import Repo
-from packaging.version import Version
+from packaging.version import VERSION_PATTERN, Version
 
 recognised_beamlines = ["dev", "i03", "i04"]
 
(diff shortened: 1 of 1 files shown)


utility_scripts/deploy/deploy_hyperion.py
```diff
@@ -8,6 +9,8 @@
 
 recognised_beamlines = ["dev", "i03", "i04"]
 
+VERSION_PATTERN = re.compile(r"^v?(\d+)\.(\d+)\.(\d+)$")
```
DominicOram, line 12: Could: A more comprehensive pattern is in `packaging.version.VERSION_PATTERN`, we could use that instead so we don't have to maintain our own?

DominicOram (approved):
Thanks, suggested a slight improvement but not too bothered so take it or leave it.

utility_scripts/deploy/deploy_hyperion.py
rtuck99: done

PR #1305, opened 2024-04-10
819 add dodal and nexgen versions to release notes

### LINKED ISSUE #819: Add dodal and nexgen version to release notes

rtuck99:
Fixes #819 

### To test:
1. Perform a release as per the instructions on the wiki https://github.com/DiamondLightSource/hyperion/wiki/Creating-a-release-and-deploying-it
2. Confirm nexgen and dodal versions appear in release notes along with normal generated release notes
3. Confirm released version setup.cfg has dependencies correctly pinned.


DominicOram:
I created the release at https://github.com/DiamondLightSource/hyperion/releases/tag/untagged-0eb09fe04a1a13594f36 but it had some issues:
* Must: The code was pinned to `dodal == 1.13.1` but the latest is https://github.com/DiamondLightSource/dodal/releases/tag/1.22.0 (https://pypi.org/project/dls-dodal/)
* Could: It would be better if rather than just saying "these were the versions" links were provided to the release notes
* Nit: I think the version numbers are less interesting than what's changed so can we put them at the bottom of the release notes?

DominicOram (dismissed):
I created the release at https://github.com/DiamondLightSource/hyperion/releases/tag/untagged-0eb09fe04a1a13594f36 but it had some issues:
* Must: The code was pinned to `dodal == 1.13.1` but the latest is https://github.com/DiamondLightSource/dodal/releases/tag/1.22.0 (https://pypi.org/project/dls-dodal/)
* Could: It would be better if rather than just saying "these were the versions" links were provided to the release notes
* Nit: I think the version numbers are less interesting than what's changed so can we put them at the bottom of the release notes?

rtuck99:
> I created the release at https://github.com/DiamondLightSource/hyperion/releases/tag/untagged-0eb09fe04a1a13594f36 but it had some issues:
> 
>     * Must: The code was pinned to `dodal == 1.13.1` but the latest is https://github.com/DiamondLightSource/dodal/releases/tag/1.22.0 (https://pypi.org/project/dls-dodal/)
> 
>     * Could: It would be better if rather than just saying "these were the versions" links were provided to the release notes
> 
>     * Nit: I think the version numbers are less interesting than what's changed so can we put them at the bottom of the release notes?

Hmm, looking at the logs, it installs the right version of dodal.... then it installs the wrong version on top. Not sure if it's related to the bluesky api conflict but will investigate

DominicOram:
Could be, looks like the `blueapi` pin was wrong, we wanted the pre-release of https://github.com/DiamondLightSource/blueapi/releases/tag/0.4.2-a1.

DominicOram:
As discussed, I think the correct answer is that we should have been pinning our `blue-api` version before the release.

rtuck99:
Ok, I think the blue-api versioning issue is another thing to the dodal versioning issue - I think these are the alternative ways of pinning.

* We `pip install` with `--pre` and explicitly pin all our "stable" dependencies to release versions, don't pin things like `dodal`, `bluesky` etc.
* We never install with `--pre` and we only explicitly pin those dependencies where we want to be on the bleeding edge pre-release versions
* We explicitly pin everything

I think it's a toss up between 1 and 2 - 1 has more things pinned but we might update less frequently

DominicOram:
Agreed that will go for option 2 and pin `bluesky`, `blueapi` and `ophyd-async`

dperl-dls (approved):
Great, thanks!

PR #1306, opened 2024-04-10
Use bit depth from the Eiger to work out datatype of gridscan VDS

### LINKED ISSUE #1245: XRC bit depth
The XRC callback should also use the bit depth to work out datatype (as in https://github.com/DiamondLightSource/hyperion/issues/629).

## Acceptance Criteria
* The XRC Nexus callback gets bit depth from the eiger and passes this into the nexus writer
* There is no longer a default for bit depth used on the nexus writer

DominicOram:
Fixes #1245

### To test:
1. Confirm new tests pass

### DIFF
--- a/src/hyperion/experiment_plans/panda_flyscan_xray_centre_plan.py
+++ b/src/hyperion/experiment_plans/panda_flyscan_xray_centre_plan.py
@@ -22,6 +22,7 @@ from hyperion.device_setup_plans.manipulate_sample import move_x_y_z
 from hyperion.device_setup_plans.read_hardware_for_setup import (
     read_hardware_for_ispyb_during_collection,
     read_hardware_for_ispyb_pre_collection,
+    read_hardware_for_nexus_writer,
 )
 from hyperion.device_setup_plans.setup_panda import (
     disarm_panda_for_gridscan,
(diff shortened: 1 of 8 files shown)


src/hyperion/experiment_plans/panda_flyscan_xray_centre_plan.py
```diff
@@ -22,6 +22,7 @@
 from hyperion.device_setup_plans.read_hardware_for_setup import (
     read_hardware_for_ispyb_during_collection,
     read_hardware_for_ispyb_pre_collection,
+    read_hardware_for_nexus_writer,
```
rtuck99, line 25: These `read_hardware` methods are quite tightly coupled to the corresponding callbacks. Perhaps we should consider making it a bit more explicit, rather than dumping them all into one general-purpose module perhaps they should be in a separate sub-package in modules named after the corresponding callback, along with the necessary decorator/wrappers to register the CB, that way it's obvious to a consumer of the plans what is needed to use them.

rtuck99 (approved):
All looks good

src/hyperion/experiment_plans/panda_flyscan_xray_centre_plan.py
DominicOram: Thanks, I agree it's not clear. Made https://github.com/DiamondLightSource/hyperion/issues/1313 to discuss

Issue #433, opened 2022-12-12
Create logging structure for oav physical tests


evalott100:
An issue to keep track of changes made while testing OAV centring on i03.

To test:
* Run OAV centring on I03.
* Check logs are expected.

DominicOram:
Closing as no longer relevant

PR #1315, opened 2024-04-12
update to match dodal #433

dperl-dls:


DominicOram (approved):
Thank you. Logging is a lot cleaner now. I've pushed a change that moves your zebra mocking into the zebra specific fixture so that we don't have to do it multiple times, let me know if you disagree.

Issue #1275, opened 2024-03-19
Implement agreed parameter model changes


dperl-dls:
In https://github.com/DiamondLightSource/hyperion/issues/1092 we settled on a new parameter model. We will implement this this in several steps:

- [ ] First, only replace the external parameters, and convert on entry to the plan. (https://github.com/DiamondLightSource/hyperion/issues/698)
E.g.:
```Python
def flyscan_xray_centre(
    composite: FlyScanXRayCentreComposite,
    parameters: GridScan,
) -> MsgGenerator:
    internal_parameters = parameters.to_internal()
    ...
```
This is unnecessary in the long run, but it lets us get rid of the ugly conversion magic in the current internal param classes without breaking everything at once.

- [ ] Switching over (mainly) the callbacks (and other things which rely on them) to use the new param types
- [ ] Removing the old internal params completely

dperl-dls:
Since we decided #1286 was unneeded, this is done!

PR #1330, opened 2024-04-24
(#1275) convert pin and grid detect plans to new params

dperl-dls:
part of #1275

Uses the new params for pin/grid detect and xray centre (leaves xray centre alone still)

### To test:
1. Run tests


rtuck99 (approved):
I think it's all there now, looks like it was just upper_left, I was probably thinking of stuff that wasn't in that last ispyb PR

Issue #1289, opened 2024-04-03
Switch over the plans to use the new param types


dperl-dls:
Should be split up into separate PRs per plan type to keep the size down a bit

PR #1331, opened 2024-04-24
Use new params in XrayCentre

dperl-dls:
Building on #1330 uses the new params for the xraycentre plans

Towards #1289 

### To test:
1. Run tests


tests/test_data/new_parameter_json_files/test_gridscan_param_defaults.json
```diff
@@ -0,0 +1,55 @@
...
+    "y_start_um": 0.0,
+    "y2_start_um": 0.0,
+    "z_start_um": 0.0,
+    "z2_start_um": 0.0,
+    "detector_distance_mm": 100.0,
+    "omega_start_deg": 0.0,
+    "exposure_time_s": 0.1,
+    "set_stub_offsets": true,
+    "ispyb_extras": {
+        "microns_per_pixel_x": 1.25,
```
rtuck99, line 32: Ditto comments in previous PR re microns_per_pixel

src/hyperion/external_interaction/callbacks/grid_detection_callback.py
```diff
@@ -1,15 +1,32 @@
...
+    y2_start_um: float
+    z_start_um: float
+    z2_start_um: float
+    x_steps: int
+    y_steps: int
+    z_steps: int
+    x_step_size_um: float
+    y_step_size_um: float
+    z_step_size_um: float
+    set_stub_offsets: bool
```
DominicOram, line 27: Should: Some of these parameters don't change from the grid detection so I don't think we need to have them in here at all. I can't see any reasonable way that we would want to change them based on an optical detection in the near future either:
* `set_stub_offsets`
* `transmission_frac`
* `exposure_time_s`

src/hyperion/external_interaction/callbacks/grid_detection_callback.py
```diff
@@ -1,15 +1,32 @@
...
+
+
 class GridDetectionCallback(CallbackBase):
```
DominicOram, line 30: Should: I don't think `run_up_distance_mm` is ever used here now so we can remove

src/hyperion/parameters/gridscan.py
```diff
@@ -94,6 +97,7 @@ def detector_params(self):
         ), "Detector distance must be filled before generating DetectorParams"
         os.makedirs(self.storage_directory, exist_ok=True)
         return DetectorParams(
+            detector_size_constants=self.detector,  # type: ignore
```
DominicOram, line 100: Could: Can you put a comment here that this should get tidied by https://github.com/DiamondLightSource/hyperion/issues/1307

DominicOram (changes requested):
Awesome, this is so much cleaner, particularly the `UpdateParams`. Some comments to make it a step cleaner but otherwise good.

DominicOram (approved):
Great, thanks. I've added a commit to take it a step further and remove stub offsets and exposure time as they're not used. Otherwise all looks good now!

PR #1333, opened 2024-04-25
Take snapshots before robot load

### LINKED ISSUE #1282: Create snapshots after robot load
We need to do snapshots and put them in ispyb

* We already do OAV snapshots for pin tip etc.
* We will need to add those snapshot lines to the robot load plan
* The ispyb callback will then need to pick up the path to this
* We don't have the off-axis webcam as a ophyd device yet, we can hopefully reuse the OAV - if this is more complicated spin into a new issue

## Acceptance Criteria
* The snapshots are put into robot load as:
image

DominicOram:
Fixes #1282

Dodal PR: https://github.com/DiamondLightSource/dodal/pull/468

### To test:
1. Confirm tests pass
2. Run the modified system tests and confirm data appears in the dev ispyb database with snapshots (note https://jira.diamond.ac.uk/browse/LIMS-1304)

### DIFF
--- a/src/hyperion/external_interaction/callbacks/robot_load/ispyb_callback.py
+++ b/src/hyperion/external_interaction/callbacks/robot_load/ispyb_callback.py
@@ -62,7 +62,12 @@ class RobotLoadISPyBCallback(PlanReactiveCallback):
                 self.action_id is not None
             ), "ISPyB Robot load callback event called unexpectedly"
             barcode = doc["data"]["robot-barcode"]
-            self.expeye.update_barcode(self.action_id, barcode)
+            oav_snapshot = doc["data"]["oav_snapshot_last_saved_path"]
+            webcam_snapshot = doc["data"]["webcam-last_saved_path"]
+            # I03 uses oav/webcam snapshots in place of before/after snapshots
+            self.expeye.update_barcode_and_snapshots(
+                self.action_id, barcode, oav_snapshot, webcam_snapshot
+            )
 
         return super().activity_gated_event(doc)
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ install_requires =
     xarray
     doct
     databroker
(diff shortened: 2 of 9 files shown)


src/hyperion/external_interaction/callbacks/robot_load/ispyb_callback.py
```diff
@@ -62,7 +62,12 @@ def activity_gated_event(self, doc: Event) -> Event | None:
                 self.action_id is not None
             ), "ISPyB Robot load callback event called unexpectedly"
             barcode = doc["data"]["robot-barcode"]
-            self.expeye.update_barcode(self.action_id, barcode)
+            oav_snapshot = doc["data"]["oav_snapshot_last_saved_path"]
```
olliesilvester, line 65: How do you know if you need dashes or hyphens in lines like `webcam_snapshot = doc["data"]["webcam-last_saved_path"]`?

olliesilvester (approved):
Looks sensible, nice!

src/hyperion/external_interaction/callbacks/robot_load/ispyb_callback.py
DominicOram: So when you have a device like:
```python
class MyDevice():
    self.signal = ...
my_device = MyDevice()
```
Then you do `yield from bps.read(my_device)` the `RE` produces a document with data like:
```
data: {
    "X": "value"
}
```
Where:
* In `ophyd` `_` is used as a separator so X is `my_device_signal`
* `ophyd_async` instead uses `-` so X is `my_device-signal`

If you tried to look in the dict for the wrong one you would fail with a `KeyError`. This is a frustrating breaking change between the two but it does make sense because `_` is a valid (and commonly used) part of a variable name it's impossible to know which was the device and signal in `my_device_signal` by just looking at the string. By using `-` instead you can work that out.

PR #1337, opened 2024-04-29
Add panda support for writing positions

rtuck99:
* Change references from PandA to HDFPanda
* Apply directory path from params

Fixes DiamondLightSource/dodal#466

Link to dodal PR (if required): DiamondLightSource/dodal#477

### To test:
1. Tests pass
2. Hyperion now saves HDF files in the location specified by the `storage_directory` request parameter

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ install_requires =
     xarray
     doct
     databroker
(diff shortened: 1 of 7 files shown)


DominicOram (approved):
Great, thank you!

PR #1340, opened 2024-04-29
1293 remove barcode

### LINKED ISSUE #1293: Create data collection group once on start document
With the exception of the barcode, which we can remove (see https://diamondlightsource.slack.com/archives/C048RD0KFFF/p1712327655071409), all the information in the data collection group is fixed at start document. Therefore we can:
* Remove barcode reading from plans and setting in ispyb
* Make a simple `create_group` function
* The logic in `activity_gated_start` can then work out whether we need a group or not then make one (in the grid case we always want to and in rotation only sometimes)
* Get the `dcgid` of either this new group or the old one and set this as the parent id
* Never worry about groups again for the rest of the plan

rtuck99:
Fixes #1293

### To test:
1. All tests should pass
2. DataCollectionGroup info should be stored in ispyb as normal


rtuck99:
I didn't add the `create_group` method as described, as it didn't really seem to add anything. After removal of the barcode and subsequent removal of all population of DataCollectionGroupInfo during updates, there is no "worrying about groups" after `activity_start`, you just let `ispyb_store` populate your `ispyb_ids` and then forget about it.

The only possible benefit might be some separation in `ispyb_store.begin_or_update_deposition()` but at the expense of managing the ispyb connection across the separate calls during `begin_deposition()`.

DominicOram (approved):
Great, this cleans things up quite a lot, thanks!

Issue #1200, opened 2024-02-29
Panda grid scans: Investigate and fix occasional errors with 500hz data collections


olliesilvester:
Testing at 500hz mostly ran smoothly. There were two instances where it broke, I haven't investigated the cause yet. In both cases, once the smargon had ended its motion, it was in a position such that the panda sequencer table thought it should still be sending triggers to the eiger.

Relevant logging messages:

First error: https://graylog2.diamond.ac.uk/messages/logs_8673/f9c18e43-d6f4-11ee-a034-1866dafadc6c
Panda sequencer values were set here: https://graylog2.diamond.ac.uk/messages/logs_8673/a7a5a881-d6f4-11ee-a034-1866dafadc6c
Hyperion parameters:  https://graylog2.diamond.ac.uk/messages/logs_8673/a662a131-d6f4-11ee-a034-1866dafadc6c

Second error: https://graylog2.diamond.ac.uk/messages/logs_8673/2e7fe3a5-d6fc-11ee-a034-1866dafadc6c
Panda sequencer values: https://graylog2.diamond.ac.uk/messages/logs_8673/deadede0-d6fb-11ee-a034-1866dafadc6c
Hyperion parameters: https://graylog2.diamond.ac.uk/messages/logs_8673/dd41b3b1-d6fb-11ee-a034-1866dafadc6c

olliesilvester:
Looking at the archived PV values for the smargon at these times, it looks like the motion itself didn't occur properly. If it stops at an unexpected point, it would explain why the panda keeps sending eiger triggers until timeout.

DominicOram:
A good place to start would be running the scans at increasing speeds. If it only breaks at high speeds, it is likely we're losing encoder counts

DominicOram:
We think about 3 days

DominicOram:
We should pair up on doing this

olliesilvester:
The PCAP h5 files from these error have the following problems:

- The Eiger starts triggering before the smargon has moved to the starting position

- The panda keeps sending triggers to the eiger while the smargon is rotating through omega. The triggers should be stopped for this part

- The panda keeps sending triggers to the eiger after the scan is complete, this is basically the same issue as the previous point

olliesilvester:
This issue was seen again https://graylog2.diamond.ac.uk/messages/logs_8711/a15e5910-fe39-11ee-92ff-1866dafadc6c.

[...]

PR #1341, opened 2024-04-29
Panda improvements

### LINKED ISSUE #1338: PandA grid scans improvements
The following changes should be made to panda grid scans:

- Load from `/dls_sw/i03/software/daq_configuration/panda_configs/flyscan_pcap_ignore_seq.yaml`, which doesn't put anything into the sequencer table and so potentially fixes an overwriting issue
- Read back the sequencer table PV and log the values before arming the panda - then we can rule out any further issues being caused by sequencer values
- Log the time taken for the motion program to complete, then we can confirm that we're getting a good amount of speed-up, even with a higher run-up distance
- Increase the default run-up distance by 0.01mm, to 0.17mm. There's not way to tell if this run-up is enough, but at 0.16mm, it ran for almost an entire day before missing a (single) trigger. It would be nice to have a surer way of finding the lowest working value, but I don't know how we'd do that
- Decrease the timeout when waiting for triggers. This point might need some discussion, but Neil mentioned that it takes ages (~5 minutes) to clear up when a trigger is missed.

olliesilvester:
Fixes #1338 and #1200

I didn't add tests here since it was mainly just logging / I'm being a bit lazy, I can add some if needed though

### To test:
1. Confirm changes are sensible and that tests pass

### DIFF
--- a/src/hyperion/device_setup_plans/setup_panda.py
+++ b/src/hyperion/device_setup_plans/setup_panda.py
@@ -174,7 +174,7 @@ def setup_panda_for_flyscan(
 
     LOGGER.info(f"PandA sequencer table has been set to: {str(table)}")
     table_readback = yield from bps.rd(panda.seq[1].table)
-    LOGGER.info(f"PandA sequencer table readback is: {str(table_readback)}")
+    LOGGER.debug(f"PandA sequencer table readback is: {str(table_readback)}")
 
     yield from arm_panda_for_gridscan(panda)
 
--- a/src/hyperion/device_setup_plans/setup_panda.py
+++ b/src/hyperion/device_setup_plans/setup_panda.py
@@ -161,8 +161,6 @@ def setup_panda_for_flyscan(
 
     table = get_seq_table(parameters, exposure_distance_mm)
 
-    LOGGER.info(f"Setting PandA sequencer values: {str(table)}")
-
     yield from bps.abs_set(panda.seq[1].table, table, group="panda-config")
 
     yield from bps.abs_set(
@@ -174,6 +172,10 @@ def setup_panda_for_flyscan(
     # Values need to be set before blocks are enabled, so wait here
     yield from bps.wait(group="panda-config", timeout=GENERAL_TIMEOUT)
 
+    LOGGER.info(f"PandA sequencer table has been set to: {str(table)}")
+    table_readback = yield from bps.rd(panda.seq[1].table)
+    LOGGER.info(f"PandA sequencer table readback is: {str(table_readback)}")
+
     yield from arm_panda_for_gridscan(panda)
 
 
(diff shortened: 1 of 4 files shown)


src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -174,6 +172,10 @@ def setup_panda_for_flyscan(
     # Values need to be set before blocks are enabled, so wait here
     yield from bps.wait(group="panda-config", timeout=GENERAL_TIMEOUT)
 
+    LOGGER.info(f"PandA sequencer table has been set to: {str(table)}")
+    table_readback = yield from bps.rd(panda.seq[1].table)
+    LOGGER.info(f"PandA sequencer table readback is: {str(table_readback)}")
```
DominicOram, line 177: Should: This is going to match the setpoint ideally so might be a bit noisy in the logs. If you think the bug is gone where we need to report it then remove it, otherwise can we put it as a debug?

src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -196,12 +197,16 @@ def do_fgs():
...
+        gridscan_start_time = time()
         LOGGER.info("Waiting for Zocalo device queue to have been cleared...")
         yield from bps.wait(
             ZOCALO_STAGE_GROUP
         )  # Make sure ZocaloResults queue is clear and ready to accept our new data
         LOGGER.info("completing FGS")
         yield from bps.complete(gridscan, wait=True)
+        LOGGER.info(
+            f"Gridscan motion program took {round(time()-gridscan_start_time,2)} to complete"
+        )
```
DominicOram, line 209: Should: We will hopefully have proper metrics soon so I don't think we should include printing timings like this

DominicOram (changes requested):
Thank you! Couple of comments in code

src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
olliesilvester: It's quite useful to see how long things are taking while we are still increasing the run-up distance. I could just add it in temporarily when I do testing instead?

src/hyperion/device_setup_plans/setup_panda.py
olliesilvester: I'll put as debug now. When we're 100% sure there's no more issues with it I will remove

src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
DominicOram: OK, can you leave it in with a comment saying to remove it when we have metrics?

DominicOram (approved):
Great, thank you!

PR #1352, opened 2024-05-07
 Convert UDCDirectoryProvider to standard pattern

### LINKED ISSUE #1355: Use standard pattern for UDCDirectoryProvider
Hyperion side of dodal change https://github.com/DiamondLightSource/dodal/pull/505

- Makes the concept of an updating DirectoryProvider which may work for multiple Runs more generic
- Avoids having duplicate singleton DirectoryProvider getter methods
- Allows for consistent handling with pre-processors if chosen to adopt
- Prevent code duplication

DiamondJoseph:
Fixes #1355 

Link to dodal PR (if required): #[505](https://github.com/DiamondLightSource/dodal/pull/505)

This ensures that the DirectoryProvider is a singleton that will reliably be shared between all devices if e.g. copying a detector from an existing beamline. It also exposes it to the pre-processor that is being used for configuring devices per-Run.

### To test:
1. Check that PandA would write into correct directory for parameters if it wrote to a directory

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ install_requires =
     xarray
     doct
     databroker
-    dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@6864607ad8cf5415fa63cafa0a8a7ff955471eeb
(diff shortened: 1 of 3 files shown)


olliesilvester (commented):
Looks good, although if you agree with my suggested dodal changes then this will need to be updated before merging

olliesilvester (approved):

PR #1353, opened 2024-05-07
Convert plans to use new parameter model

### LINKED ISSUE #1278: grid_width_microns is required by grid_detection_plan but is not in the experiment params schema
Currently `grid_width_microns` is required by `grid_detection_plan`. It is obtained from the `GridScanWithEdgeDetectParams.grid_width_microns` via `PinCentreThenXrayCentreInternalParameters`. However in `wait_for_robot_load_then_centre_plan` the upstream `PinCentreThenXrayCentreInternalParameters` is constructed from `WaitForRobotLoadThenCentreInternalParameters` which doesn't have this attribute, and it is also absent from the corresponding json schema.

This may possibly be mitigated by GDA populating this field anyway, but really we should not be round-tripping parameters via json between unrelated types as this is a source of potential bugs

### LINKED ISSUE #1289: Switch over the plans to use the new param types

dperl-dls:
Fixes #1289, fixes #1278

Now based on main! Finishes off using the new parameters in all plans.

### To test:
1. Run tests


src/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -153,44 +148,47 @@ def prepare_for_robot_load(composite: RobotLoadThenCentreComposite):
...
+        assert params.sample_puck is not None
+        assert params.sample_pin is not None
         yield from bps.abs_set(
             composite.robot,
             SampleLocation(
-                parameters.experiment_params.sample_puck,
-                parameters.experiment_params.sample_pin,
+                params.sample_puck,
+                params.sample_pin,
             ),
```
DominicOram, line 178: Nit: This fits on one line now

src/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -203,45 +201,53 @@ def robot_load():
...
-    wavelength_angstroms = convert_eV_to_angstrom(actual_energy_ev)
-    parameters.hyperion_params.ispyb_params.resolution = resolution(
-        parameters.hyperion_params.detector_params,
+    wavelength_angstroms = convert_eV_to_angstrom(use_energy)
+    use_resolution = resolution(
+        params.detector_params,
         wavelength_angstroms,
-        parameters.experiment_params.detector_distance,
+        params.detector_distance_mm
+        or (yield from bps.rd(composite.detector_motion.z.user_readback)),
```
DominicOram, line 228: Nit: I think this internal yield adds a little too much complexity, I would suggest we just find the `det_distance` first then call `resolution`

DominicOram (changes requested):
Thanks, a few comments but it's a bit hard to see the wood through the trees given it's based on two other PRs. Could we get those merged then rebase this?

DominicOram (changes requested):
Thanks, some comments in code.

DominicOram (approved):
Great, thank you

PR #1366, opened 2024-05-08
1276 use new params in callbacks

### LINKED ISSUE #1276: Modify the callbacks to use the new external param classes
See https://github.com/DiamondLightSource/hyperion/issues/1092 and https://github.com/DiamondLightSource/hyperion/issues/1275

## Acceptance Criteria
* The callbacks use the new parameters and have no dependency on the old models
* Tests all pass
* In doing this think about callback structure, what parameters it needs and spin out new issues if there are ways to simplify

### LINKED ISSUE #1277: Remove old internal param models
See https://github.com/DiamondLightSource/hyperion/issues/1092 and https://github.com/DiamondLightSource/hyperion/issues/1275

dperl-dls:
Fixes #1276 fixes #1277
Uses new params in callbacks and removes old params

### To test:
1. Run tests


tests/system_tests/external_interaction/conftest.py
```diff
@@ -146,14 +143,12 @@ def fetch_datacollectiongroup_attribute(sqlalchemy_sessionmaker) -> Callable:
 
 @pytest.fixture
 def dummy_params():
-    dummy_params = GridscanInternalParameters(
+    dummy_params = ThreeDGridScan(
         **raw_params_from_file(
             "tests/test_data/parameter_json_files/system_test_gridscan_param_defaults.json"
```
rtuck99, line 148: ispyb system test is broken because this file no longer exists

src/hyperion/parameters/gridscan.py
```diff
@@ -249,23 +173,21 @@ def panda_FGS_params(self) -> PandAGridScanParams:
...
-        x_end = self.x_start_um + self.x_step_size_um * self.x_steps
-        y1_end = self.y_start_um + self.y_step_size_um * self.y_steps
+        x_end = self.x_start_um + self.x_step_size_um * (self.x_steps - 1)
+        y1_end = self.y_start_um + self.y_step_size_um * (self.y_steps - 1)
         grid_1_x = Line("sam_x", self.x_start_um, x_end, self.x_steps)
         grid_1_y = Line("sam_y", self.y_start_um, y1_end, self.y_steps)
-        grid_1_omega = Static("omega", self.grid1_omega_deg)
         grid_1_z = Static("sam_z", self.z_start_um)
-        return grid_1_x.zip(grid_1_z).zip(grid_1_omega) * ~grid_1_y
+        return grid_1_y.zip(grid_1_z) * ~grid_1_x
```
rtuck99, line 181: I don't really understand why this has changed - I'm assuming there's a reason but I don't understand scanspecs enough to work out what this does

src/hyperion/external_interaction/nexus/write_nexus.py
```diff
@@ -13,68 +13,51 @@
...
+        *,
+        scan_points: AxesPoints | None = None,
         run_number: int | None = None,
+        omega_start_deg: float | None = None,
         vds_start_index: int = 0,
     ) -> None:
         self.beam: Optional[Beam] = None
         self.attenuator: Optional[Attenuator] = None
-        self.scan_points: dict = scan_points
+        self.scan_points: dict = scan_points or parameters.scan_points
```
rtuck99, line 40: Not sure I'm really a fan of sometimes-we-initialise-from-parameters, sometimes we don't, I think it makes things less clear at the call site.

src/hyperion/parameters/gridscan.py
dperl-dls: because I got it wrong the first time, haha. it changes which axis is the slow/fast dimension. it wasn't a problem before because the scanspec was recalculated in the old params.

src/hyperion/parameters/gridscan.py
```diff
@@ -287,32 +209,15 @@ def scan_points(self):
...
-            experiment_params=self.FGS_params,
-        )
+    @property
+    def scan_points_2(self):
+        """A list of all the points in the second grid scan."""
+        return ScanPath(self.grid_2_spec.calculate()).consume().midpoints
 
-    # Can be removed in #1277
-    def panda_old_parameters(self):
-        if self.y_steps % 2 and self.z_steps > 0:
```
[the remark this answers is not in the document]
dperl-dls: good catch, thanks

rtuck99 (approved):
Changes to system tests are good, much more concise. Thanks!

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_mapping.py
```diff
@@ -25,7 +28,7 @@ def populate_xz_data_collection_info(detector_params) -> DataCollectionInfo:
     return info
 
 
-def populate_xy_data_collection_info(detector_params):
+def populate_xy_data_collection_info(detector_params: DetectorParams):
```
DominicOram, line 31: Nit: Could you add a type hint on the return whilst we're here?

src/hyperion/external_interaction/callbacks/xray_centre/nexus_callback.py
```diff
@@ -48,18 +46,27 @@ def __init__(self) -> None:
...
-            nexus_data_2 = parameters.get_nexus_info(2)
-            self.nexus_writer_1 = NexusWriter(parameters, **nexus_data_1)
+            parameters = ThreeDGridScan.from_json(json_params)
+            d_size = parameters.detector_params.detector_size_constants.det_size_pixels
+            grid_n_img_1 = parameters.scan_indices[1]
+            grid_n_img_2 = parameters.num_images - grid_n_img_1
+            data_shape_1 = (grid_n_img_1, d_size.width, d_size.height)
+            data_shape_2 = (grid_n_img_2, d_size.width, d_size.height)
+            run_number_2 = parameters.detector_params.run_number + 1
+            self.nexus_writer_1 = NexusWriter(
```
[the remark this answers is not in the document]
DominicOram: As a general comment this feels less neat but I think we can clean it another time

src/hyperion/external_interaction/nexus/write_nexus.py
```diff
@@ -13,68 +13,51 @@
...
+from hyperion.parameters.gridscan import ThreeDGridScan
+from hyperion.parameters.rotation import RotationScan
 
 
 class NexusWriter:
     def __init__(
         self,
-        parameters: InternalParameters,
-        scan_points: dict,
+        parameters: ThreeDGridScan | RotationScan,
```
DominicOram, line 30: Could: Better to depend on the lowest class required but in this case is something which inherits from both `DiffractionExperimentWithSample` and `OptionalGonioAngleStarts`. So we would have to make a new type for it. I think it's work we may end up having to do but it's fine for now

DominicOram: In fact I think the better way to solve this is probably for this if for the caller to fill in `omega_start_deg` and `chi_start_deg` then take a `DiffractionExperimentWithSample` set of parameters. Could we either do that, argue about it or spin into a new ticket to argue about it. Up to you

src/hyperion/parameters/gridscan.py
DominicOram: Could: If you change this back e.g. remove the `-1` it's only picked up in a test for writing nexus files and so the error is not obvious. It would be nice if there was some more explicit testing around this code, if nothing else it would act as more documentation. Happy to spin into a new issue

src/hyperion/parameters/gridscan.py
```diff
@@ -281,33 +210,19 @@ def scan_points(self):
...
 
+    @property
+    def scan_points_1(self):
+        """A list of all the points in the first grid scan."""
+        return ScanPath(self.grid_1_spec.calculate()).consume().midpoints
+
+    @property
+    def scan_points_2(self):
+        """A list of all the points in the second grid scan."""
+        return ScanPath(self.grid_2_spec.calculate()).consume().midpoints
```
DominicOram, line 221: Could: Can you call these something more like `scan_points_first_grid`?

src/hyperion/parameters/gridscan.py
```diff
@@ -225,6 +152,10 @@ def FGS_params(self) -> GridScanParams:
 
     @property
     def panda_FGS_params(self) -> PandAGridScanParams:
+        if self.y_steps % 2 and self.z_steps > 0:
+            raise OddYStepsException(
+                "The number of Y steps must be even for a PandA gridscan"
+            )
```
DominicOram, line 158: Must: I can comment this out and all tests still pass, can we make sure it's caught?

DominicOram (changes requested):
Thank you, this is so much cleaner now! Some comments in code but mostly there.

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_mapping.py
dperl-dls: nah, and in fact I will remove it from the other function to save two lines; pyright can tell that `return DataCollectionInfo(...)` returns a `DataCollectionInfo` and we aren't using mypy here

src/hyperion/external_interaction/nexus/write_nexus.py
dperl-dls: sure, I think that's a good solution

src/hyperion/external_interaction/callbacks/xray_centre/nexus_callback.py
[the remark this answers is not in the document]
dperl-dls: it's better than it was before ;)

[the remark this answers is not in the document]
DominicOram: Oh, I agree, hence just saying it as a general comment

DominicOram (approved):
Brilliant thanks. Minor nit that I did myself, otherwise all good.

PR #1368, opened 2024-05-09
Correctly handle passed-in ISPyB experiment type (i.e. "characterization")

### LINKED ISSUE #1360: Screen and collect not working
When we do a screen and collect the screening should first put into ispyb with the experiment kind of `Characterization`, this causes agamemnon to tell GDA to wait for the screening to finish before the next collection. The experiment kind is passed into Hyperion straight from GDA. It looks as though the latest Hyperion version is not passing this into ispyb e.g. https://ispyb.diamond.ac.uk/samples/sid/5536566 should be a screen + collect.

## Acceptance Criteria
* Hyperion correctly passes experiment kind into ispyb from GDA
* Screen and collect works

dperl-dls:
Fixes #1360 
Ensures ispyb experiment type is correctly passed along from parameters

### To test:
1. Confirm updated test now checks the right thing

### DIFF
--- a/src/hyperion/parameters/components.py
+++ b/src/hyperion/parameters/components.py
@@ -235,7 +235,7 @@ class OptionalGonioAngleStarts(BaseModel):
 
 
 class TemporaryIspybExtras(BaseModel):
-    # for while we still need ISpyB params_to be removed in #1277 and/or #43
+    # for while we still need ISpyB params - to be removed in #1277 and/or #43
     class Config:
         arbitrary_types_allowed = True
         extra = Extra.forbid
--- a/src/hyperion/parameters/components.py
+++ b/src/hyperion/parameters/components.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import datetime
 from abc import abstractmethod
-from enum import Enum
+from enum import StrEnum
 from pathlib import Path
 from typing import Sequence, SupportsInt, TypeVar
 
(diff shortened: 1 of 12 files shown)


src/hyperion/parameters/components.py
```diff
@@ -45,19 +45,64 @@ def __modify_schema__(cls, field_schema):
...
 
 
-class XyzAxis(str, Enum):
+class XyzAxis(StrEnum):
     X = "sam_x"
     Y = "sam_y"
     Z = "sam_z"
 
 
+class IspybExperimentType(StrEnum):
```
DominicOram, line 61: Should: Does it make sense to just have the ones we can do here? It's a bit deceptive to have the whole list when I suspect some of them aren't even used at DLS.

src/hyperion/parameters/components.py
```diff
@@ -189,7 +235,7 @@ class OptionalGonioAngleStarts(BaseModel):
 
 
 class TemporaryIspybExtras(BaseModel):
-    # for while we still need ISpyB params - to be removed in #1277 and/or #43
+    # for while we still need ISpyB params_to be removed in #1277 and/or #43
```
DominicOram, line 238: Nit: Why did this get changed?

DominicOram (changes requested):
Awesome, thanks. Minor comment in code

src/hyperion/parameters/components.py
dperl-dls: overeager find and replace :man_facepalming:

src/hyperion/parameters/components.py
dperl-dls: I just moved this from the ISPyB modules to avoid circular dependencies, it was already like this. Anyway, I think enums should be exhaustive, it shouldn't hurt us that some are rarely if ever used.

DominicOram (approved):

PR #1369, opened 2024-05-09
Make deploy script create venv while SSH'd into control machine

### LINKED ISSUE #1103: Modify the deploy script to run on the control machine so that the correct versions of dependencies are installed
RHEL 7 is limited to GLIBC 2.17, which limits pyzmq to <25 for the default wheel. However, there is an `pyzmq-26.0.2-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl` build which is still compatible, and will be installed if the pip install is run on the control machine instead of a RHEL8(+) machine

### Acceptance criteria:
- deploy script builds the `venv` ssh'd to the control machine
- dependency pin for pyzmq is removed

olliesilvester:
Fixes #1103 

Now uses `subprocess` to SSH, so that it only asks for the password if it needs it

### To test:
1. Confirm deploy script in dev mode still works as expected, using `python deploy_hyperion.py "dev"` (NOTE: this won't actually work until we make a release with this change merged in, since the script will search the latest release for the file created in this PR)
2. We don't have unit tests for the deploy script, so test live on beamline for next deployment

### DIFF
--- a/utility_scripts/deploy/deploy_hyperion.py
+++ b/utility_scripts/deploy/deploy_hyperion.py
@@ -82,9 +82,10 @@ def get_hyperion_release_dir_from_args() -> str:
 
 def create_environment_from_control_machine():
     ssh_client = paramiko.SSHClient()
-    ssh_client.set_missing_host_key_policy(
-        paramiko.AutoAddPolicy()
-    )  # This adds i03-control to your client's known hosts.
+
+    # This adds i03-control to your client's known hosts.
+    ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+
     username = input(
         "Setting up the venv requires SSH'ing to i03-control.\n Enter FedID:"
     )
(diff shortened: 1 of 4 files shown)


utility_scripts/deploy/deploy_hyperion.py
```diff
@@ -71,11 +75,45 @@ def get_hyperion_release_dir_from_args() -> str:
...
+        return DEV_DEPLOY_LOCATION
     else:
         return f"/dls_sw/{args.beamline}/software/bluesky"
 
 
+def create_environment_from_control_machine():
+    ssh_client = paramiko.SSHClient()
+
+    # This adds i03-control to your client's known hosts.
+    ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
```
dperl-dls, line 87: is it really necessary to have username/password for this? I don't have to enter them when I `ssh i03-control` normally so I would prefer an automated solution, probably a little more secure too

dperl-dls (commented):
Script changes look good, couple comments

utility_scripts/deploy/deploy_hyperion.py
olliesilvester: I tested out just the SSH part, and for me I do need the credentials. Potentially if you set up SSH properly in your `bashrc` then this won't be needed, but then this needs to be a prerequisite for deployment

olliesilvester: @DominicOram  @rtuck99 do you have opinions on what is preferred?

olliesilvester: I think so, it might look a little more clunky, but I guess it would be better if it worked it out for us

DominicOram: Yh, lets try it. This is always going to be a bit temporary anyway

dperl-dls (approved):
awesome, thanks!

PR #1370, opened 2024-05-09
Calculate resolution in ispyb callback

### LINKED ISSUE #1359: Calculate resolution if not passed in
Currently the rotation and XRC plans rely on GDA to pass resolution in for us to put in ispyb. Hyperion should be able to calculate this itself, in fact it does so in `robot_load_then_centre`. This calculation may make sense to do in the callback?

To caclulcate the resolution we need to read various parameters on the beamline and we need to make sure to do them after they are set up for data collection.

## Acceptance Criteria
* Hyperion calculates resolution itself for XRC/rotations based on the state of the beamline during the collection
* Resolution is removed as something that can be passed into hyperion

d-perl:
Fixes #1359 

So that we always pass the resolution to ISPyB, it is now calculated in the callback
Requires GDA change to remove resolution passed in

### To test:
1. run tests


tests/test_data/new_parameter_json_files/good_test_rotation_scan_parameters_nomove.json
```diff
@@ -44,7 +44,6 @@
         "beam_size_x": 1.0,
         "beam_size_y": 1.0,
         "focal_spot_size_x": 1.0,
-        "focal_spot_size_y": 1.0,
```
olliesilvester, line 47: How come these parameters are removed on only some of the test jsons?

olliesilvester (approved):
Changes look sensible. thanks

tests/test_data/new_parameter_json_files/good_test_rotation_scan_parameters_nomove.json
dperl-dls: uh, I dunno, but they'll all get deleted soon

Issue #516, opened 2023-02-01
Remove prefix from machine PVs in S03 and make sure ports don't conflict


dperl-dls:
Some system tests (`read_hardware_for_ispyb`) have started failing because some new PVs which `fgs` connects to are not present:

```
TimeoutError: Failed to connect to all signals: fgs.synchrotron.machine_status.synchrotron_mode (CS-CS-MSTAT-01:MODE), fgs.synchrotron.machine_status.user_countdown (CS-CS-MSTAT-01:USERCOUNTDN), fgs.synchrotron.machine_status.beam_energy (CS-CS-MSTAT-01:BEAMENERGY), fgs.synchrotron.top_up.start_countdown (SR-CS-FILL-01:COUNTDOWN), fgs.synchrotron.top_up.end_countdown (SR-CS-FILL-01:ENDCOUNTDN), fgs.synchrotron.ring_current (SR-DI-DCCT-01:SIGNAL)
```

DominicOram:
So this does work for me but that's because it's looking at the real PVs. `S03` does have fake PVs but they're on `BL03S-CS-CS...`, which is hard to fit in given that there is no equivalent prefix for the real case. We will need to fix this when we do https://github.com/DiamondLightSource/python-artemis/issues/458 as then the test won't work against the real PV any more. I think the solution is to remove the additional prefix in S03

PR #1371, opened 2024-05-09
Create panda subdirectory within panda plan

### LINKED ISSUE #1339: Panda: Write positions for every gridscan
As a scientist/developer I would like to know that the panda is triggering at the right places. The panda can write triggering positions out as HDF files, we should configure it so that it always does this.

Note:
* We provided the panda with a directory for this at https://github.com/DiamondLightSource/hyperion/pull/1337 but that directory will not exist, we may need to create it

## Acceptance Criteria
* Every time a panda grid scan is run it writes the positions out to an appropriate file

olliesilvester:
Fixes #1339 

Link to dodal PR (if required): [#516](https://github.com/DiamondLightSource/dodal/pull/516)

### To test:
1. Confirm changes are sensible and tests pass


tests/unit_tests/experiment_plans/test_flyscan_xray_centre_plan.py
```diff
@@ -61,6 +61,9 @@
 from hyperion.log import ISPYB_LOGGER
 from hyperion.parameters.constants import CONST
 from hyperion.parameters.gridscan import ThreeDGridScan
+from hyperion.parameters.plan_specific.gridscan_internal_params import (
```
dperl-dls: this probably shouldn't be here anymore

tests/unit_tests/experiment_plans/test_panda_flyscan_xray_centre_plan.py
```diff
@@ -53,6 +52,9 @@
 from hyperion.log import ISPYB_LOGGER
 from hyperion.parameters.constants import CONST
 from hyperion.parameters.gridscan import ThreeDGridScan
+from hyperion.parameters.plan_specific.panda.panda_gridscan_internal_params import (
```
dperl-dls: as above

dperl-dls (approved):
looks good, thank you!

src/hyperion/experiment_plans/panda_flyscan_xray_centre_plan.py
```diff
@@ -189,9 +189,11 @@ def run_gridscan_and_move(
         time_between_x_steps_ms,
     )
 
-    get_directory_provider().update(
-        directory=Path(parameters.detector_params.directory)
-    )
+    panda_directory = Path(parameters.detector_params.directory)
+
+    # Don't create directory once implemented in ophyd-async
```
DominicOram, line 194: Should: Does this have an `ophyd_async` issue we can link to?

src/hyperion/experiment_plans/panda_flyscan_xray_centre_plan.py
```diff
@@ -189,9 +189,11 @@ def run_gridscan_and_move(
         time_between_x_steps_ms,
     )
 
-    get_directory_provider().update(
-        directory=Path(parameters.detector_params.directory)
-    )
+    panda_directory = Path(parameters.detector_params.directory)
```
DominicOram, line 192: Should: I think `parameters.storage_directory` is where we get the detector directory from originally so we should probably use that here rather than digging into the detector.

DominicOram: Should: I'm also mildly concerned that people will get detector and panda data confused. Can we put it in a `panda` subfolder for now?

DominicOram (changes requested):
Couple of comments in code. For interest, what are the names of the files that panda will write out?

dperl-dls (approved):
lgtm!

PR #1374, opened 2024-05-13
1350 nexus file cannot be read by imginfo

### LINKED ISSUE #1350: Nexus file cannot be read by `imginfo`
Nexus files being written by the latest Hyperion version cannot be parsed by `imginfo`. To reproduce the error see https://diamondlightsource.slack.com/archives/C048RD0KFFF/p1715011760143679?thread_ts=1714765475.230639&cid=C048RD0KFFF

## Acceptance Criteria
* `imginfo` passes on all nexus files written by Hyperion
* There is an automated test that confirms nexus files can be read by imginfo

rtuck99:
Fixes #1350 

Link to nexgen PR (if required): DiamondLightSource/nexgen#237

### To test:
1. Imginfo and autoPROC pipeline run successfully against generated nexus files

* The nexgen write operation has been moved even though now I believe it's not necessary since there's no update method in nexgen anymore, but I have left in rather than move it back, I am guessing that perhaps it may come in handy in future if we ever need to do it? If people would rather I restore the old ordering I can do this
* Running it as system test requires DOCKER environment variable to be set to `podman` if you want to use podman otherwise it defaults to `docker`. Same goes for the utility scripts
* System tests is marked with `s03` but not sure if we want a different flag
* System tests require you to have run `build_imginfo.sh` to build `imginfo`
* `imginfo` is pinned to the latest commit hash, do we want to track the `HEAD` - they don't appear to have version tags
* I'm assuming the reference imginfo output is correct for these files.


dperl-dls (approved):
brilliant, thanks

PR #1378, opened 2024-05-14
Nexgen 231 thoroughly test nexgen created files against gda

rtuck99:
Does part of DiamondLightSource/nexgen#231

### To test:
Unit tests should pass

The expected/exceptional vaues have NOT been verified; the values have been entered based on what the code currently does, on the expectation that it works.

Also perhaps we should split into two tests, one against GDA old data, and one against nexus standard conformance.

Another Q is, whether this test or something like it ought to be in `nexgen` rather than here - I am open to suggestions


tests/unit_tests/external_interaction/test_write_rotation_nexus.py
```diff
@@ -71,6 +75,152 @@ def plan():
     return plan()
 
 
+def dectris_device_mapping(meta_filename: str):
```
rtuck99, line 78: Perhaps when we get more sample files with other detectors we could factor this out and drive this from the test data files

dperl-dls (approved):
brilliant, thanks!

dperl-dls:
I would like to merge this to add on to it but leave the issue open until we're sure about the contents of the exceptions etc

PR #1381, opened 2024-05-15
Run imginfo on nexus files in CI

dperl-dls:
Fixes https://github.com/DiamondLightSource/nexgen/issues/231
Adds nexus file validation steps to our CI runs
Adds tests which compare nexus files in detaul

- I will spin off a new ticket to tidy messy duplicated test code if this is accepted without doing that
- we will need a new issue to determine exactly what acceptable differences from GDA are

### To test:
1. Inspect the new tests and workflow


tests/unit_tests/external_interaction/nexus/test_compare_nexus_to_gda_exhaustively.py
```diff
@@ -0,0 +1,266 @@
...
+
+    add_layer(e, dataset)
+    return e
+
+
+def has_equiv_in(item: str, groups: set, exception_table: dict[str, set[str]]):
+    if item not in exception_table.keys():
+        return False
+    # one of the items in exception_table[item] must be in the tested groups
+    return exception_table[item] & groups != set()
```
rtuck99, line 38: wouldn't it be clearer just to do `not exception_table[item] in groups` ?

tests/unit_tests/external_interaction/nexus/test_compare_nexus_to_gda_exhaustively.py
```diff
@@ -0,0 +1,266 @@
...
+    print(diffs)
+
+
+def test_hyperion_vs_gda_datatypes(
+    files_and_groups: FilesAndgroups,
+):
+    gda_nexus, gda_groups, hyperion_nexus, hyperion_groups = files_and_groups
+    for item in gda_groups:
+        # we checked separately if all expected items should be there
+        # but we should probably still check the excepted ones here??
```
rtuck99, line 239: If the gda file really is from GDA and we don't update it to a new baseline over time I think the number of things that are not in the original GDA file is going to make this test increasingly less useful, unless we explicitly check the exceptions here.

tests/unit_tests/external_interaction/nexus/test_compare_nexus_to_gda_exhaustively.py
```diff
@@ -0,0 +1,266 @@
...
+        if item in hyperion_groups:
+            hyperion_group = hyperion_nexus[item]
+            gda_group = gda_nexus[item]
+            print(hyperion_group, gda_group)
+            assert isinstance(hyperion_group, Group) and isinstance(gda_group, Group)
+            for dset_or_attr in hyperion_group:
+                hyperion_item = mockitem("hyperion", dset_or_attr)
+                gda_item = mockitem("gda", dset_or_attr)
+                try:
+                    hyperion_item = hyperion_group[dset_or_attr]
```
rtuck99, line 249: Won't this always succeed?

tests/unit_tests/external_interaction/nexus/test_compare_nexus_to_gda_exhaustively.py
dperl-dls: the equivalent would be something like `any([ex in groups for ex in exception_table[item]])`, it's testing that at least one item (but possibly more) in the set in the exception table is also in the set of groups

rtuck99 (approved):

tests/test_data/nexus_files/rotation/rotation_scan_test_nexus_0.nxs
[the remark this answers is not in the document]
dperl-dls: Ah, I fixed this, it now uses a newly generated file rather than that one

PR #1382, opened 2024-05-16
Fix tests

dperl-dls:
- for nexgen v0.9.3 change
- for robot barcode reader change

And fixes linting too

### To test:
1. Run tests


DominicOram (approved):
Great, thank you

PR #1389, opened 2024-05-17
(#1102) Remove undulator_gap from ispyb_params

### LINKED ISSUE #1102: Use energy as read during collection in ispyb
https://github.com/DiamondLightSource/hyperion/pull/1074 adds an energy read during collection but also still adds the energy to the ispyb parameters before collection. If we're reading the energy we no longer need to provide it upfront.

## Acceptance Criteria
* The energy that is read during collection gets put in ispyb
* We no longer set the energy upfront anywhere, including when called from GDA

rtuck99:
Fixes #1102

This removes the now redundant undulator_gap from the ispyb params

### To test:
1. Tests pass


DominicOram (approved):
Great, thank you!

PR #1390, opened 2024-05-17
Fix for system tests after omega_end fix

rtuck99:
Fixes the broken test_nexgen system test after the change to the omega end parameter calculation.

### To test:
1. System tests pass


dperl-dls (approved):
lgtm, thank you!

PR #1398, opened 2024-05-20
Set dev shm on Eiger based on using GPU

### LINKED ISSUE #1372: Provide GDA property to use GPU analysis
By enabling the dev/shm cache on ODIN we can now have the GPU analysis working in parallel to zocalo. This is useful for testing results coming out but we're still using the results from zocalo as before. We should allow the user to set whether the GPU code is running or not. However, if the dev/shm cache is left on the EPU disk can fill up and it is cleared every time the cache is turned off. To be safe we should be overly cautious about turning off the cache whenever we don't need it

## Acceptance Criteria
* There is a property in GDA for using the GPU
* If set when Hyperion XRC starts it toggles `BL03I-EA-EIGER-01:OD:FAN:DevShmCache` to high, this starts the GPU analysis
* If not set we toggle `BL03I-EA-EIGER-01:OD:FAN:DevShmCache` to low
* When Hyperion XRC gets a result from zocalo and is finished it toggles `BL03I-EA-EIGER-01:OD:FAN:DevShmCache` to low, this stops the GPU and clears the cache on the Eiger
* If there is an error in running a plan we toggle `BL03I-EA-EIGER-01:OD:FAN:DevShmCache` to low

DominicOram:
Fixes #1372 

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/560

### To test:
1. Confirm tests pass


olliesilvester (approved):
Looks good, I also put this change on the latest version when testing yesterday and it seemed to work

PR #1409, opened 2024-05-23
Move aperture at the start of rotation

### LINKED ISSUE #1336: Move to aperture from GDA before rotation

olliesilvester:
Fixes #1336 

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/573

Not to be merged until the WIP GDA changes are done

### To test:
1. Confirm changes satisfy issue
2. Confirm tests still pass


src/hyperion/experiment_plans/rotation_scan_plan.py
```diff
@@ -95,6 +107,13 @@ class RotationMotionProfile:
     max_velocity_deg_s: float
 
 
+def move_aperture(aperture: ApertureScatterguard, position: SingleAperturePosition):
```
dperl-dls, line 110: Nit: does this really need to be in a separate function?

dperl-dls (approved):
LGTM

src/hyperion/experiment_plans/rotation_scan_plan.py
```diff
@@ -264,6 +283,13 @@ def rotation_scan(
     def rotation_scan_plan_with_stage_and_cleanup(
         params: RotationScan,
     ):
+        assert composite.aperture_scatterguard.aperture_positions is not None
+        yield from move_aperture(
```
DominicOram, line 287: Must: I think this should probably go in or next to `setup_sample_environment`. We should be staging the detector as the first thing we do then asynchronously moving this into the correct place, we then only need to wait for it to have finished moving just before we do `rotation_scan_plan`

DominicOram (changes requested):

src/hyperion/experiment_plans/rotation_scan_plan.py
olliesilvester: Yeah good point!

src/hyperion/device_setup_plans/manipulate_sample.py
```diff
@@ -12,16 +16,24 @@
...
     backlight: Backlight,
     attenuator: Attenuator,
     transmission_fraction: float,
     detector_distance: float,
     group="setup_senv",
 ):
-    """Move out the backlight, retract the detector shutter, and set the attenuator to
-    transmission."""
+    """Move the aperture into the position determined during XRC, Move out the backlight, retract the detector shutter,
+    and set the attenuator to transmission."""
```
DominicOram, line 29: Should: It is from XRC in this case but it won't be generally, I think just `move the aperture into required position` is fine

DominicOram (changes requested):
Apologies, another couple of comments

src/hyperion/parameters/rotation.py
```diff
@@ -44,6 +45,7 @@ class RotationScan(
         default=IspybExperimentType.ROTATION
     )
     transmission_frac: float
+    selected_aperture: AperturePositionGDANames
```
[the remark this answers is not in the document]
olliesilvester: David and I talked about this a bit yesterday - `GridCommon` uses `DiffractionExperiment`. Since the grid scans determine the aperture, why should the aperture bt an input parameter?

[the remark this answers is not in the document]
olliesilvester: Actually, your way works fine too, happy with that

olliesilvester:
After discussion with @DominicOram , this PR should now allow the following:
- `selected_aperture` is an optional parameter of `DiffractionExperiment`

- XRC: If set to `None`, XRC won't change the aperture, and whatever is currently in position will be used. Otherwise, XRC will set the aperture according to the parameter

- Rotation: Same as above, and GDA should pass back the selected parameter from Hyperion's XRC

- Robot load: Inherits from `DiffractionExperiment`, but the plan always sets aperture to `RobotLoad` position, regardless of parameter. This is because, for now, robot load and XRC use the same set of parameters.

DominicOram (approved):
Great, thank you! I think I'm happy to merge w/o the GDA changes being done yet

PR #1422, opened 2024-05-29
(#1387) snapshot_directory request parameter can now be optionally specified

### LINKED ISSUE #1387: Use `storage_directory`/`snapshots` as snapshot path
From discussion with Neil, he does not care where the snapshots live. For ease we can just use the `storage_directory` + `snapshots`. We may still want to optionally take a snapshot directory in case other people care.

## Acceptance Criteria
* Snapshot directory is optional from GDA
* If not provided it defaults to `storage_directory`/`snapshots`

rtuck99:
Fixes #1387

### To test:
1. If the snapshot directory parameter is specified in the request then this path will be used to store the snapshots

### DIFF
--- a/src/hyperion/parameters/components.py
+++ b/src/hyperion/parameters/components.py
@@ -12,7 +12,7 @@ from dodal.devices.detector import (
     TriggerMode,
 )
 from numpy.typing import NDArray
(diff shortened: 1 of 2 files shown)


DominicOram (approved):
Great, thank you!

Issue #539, opened 2023-02-17
Zebra sim: add position compare


dperl-dls:
In order to test new functionally required for, and eventually the execution of, a rotation data collection plan, the zebra sim should implement a basic position compare algorithm

**Acceptance criteria:**
If `PC_GATE_START', and `PC_GATE_WID` are set, and the signal which is the input to `PC_ENC` is in between these values, `PC_OUT` should be 1, and otherwise 0

PR #1423, opened 2024-05-30
Move WarningException and some oav plans to dodal

olliesilvester:
Fixes dodal [#539](https://github.com/DiamondLightSource/dodal/issues/539)

Link to dodal PR (if required): [#XXX](https://github.com/DiamondLightSource/dodal/pull/589)

### To test:
1. Do thing x
2. Confirm thing y happens


olliesilvester:
It seems like `pre_centring_setup_oav` and `setup_pin_tip_detection_params` are a little too specific to also be moved into dodal? I might be wrong though

tests/unit_tests/experiment_plans/test_grid_detect_then_xray_centre_plan.py
```diff
@@ -30,7 +28,7 @@ def _fake_grid_detection(
     box_size_um: float = 0.0,
 ):
     oav = i03.oav(fake_with_ophyd_sim=True)
-    smargon = fake_smargon()
+    smargon = i03.smargon(fake_with_ophyd_sim=True)
```
DominicOram, line 31: Should: This isn't equivalent as `fake_smargon` patched the motors too. Can you use the `smargon()` in `conftest` here?

DominicOram (changes requested):
Great, thanks, one comment but also https://github.com/DiamondLightSource/dodal/pull/589#discussion_r1622494776. I think you're right about not moving more. Wanting to know where the beam is on an OAV image seems generically useful. Doing a centring algorithm on that seems potentially more specific, probably something more for `mx_bluesky`

src/hyperion/exceptions.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def catch_exception_and_warn(
+    exception_to_catch: Type[Exception],
+    func: Callable[..., Generator[Msg, None, T]],
+    *args,
+    **kwargs,
+) -> Generator[Msg, None, T]:
+    """A plan wrapper to catch a specific exception and instead raise a WarningException,
+    so that UDC is not halted
```
DominicOram, line 24: Love this helper function, thank you!

tests/unit_tests/test_exceptions.py
```diff
@@ -0,0 +1,23 @@
...
+from hyperion.exceptions import WarningException, catch_exception_and_warn
+
+
+class TestException(Exception):
+    pass
+
+
+def dummy_plan():
+    raise TestException
+    yield from null
```
DominicOram, line 13: Nit: If you swap the `yield from` and the raise then you can remove the `type: ignore` later on

DominicOram, line 13: Must: I think it needs to be `yield from null()`

tests/unit_tests/experiment_plans/test_grid_detect_then_xray_centre_plan.py
```diff
@@ -26,11 +28,10 @@ def _fake_grid_detection(
     snapshot_dir: str,
     grid_width_microns: float = 0,
     box_size_um: float = 0.0,
+    smargon=None,
```
DominicOram: I think https://github.com/DiamondLightSource/hyperion/pull/1423/commits/2274ac616c1bacec1df04d315ef67f4e36e94556 is slightly cleaner, feel free to disagree though

src/hyperion/exceptions.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            raise
+
+    return (
+        yield from contingency_wrapper(
+            func(*args, **kwargs),
+            except_plan=warn_if_exception_matches,
+            auto_raise=True,
+        )
+    )
```
DominicOram, line 47: Should: There's a bit of redundancy here. I think we either need the else raise and then `auto_raise=False` or no else raise and then `auto_raise=False`. Both having the else cause and the auto_raise is redundant, which is fine but maybe a little misleading

DominicOram (changes requested):
Thanks, couple of comments but basically there.

src/hyperion/exceptions.py
```diff
@@ -42,6 +42,5 @@ def warn_if_exception_matches(exception: Exception):
         yield from contingency_wrapper(
             func(*args, **kwargs),
             except_plan=warn_if_exception_matches,
-            auto_raise=True,
```
DominicOram: Nit: This defaults to `True` so you still have the redundancy that you had before. Though I'm not too bothered as this protects us against future changes in the default.

DominicOram (approved):
Great, thank you! Minor comment, take it or leave

src/hyperion/exceptions.py
olliesilvester: I will take it out. I also just realise that `warn_if_exception_matches` is a plan so it needs a `yield from null()`

PR #1424, opened 2024-05-31
Fix unit tests after aperture change

olliesilvester:
A recent change in dodal required a change in BlueAPI, which hasn't had a release since the change, so breaks one of our tests. This PR pins blueapi to that commit, as well as fix one of the tests I added in #1336 

### To test:
1. Confirm tests pass

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,8 +44,8 @@ install_requires =
     ophyd == 1.9.0
     ophyd-async >= 0.3a5
     bluesky >= 1.13.0a3
-    blueapi >= 0.4.3-a1
-    dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@4b53c861ac2babb4c96fcad55405083eb549e531
+    blueapi @ git+https://DiamondLightSource/blueapi.git@59507723e7e8abddbd5e0607f656d3c5fb3a14ac
+    dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git
 
 
 [options.entry_points]
--- a/tests/unit_tests/device_setup_plans/test_manipulate_sample.py
+++ b/tests/unit_tests/device_setup_plans/test_manipulate_sample.py
@@ -27,12 +27,10 @@ async def test_move_aperture_goes_to_correct_position(
     )
 
 
-@patch("bluesky.plan_stubs.abs_set")
 async def test_move_aperture_does_nothing_when_none_selected(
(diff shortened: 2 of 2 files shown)


setup.cfg
```diff
@@ -44,8 +44,8 @@ install_requires =
     ophyd == 1.9.0
     ophyd-async >= 0.3a5
     bluesky >= 1.13.0a3
-    blueapi >= 0.4.3-a1
-    dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@4b53c861ac2babb4c96fcad55405083eb549e531
+    blueapi @ git+https://github.com/DiamondLightSource/blueapi.git@8bb938848ab11fe2aa8601519783ef6ad605dd40
```
DominicOram, line 47: Should: Can you add a link to the PR we need released for us to unpin

tests/conftest.py
```diff
@@ -486,7 +487,7 @@ def aperture_scatterguard(RE):
         patch_async_motor(ap_sg.scatterguard.x),
         patch_async_motor(ap_sg.scatterguard.y),
     ):
-        RE(bps.abs_set(ap_sg, ap_sg.aperture_positions.SMALL))  # type:
+        RE(bps.abs_set(ap_sg, ap_sg.aperture_positions.SMALL))  # type: ignore
```
DominicOram, line 490: Nit:  A better fix would be to add `assert ap_sg.aperture_positions is not None` before this

tests/unit_tests/device_setup_plans/test_manipulate_sample.py
```diff
@@ -27,12 +27,10 @@ async def test_move_aperture_goes_to_correct_position(
...
 async def test_move_aperture_does_nothing_when_none_selected(
-    mock_set: MagicMock, aperture_scatterguard: ApertureScatterguard, RE: RunEngine
+    aperture_scatterguard: ApertureScatterguard, RE: RunEngine
 ):
     assert aperture_scatterguard.aperture_positions
-    await aperture_scatterguard.set(aperture_scatterguard.aperture_positions.ROBOT_LOAD)
-
-    RE(move_aperture_if_required(aperture_scatterguard, None))
-    mock_set.assert_not_called()
+    with patch("bluesky.plan_stubs.abs_set") as mock_set:
```
DominicOram: Thank you! I think you can go slightly further with this. I pushed a commit that does so, feel free to revert if you disagree

DominicOram (approved):
Couple of small comments, happy for you to merge if you address them.

PR #1425, opened 2024-06-03
Rotation scan puts in correct comment to ispyb

### LINKED ISSUE #933: Rotation scan ispyb comment should contain metadata
Currently the comment in ispyb for a rotation scan is fixed:
image
We should add the same metadata that's in the comment for GDA:
image

That is:
* Motor position
* Some user defined comment - which comes in through UDC
* Xray centre box selected
* Aperture

For now, don't do the XRC boxes as this is a little trickier. Ask Neil if this is important.

olliesilvester:
Fixes #933 

### To test:
1. Confirm tests are sensible and pass


DominicOram (changes requested):
Thank you, some comments in code

DominicOram (approved):
Awesome, thanks! Great!

PR #1431, opened 2024-06-06
Pass phi on to nexgen

### LINKED ISSUE #1410: Phi is not included in the nexus output
Currently in the imginfo output when run on the generated nexus file, the phi angle is always reported as 0

We should include this in the nexus output

DominicOram:
Fixes #1410 

### To test:
1. Confirm tests still pass


src/hyperion/external_interaction/callbacks/rotation/nexus_callback.py
```diff
@@ -95,5 +95,8 @@ def activity_gated_start(self, doc: RunStart):
                 shape,
                 parameters.scan_points,
                 omega_start_deg=parameters.omega_start_deg,
-                chi_start_deg=parameters.chi_start_deg or 0,
+                chi_phi_start_deg=(
```
dperl-dls, line 98: Could: I would rather either these all be separate params, or better yet an `OptionalGonioAngleStarts` so that the defaults only have to be defined in one place

dperl-dls (approved):
Looks good, one optional comment

dperl-dls (approved):
great, thanks

PR #1434, opened 2024-06-10
Update to work with ophyd_async xspress3

DominicOram:
Fixes https://github.com/DiamondLightSource/dodal/issues/473

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/524

### To test:
1. Confirm tests still pass


src/hyperion/experiment_plans/optimise_attenuation_plan.py
```diff
@@ -154,7 +149,8 @@ def open_and_run():
         yield from bps.abs_set(
             composite.sample_shutter, ZebraShutterState.OPEN, wait=True
         )
-        yield from bps.abs_set(composite.xspress3mini.do_arm, 1, wait=True)
+        yield from bps.stage(composite.xspress3mini, wait=True)
+        yield from bps.unstage(composite.xspress3mini, wait=True)
```
Relm-Arrowny, line 153: Do we want to unstage here? The original version never unstage, e.g. it never try to stop xspress3mini.  Unstage here will hit the stop button and stop the data collection immediately rather than wait for it to finish naturally.

olliesilvester: I think it's OK to unstage here, as it looks like we're waiting on all the same values are before, but it would be better to create a finalise plan which includes closing the shutter and disarming the detector, then putting that in `@bpp.finalize_decorator(close_shutter)`. Then it will disarm on an error

olliesilvester: I'm not sure if we should disarm the detector after every iteration, or just once at the end of all the iterations

olliesilvester (approved):
These changes look good, but I'm not 100% convinced on what the disarming behaviour should be. I can double check what the old GDA code used to do for this?

Issue #349, opened 2022-11-11
Take diagnostic snapshots between gridscan and rotation


DominicOram:
Between gridscans and data collections we should take 4 snapshots at 0, 90, 180, 270 omega. The path to these is then added to the ispyb entry for the data collection. (see https://ispyb.diamond.ac.uk/dc/visit/cm31105-4/id/9153427).

## Acceptance Criteria
* There is a plan that takes 4 snapshots at the correct locations
* The plan can be performed synchronously or asynchronously

## Notes
* There may be an optimisation where we take the snapshots backwards so we end at 0 for the next scan

DominicOram:
It looks like we're currently only putting 3 snapshots into Hyperion, rather than 4, we should fix that as part of this https://ispyb.diamond.ac.uk/dc/visit/cm37235-2/id/13648572

DominicOram:
Before doing this we will need to do https://github.com/DiamondLightSource/dodal/issues/471. @npat113 is happy with us just doing the crosshairs for now, see https://diamondlightsource.slack.com/archives/C048RD0KFFF/p1714062911126409

DominicOram:
Need to do https://github.com/DiamondLightSource/hyperion/issues/1336 before this

dperl-dls:
Should this be inlined in the rotartion scan plan, or should it be a separate generic plan?

DominicOram:
Yes, let's do a separate plan that we can parameterize cleanly. In which case Large.

DominicOram:
I believe the only part left of this that we don't have code for is:
> There is a way to switch between doing this and doing it in GDA as we currently do
Which is probably easiest done whilst testing on the beamline

PR #1437, opened 2024-06-11
349 Create OAV snapshot plan

rtuck99:
Part 1 of fix for #349

This creates an experiment plan for taking oav snapshots and also removes some unused parameters from the `ispyb_extras` parameters. `ispyb_extras` is now no longer needed for grid scans. OAV snapshots for rotations are still taken by GDA and snapshot paths supplied in parameters; this new plan is not yet invoked by the rotation plan, that work will be in a subsequent PR.

### To test:
1. Tests pass


src/hyperion/experiment_plans/oav_snapshot_plan.py
```diff
@@ -0,0 +1,61 @@
...
+    yield from bps.create(DocDescriptorNames.OAV_ROTATION_SNAPSHOT_TRIGGERED)
+    yield from bps.read(composite.oav.snapshot)
+    yield from bps.save()
+
+
+def oav_snapshot_plan(
+    composite: OavSnapshotComposite,
+    parameters: WithSnapshot,
+    oav_parameters: OAVParameters,
+    wait: bool = True,
```
rtuck99, line 57: This wait parameter is here as requested in original issue, but currently unused since we have to wait for the snapshot trigger completion before firing the event. Am open to suggestions to remove it / alternate structuring to make it useful.

docs/param_hierarchy-Hyperion_Parameter_Model.png
dperl-dls, line 1: nice!

src/hyperion/experiment_plans/oav_snapshot_plan.py
dperl-dls: As it is we can remove this, I agree. If we need to do something more complex async we'll need to do it at device level anyway.

PR #1438, opened 2024-06-11
Update DEV_ISPYB_DATABASE_CFG to point to ispyb-restore now that ispyb-dev is defunct

rtuck99:
Changes the credentials file to use ispyb-restore rather than ispyb-dev so that system tests continue to work

### To test:
1. System tests pass

### DIFF
--- a/src/hyperion/parameters/constants.py
+++ b/src/hyperion/parameters/constants.py
@@ -14,7 +14,7 @@ class SimConstants:
     # this one is for unit tests
     ISPYB_CONFIG = "tests/test_data/test_config.cfg"
     # this one is for system tests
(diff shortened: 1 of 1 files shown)


DominicOram (changes requested):
I don't think this will work as the dev version also contains information for `exp-eye` that we need. See discussion at https://diamondlightsource.slack.com/archives/CKEPJ8E01/p1718187048335559?thread_ts=1718033249.962799&cid=CKEPJ8E01

DominicOram:
Karl has now updated the old files to point to the new database, we have some other issues (https://github.com/DiamondLightSource/hyperion/issues/1441) but I think they're best fixed in a different PR. Suggest we close this?

PR #1440, opened 2024-06-11
Thaw for a set time straight after robot load

### LINKED ISSUE #1418: Turn thawing on after robot load
After robot load we thaw the pin to try and avoid ice forming. We want thawing to be on for a user set amount of time, carry on with the rest of the plan then have it turn off in the background.

Dodal control of the thawing was done in https://github.com/DiamondLightSource/dodal/issues/538

## Acceptance Criteria
* Discuss with Neil how long this should be on for etc and write this here
* Do the thawing control as part of a plan

DominicOram:
Fixes #1418

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/609

### To test:
1. Confirm that this would kick off the thawer after robot load


src/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -182,6 +184,8 @@ def robot_load():
 
         yield from bps.wait("robot_load")
 
+        yield from bps.abs_set(composite.thawer.thaw_for_time_s, params.thawing_time)
```
olliesilvester, line 187: Is there some point in this plan where we need to check that the thawer is off before continuing?

DominicOram: I don't think so. I will double check with Neil

DominicOram: In fact @neil-i03 do we need to wait at some point for the thawing to have stopped or we just fire and forget? It's ok to be thawing whilst collecting

npat113: fire and forget. We can collect while thawing (and may have to).

olliesilvester (approved):
Looks good, could optionally add a test to check the new `abs_set` and for the `pin_centre_then_xray_centre_params`, but they're simple enough functions I think

DominicOram:
> Looks good, could optionally add a test to check the new abs_set and for the pin_centre_then_xray_centre_params

You're correct, I was being lazy and hoping people wouldn't pick it up, I will do this now

DominicOram:
Tests are failing due to https://github.com/DiamondLightSource/hyperion/pull/1448

olliesilvester (approved):

DominicOram:
Hmm... This is failing because https://github.com/DiamondLightSource/dodal/pull/609#discussion_r1651190757 breaks the assumption that we have here that all `Movable` things have a name

olliesilvester:
I think the 

> Hmm... This is failing because [DiamondLightSource/dodal#609 (comment)](https://github.com/DiamondLightSource/dodal/pull/609#discussion_r1651190757) breaks the assumption that we have here that all `Movable` things have a name

See https://github.com/DiamondLightSource/dodal/pull/609#discussion_r1652376288

PR #1444, opened 2024-06-13
Get feature flags from config service

### LINKED ISSUE #1397: Make a properties/configuration service
which we can query for:
- "should we use GPU or traditional XRC results"?
- "should we set the stub offsets"
- beamline parameters (will remain file-based for now)
- etc.?

dperl-dls:
Fixes #1397 

Does a best effort attempt at getting feature flags (`set_stub_offsets`) from the new config service

### To test:
1. Have a look at the config service repo and play with the example
2. Run the new system tests
3. See that nothing else is broken


src/hyperion/external_interaction/config_server.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def config_server() -> ConfigServer:
+    global _CONFIG_SERVER
+    if _CONFIG_SERVER is None:
+        _CONFIG_SERVER = ConfigServer(CONST.CONFIG_SERVER_URL, LOGGER)
+    return _CONFIG_SERVER
+
+
+def best_effort_get_feature_flag(flag_name: str, fallback: T = None) -> bool | T:
```
rtuck99, line 20: why does this method exist? get rid of it

src/hyperion/external_interaction/config_server.py
```diff
@@ -0,0 +1,37 @@
...
+
+class FeatureFlags(BaseModel):
+    # The default value will be used as the fallback when doing a best-effort fetch
+    # from the service
+    use_panda_for_gridscan: bool = False
+    use_gpu_for_gridscan: bool = False
+    set_stub_offsets: bool = False
+
+    @classmethod
+    def best_effort(cls):
```
rtuck99, line 32: I think we should have a bulk get REST call rather than get the feature flags one by one, if we are always going to be getting them all in one go anyway.

rtuck99, line 32: This method gets invoked every time we want only one of the feature flags, but it gets all of the feature flags. Why don't we just get all of the feature flags right at the start and put them all in the parameters in one go?

I realise that while it is a rest call this is just going to likely be quite fast, but at the end of the day we are in the long run ultimately trying to get end-to-end times of < 10s and say a few hundred ms extra if you end up with a lot of calls baked in to the code due to bad API design choices early on, it isn't completely insignificant.

rtuck99 (changes requested):
See comments with code

rtuck99 (approved):
Lovely

Issue #1391, opened 2024-05-18
Rotation start omega uses the wrong sign


DominicOram:
When GDA gives a start angle for rotation e.g. 90 deg the rotation actually starts at -90 from GDA's perspective. This is likely because the co-ordinate system in GDA is inverse to that in EPICS e.g. 90 in GDA is -90 in EPICS.

## Acceptance Criteria
* From GDA's perspective a rotation starting at +90 deg starts at +90 deg

DominicOram:
Look at fixing this in `dodal`

DominicOram:
If the user says start at 90 it should start at 90 in EPICS

PR #1445, opened 2024-06-14
(#1327) ispyb omega_start, axis_start now read from hardware for gridscans

### LINKED ISSUE #1327: ispyb omega_start value should be read from hardware instead of supplied by params
from PR review comment by @DominicOram in #1279
"Should: We should be able to read the omega from the smargon when we do the pin tip detection, can we make an issue for that?"

### LINKED ISSUE #1454: Improve structure of snapshots in ispyb callbacks
Currently in `BaseISPyBCallback` we have `_handle_oav_rotation_snapshot_triggered` and `_handle_oav_grid_snapshot_triggered` but `BaseISPyBCallback` is subclassed by `RotationISPyBCallback` and `GridscanISPyBCallback`. The functions that are specific to the different usecases should live in the corresponding classes.

rtuck99:
Fixes #1327 and fixes #1454

Gridscans should now read populate the `omega_start` and `axis_start` values in ispyb by reading the smargon omega directly during grid detection when snapshots are taken.
The corresponding changes for rotations will be made in a separate PR ~as this depends on #1443~
Note that the values will be inverted from previous as they are taken from hardware; this change should be temporary until #1391 is fixed

### To test:
1. Tests pass
2. ispyb reports correct `omega_start`/`axis_start`/`axis_range`/`axis_end` values

### DIFF
--- a/src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
+++ b/src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
@@ -60,7 +60,7 @@ class RotationISPyBCallback(BaseISPyBCallback):
     def activity_gated_start(self, doc: RunStart):
         if doc.get("subplan_name") == CONST.PLAN.ROTATION_OUTER:
             ISPYB_LOGGER.info(
-                "ISPyB callback recieved start document with experiment parameters."
+                "ISPyB callback received start document with experiment parameters."
             )
             self.params = RotationScan.from_json(doc.get("hyperion_parameters"))
             dcgid = (
@@ -104,11 +104,6 @@ class RotationISPyBCallback(BaseISPyBCallback):
             self.uid_to_finalize_on = doc.get("uid")
         return super().activity_gated_start(doc)
 
-    def populate_axis_info_for_snapshot(
-        self, data_collection_info: DataCollectionInfo, omega_start: float | None
-    ):
-        pass
-
     def populate_info_for_update(
         self,
         event_sourced_data_collection_info: DataCollectionInfo,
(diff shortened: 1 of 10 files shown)


src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
```diff
@@ -104,6 +104,12 @@ def activity_gated_start(self, doc: RunStart):
             self.uid_to_finalize_on = doc.get("uid")
         return super().activity_gated_start(doc)
 
+    def populate_axis_info_for_snapshot(
+        self, data_collection_info: DataCollectionInfo, omega_start: float | None
+    ):
+        # TODO in subsequent PR which depends on rotation snapshots 349
+        pass
+
```
DominicOram, line 112: Should: I don't think we should do this at all for the rotation we can probably always just use the start angle given in the parameters. The snapshots for XRC and rotation have subtlety different purposes for the user:
* For XRC the snapshots are used work out what grids to do, therefore they will always be taken at the same angles as the datacollection
* For rotation they are just diagnostics to prove to the user we have centered on the crystal, they could be taken at arbitrary rotations that don't match the rotation we would be doing during data collection. So I think it'll potentially cause us issues in the future to tie the two together.

tests/system_tests/external_interaction/test_ispyb_dev_connection.py
```diff
@@ -662,10 +662,10 @@ def test_ispyb_deposition_in_gridscan(
     assert position_id is None
     DC_EXPECTED_VALUES.update(
         {
-            "axisstart": 90.0,
-            "axisend": 90.0,
+            "axisstart": -90.0,
+            "axisend": -90.0,
             "datacollectionnumber": 2,
-            "omegastart": 90.0,
+            "omegastart": -90.0,
```
DominicOram, line 668: Must: Sorry, I know we discussed this but upon reflection I think we should make sure that we get the sign right in this PR (maybe just invert it in the callback?) then we can remove in the follow up issue. Otherwise if we release this before getting round to the next issue we break analysis

DominicOram (changes requested):
Great, thank you, some comments in code.

src/hyperion/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -163,11 +162,20 @@ def activity_gated_event(self, doc: Event):
...
         return doc
 
+    def populate_axis_info_for_snapshot(
+        self, data_collection_info: DataCollectionInfo, omega_start: float | None
+    ):
+        if omega_start is not None:
+            data_collection_info.omega_start = omega_start
+            data_collection_info.axis_start = omega_start
+            data_collection_info.axis_end = omega_start
+            data_collection_info.axis_range = 0
```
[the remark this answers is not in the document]
rtuck99: It's got the name because it is an implementation of an abstract method which is called from the base class `_handle_oav_snapshot_triggered` method - so the method definitely is snapshot-specific.

src/hyperion/external_interaction/callbacks/rotation/ispyb_callback.py
rtuck99: This was going to be implemented for rotations before we had our previous discussion, however the PR for that never got raised as we decided not to do it. I will have removed the comment and leave it empty.

I was originally going to in the second PR for rotations, create a second snapshot event type that was rotation specific for this very reason, and then move the snapshot handlers into the callback subclasses and rename the event types, since they are plan-specific.

If you would like me to do that, I can move those changes across from the branch where I have saved them (minus the rotation snapshot event).

DominicOram (approved):
This looks good, thank you!

PR #1456, opened 2024-06-19
fix snapshot test for thumbnail

dperl-dls:
Adding thumbnail snapshots broke a test which checks the number of snapshots saved


rtuck99 (approved):
Approved!

PR #1467, opened 2024-06-28
Cache lower gonio positions before robot load and restore them afterwards

### LINKED ISSUE #1439: Save and restore lower gonio poisitions before/after robot load
The smargon is mounted on a set of lower gonio stages `BL03I-MO-GONP-01:X` etc. The robot is taught to load/unload on to the smargon when these stages are at 0. However, throughout the run this gonio is recalibrated to make sure pins are still correctly aligned with the OAV when they are mounted. 

# Acceptance criteria

For every robot load we need to:
* Cache where these lower gonio positions are
* Move to 0 so that the robot can load to the known position
* Do the robot load (move of gonio must have finished)
* Move back to the cached positions so that the sample is still well aligned (this must be finished by the time we start on pin tip centring but doesn't need to be finished for robot load snapshots)

DominicOram:
Fixes #1439 

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/653

### To test:
1. Confirm unit tests pass


src/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -146,6 +149,36 @@ def prepare_for_robot_load(composite: RobotLoadThenCentreComposite):
...
+    demand_energy_ev: float | None,
+):
+    yield from bps.abs_set(
+        composite.robot,
+        sample_location,
+        group="robot_load",
+    )
+
+    if demand_energy_ev:
+        yield from set_energy_plan(
```
dperl-dls, line 164: bit tangential, but does the set energy plan always have to block during robot load or could we wait on it later?

src/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -166,23 +199,32 @@ def robot_load_then_centre_plan(
...
 
-        if params.demand_energy_ev:
-            yield from set_energy_plan(
-                params.demand_energy_ev / 1000,
-                cast(SetEnergyComposite, composite),
-            )
-
-        yield from bps.wait("robot_load")
+        # The lower gonio must be in the correct position for the robot load and we
+        # want to put it back afterwards. Note we don't wait as we assume the lower
```
dperl-dls, line 214: would be good to add explanation of why this is safe here

src/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -146,6 +149,36 @@ def prepare_for_robot_load(composite: RobotLoadThenCentreComposite):
...
+    if demand_energy_ev:
+        yield from set_energy_plan(
+            demand_energy_ev / 1000,
+            cast(SetEnergyComposite, composite),
+        )
+
+    yield from bps.wait("robot_load")
+
+
+def raise_exception_if_moved_out_of_cryojet(exception):
```
dperl-dls, line 172: this is nice

dperl-dls (approved):
Looks nice, one optional comment

src/hyperion/experiment_plans/robot_load_then_centre_plan.py
DominicOram: No, we want https://github.com/DiamondLightSource/hyperion/issues/1495 really but needs the RE changes

PR #1472, opened 2024-07-03
(#1466) Wait for aperture move before DC and read it during DC

### LINKED ISSUE #880: Move or remove snapshot_plan
The snapshot_plan is in an odd place and doesn't really look like other plans we should either move or remove it (given that it's not actually used explicitly)

### LINKED ISSUE #1466: Wait on aperture_scatterguard move and read during collection
Following https://github.com/DiamondLightSource/hyperion/issues/1465 we're not waiting for the `aperture_scatterguard` to settle before trying to read it for ispyb. Given this is quite a slow move it would be beneficial to wait on it as late as possible. As such, the proposed solution is:

* Change the default group on the aperture move to `ready_for_data_collection`. This will make sure we wait on the movement as late as possible for the collection
* Put the reading of the aperture into `read_hardware_for_ispyb_during_collection` rather than `read_hardware_for_ispyb_pre_collection` so that we can wait for it later
* Move `read_hardware_for_ispyb_during_collection` to actually happen during collection

We probably want to do a similar thing for the grids and the rotation

## Acceptance Criteria
* We make sure to wait on the aperture move before putting in ispyb and before collection
* This happens as late as possible in the plan

dperl-dls:
Fixes #1466 and fixes #880

### To test:
1. Run tests


DominicOram (changes requested):
Couple of small changes, will address myself

DominicOram (approved):

PR #1481, opened 2024-07-11
Remove override for faking devices

### LINKED ISSUE #1480: Do not override `fake_with_ophyd_sim` in `setup_context`
There are some times we want to set a device to be simulated on a beamline and we would like `setup_context` to respect this. E.g. when a specific device is known to not be working. We should make this as easy as possible. Currently just setting `fake_with_ophyd_sim = True` on the device in `i03.py` does not work as `setup_context` then overrides it.

## Acceptance Criteria
* Default values for `fake_with_ophyd_sim` in a beamline module are respected in production

DominicOram:
Fixes #1480

### To test:
1. Confirm tests still pass

### DIFF
--- a/src/hyperion/utils/context.py
+++ b/src/hyperion/utils/context.py
@@ -70,16 +70,13 @@ def device_composite_from_context(context: BlueskyContext, dc: Type[DT]) -> DT:
     return dc(**devices)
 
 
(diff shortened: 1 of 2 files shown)


rtuck99 (approved):
Approved

PR #1484, opened 2024-07-15
Combine nexus and ispyb hardware readings and improve naming

### LINKED ISSUE #1482: Nexus callback: issue with event ordering
Currently the nexus callbacks assume that `read_hardware_for_ispyb_during_collection` is done before `read_hardware_for_nexus_writer`. Since https://github.com/DiamondLightSource/hyperion/pull/1472 this is no longer true and so the nexus writer fails with `beam` not being defined.

We should fix this by either:
* Combining the two reads into one call
* Making the nexus writer ambivalent to the order that events come in on

## Acceptance Criteria
* The nexus callback can write nexus files during plans

dperl-dls:
Fixes #1482 

- Combines "read hardware for nexus" into the reading during the plan
- Removes "ispyb" from plan stub names, to make them just "read pre/during collection"

### To test:
1. Run tests

### DIFF
--- a/src/hyperion/device_setup_plans/read_hardware_for_setup.py
+++ b/src/hyperion/device_setup_plans/read_hardware_for_setup.py
@@ -16,16 +16,16 @@ from hyperion.log import LOGGER
 from hyperion.parameters.constants import CONST
 
 
-def read_hardware_for_ispyb_pre_collection(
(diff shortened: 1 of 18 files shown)


rtuck99 (approved):
Approved, all looks good

PR #1498, opened 2024-07-24
1474 panda gridscan smargon speed

### LINKED ISSUE #1474: PandA grid scans: Safeguard against varying smargon speed
Our current panda setup is meant to do the following:

1. At the start of a row, wait for a GPIO signal from the motion controller
2. After encoder counts go above a position, enable a PULSE block which sends perioidic triggers (period = deadtime+exposure time)
3. After encoder counts goes above another postiion, disable this PULSE block so no more triggers are sent
4. Other stuff not relevant to this issue

This assumes that the PandA is going at exactly the right speed by the time we get to bullet point 2. If we are going too quickly or too slowly, the number of triggers received may be off by a tiny amount. The last time we tested the panda, we saw an occasional missing trigger. We concluded that this was either down to what has just been described, or due to lost encoder counts due to accelerating the smargon too quickly.

We should instead do the following:

- When Hyperion sets up the PandA, calculate how many triggers are expected and send it over to the panda
- The PandA should wait for the GPIO signal, then wait for encoder counts to go above an initial value, and then periodically send out triggers until the expected number of triggers have been sent. This can all be done on the sequencer table, we don't need to use the CLOCK or PULSE blocks
- Other stuff not relevant to this issue

This will rule out missing triggers from the smargon going at the slight wrong speed. If the speed is wrong, we will still get the triggers, but they will be in a weird place (and this will be obvious from data collection). When PVT scans are working, it will setup the panda in a similar way, but it will be nice to do it here first so we can compare the times of panda scans using the motion controller and using the ophyd-async flyscan

## Acceptance Criteria
Do the above, and update documentation https://github.com/DiamondLightSource/hyperion/wiki/PandA-constant%E2%80%90motion-scanning

rtuck99:
Fixes #1474 

Corresponding save-panda cli utility:
Link to dodal PR (if required): DiamondLightSource/dodal#702

### To test:
1. Panda gridscans can be captured
2. tests pass


src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -32,30 +35,37 @@ class PcapArm(Enum):
...
-        3:Wait for POSA (X2) to be greater than X_START + X_STEP_SIZE + a safe distance for the final trigger, then cut out the signal
-        4:Wait for physical trigger from motion script to mark change of direction
-        5:Wait for POSA (X2) to be less than X_START + X_STEP_SIZE + exposure distance, then
-            send a signal out every (minimum eiger exposure time + eiger dead time)
-        6:Wait for POSA (X2) to be less than (X_START - safe distance + exposure distance), then cut out signal
-        7:Go back to step one.
+
+        1. Wait for physical trigger from motion script to mark start of scan / change of direction
+        2. Wait for POSA (X2) to be greater than X_START and send 1 trigger
+        3. Send out the remaining x_steps - 1 triggers every time_between_steps_ms
```
olliesilvester, line 50: Can we combine these two steps? Eg the second row could look like
repeats = num_pulses, trigger = POSA>POSITION, TIME1=exposure_time, TIME2=`time_between_x_steps`, outa1 = 1, outa2 = 0

olliesilvester: Same for 5 and 6

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -67,44 +77,65 @@ def get_seq_table(
 
     exposure_distance_x_counts = int(exposure_distance_mm * MM_TO_ENCODER_COUNTS)
 
+    num_pulses = parameters.x_steps
+
+    delay_between_pulses = time_between_steps_ms * TICKS_PER_MS
+
+    PULSE_WIDTH_US = 1
```
olliesilvester, line 84: I believe we don't actually need a pulse block. The SEQ.OUT.A can be treated as our pulse, with width determined by `time1` and period determined by `time2`

olliesilvester (changes requested):
Looks mostly good, but the main comments:

- I don't think we need a separate pulse block. I also think we can combine the 'wait for position and send one pulse' and 'send all other pulses' steps.

- I know this was me originally, but I think we should try making the deadtime equal to what it says in the eiger manual, 100 ns

- I haven't looked at the yaml file in massive detail, but we should make sure the position part of this sequencer table is ignored now, like in `/dls_sw/i03/software/daq_configuration/panda_configs/flyscan_pcap_ignore_seq.yaml`. This was previously causing a race condition in the panda ioc, which has apparently been fixed now, but still better to be safe.

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -32,30 +35,37 @@ class PcapArm(Enum):
...
+        1. Wait for physical trigger from motion script to mark start of scan / change of direction
+        2. Wait for POSA (X2) to be greater than X_START and send 1 trigger
+        3. Send out the remaining x_steps - 1 triggers every time_between_steps_ms
+        4. Wait for physical trigger from motion script to mark change of direction
+        5. Wait for POSA (X2) to be less than X_START + X_STEP_SIZE * x_steps + exposure distance, then
+            send 1 trigger
+        6. Send the remaining x_steps - 1 triggers every time_between_steps_ms
+        7. Go back to step one.
 
         For a more detailed explanation and a diagram, see https://github.com/DiamondLightSource/hyperion/wiki/PandA-constant%E2%80%90motion-scanning
```
olliesilvester, line 57: We need to update this too now :(

src/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -32,30 +35,37 @@ class PcapArm(Enum):
...
 ) -> SeqTable:
     """
-    -Exposure distance is the distance travelled by the sample each time the detector is exposed: exposure time * sample velocity
-    -Setting a 'signal' means trigger PCAP internally and send signal to Eiger via physical panda output
-    -When we wait for the position to be greater/lower, give a safe distance (X_STEP_SIZE/2 * MM_TO_ENCODER counts) to ensure the final trigger point
-    is captured
+    Generate the sequencer table for the panda.
+
+    - Sending a 'trigger' means trigger PCAP internally and send signal to Eiger via physical panda output
+
```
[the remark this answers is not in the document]
rtuck99: see args: section in docstring

rtuck99:
Have simplified the sequence table but not removed the pulse generator or reduced the deadtime as discussed

olliesilvester (changes requested):
I think this is in a testable state for tomorrow! I have a few more comments, but they're mostly nits

olliesilvester (approved):
Good to merge now I think, after that missing docstring is added

PR #1502, opened 2024-07-30
(#842) Fix ispyb activation so that messages can be logged from pin tip detection

### LINKED ISSUE #824: Signify on synchweb pin too long/short due to pin detect
After https://github.com/DiamondLightSource/python-artemis/issues/499 we will not be doing a grid scan at all if the pin is too/short long and so not registering a grid collection in ispyb. Previously when the pin was too long/short we signified this to the user by putting the grid collection in ispyb with an error state and a comment saying the error. Given that we're not registering a grid at all now we need a different way of informing the user.

## Acceptance Criteria
* On pin too long/short there is a clear indication to the user in synchweb

rtuck99:
Fixes #824

### To test:
1. ispyb comments should now appear when the pin tip is too short or too long
2. Tests pass


olliesilvester (approved):
Looks good!

PR #1512, opened 2024-08-06
changes for dodal 692

dperl-dls:
changes to match https://github.com/DiamondLightSource/dodal/pull/726


DominicOram (approved):
Thanks! Happy with these changes but a couple of comments on the `dodal` side

PR #1513, opened 2024-08-08
Unpin dodal before release

rtuck99:
Unpins the dodal dependency in setup.cfg so that it pulls from main rather than a PR commit hash


DominicOram (approved):
LGTM

PR #1524, opened 2024-08-13
use daq config service from pypi

### LINKED ISSUE #1523: Use daq-config-service from pypi
now that it has a pypi release we should use it

dperl-dls:
Fixes #1523

### To test:
1. Run tests


DominicOram (approved):
Great, thank you!

PR #1529, opened 2024-08-13
Stage eiger asynchronously in rotation plan

### LINKED ISSUE #206: Do not wait for eiger to stage before snapshots

### LINKED ISSUE #1522: Do not wait for eiger to stage before snapshots
When we call `rotation_with_cleanup_and_stage` we wrap it in the eiger stage decorator. However, the snapshots are being done inside this plan. This means we're waiting for eiger to be staged before we do the snapshots, which is not needed. Really we need to trigger the eiger stage early then wait for it as late as possible. We may not be able to do this with `stage_decorator` unless we add in the ability to give it a group.

## Acceptance Criteria
* Eiger arming starts as early as possible in the rotation scan
* Waiting on the eiger stage happens just before we actually need to do it for the collection

DominicOram:
Fixes DiamondLightSource/mx-bluesky#206

Makes use of the `start_preparing_data_collection_then_do_plan` plan stub we have for setting up the collection that we use in the XRC. `begin_sample_environment_setup` will ultimately be fully removed by https://github.com/DiamondLightSource/mx-bluesky/issues/205

### To test:
1. Confirm tests pass
2. Manually confirm `grid_detect_then_xray_centre_composite` still passes

### DIFF
--- a/src/hyperion/device_setup_plans/manipulate_sample.py
+++ b/src/hyperion/device_setup_plans/manipulate_sample.py
@@ -16,15 +16,11 @@ LOWER_DETECTOR_SHUTTER_AFTER_SCAN = True
 
 
 def begin_sample_environment_setup(
-    detector_motion: DetectorMotion,
     attenuator: Attenuator,
     transmission_fraction: float,
-    detector_distance: float,
     group="setup_senv",
 ):
     """Start all sample environment changes that can be initiated before OAV snapshots are taken"""
-    yield from bps.abs_set(detector_motion.shutter, 1, group=group)
-    yield from bps.abs_set(detector_motion.z, detector_distance, group=group)
     yield from bps.abs_set(attenuator, transmission_fraction, group=group)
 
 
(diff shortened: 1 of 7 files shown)


olliesilvester:
I will make a PR for this branch which fixes the merge conflict, as well as makes similar changes for the `multi_rotation_scan`

DominicOram:
Replaced by https://github.com/DiamondLightSource/mx-bluesky/pull/190

PR #1533, opened 2024-08-14
Removed some references to ophyd

DominicOram:
Remove a few references to ophyd. Most of what's left should be those needed for devices that are still ophyd or `NullStatus`, which will need to think of a clean replacement for

### To test:
1. Confirm tests still pass


dperl-dls (approved):
nice! Maybe a different issue but reading this made me notice that I think we should be tagging the statuses from fixtures with the test name from the `request`

edit: oh, that won't be possible anymore with `AsyncStatus`, maybe we should add that

PR #1537, opened 2024-08-19
Move the basic parameter model components to MXB

dperl-dls:
For https://github.com/DiamondLightSource/mx-bluesky/issues/162

Moves the parameted model base components to mx-bluesky, re-adds the default values here in hyperion

Link to MXB PR (if required): #[163](https://github.com/DiamondLightSource/mx-bluesky/pull/163)

### To test:
1. Run tests

### DIFF
--- a/.vscode/hyperion-mx-bluesky.code-workspace
+++ b/.vscode/hyperion-mx-bluesky.code-workspace
@@ -1,0 +1,23 @@
+{
+	"folders": [
+		{
+			"path": ".."
+		},
+		{
+			"path": "../../dodal"
+		},
+		{
+			"path": "../../mx-bluesky"
+		}
+	],
+	"settings": {
+		"python.languageServer": "Pylance",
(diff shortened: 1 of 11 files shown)


DominicOram (approved):
Thanks, I think this is good as a first pass for getting things generic but I think there's some extra things we should think about (feel free to make new issues or solve here):
* I dislike that we need to have a subclass to basically just fill in the defaults. I feel like you should be able to do something like the way we fill in the devices using the BlueAPI context where we also fill in these constants from the context (which picks them out the appropriate `dodal.beamline`)
* The deploy script currently gives us a `dodal` deployment in the expected place. Should we now do the same for `mx_bluesky`?

dperl-dls:
- Grabbing defaults: https://github.com/DiamondLightSource/mx-bluesky/issues/166
- have added making an mx-bluesky release to the hyperion release instructions
- https://github.com/DiamondLightSource/hyperion/issues/1540 adding to deploy script

PR #1541, opened 2024-08-20
Fix to work with updated aperture scatterguard

DominicOram:
Fixes https://github.com/DiamondLightSource/dodal/issues/740

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/755

### To test:
1. Confirm tests still pass


src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -376,13 +367,10 @@ def set_aperture_for_bbox_size(
...
-    assert aperture_device.aperture_positions is not None
-
-    new_selected_aperture: SingleAperturePosition = (
-        aperture_device.aperture_positions.MEDIUM
-        if bbox_size[0] < 2
-        else aperture_device.aperture_positions.LARGE
+    new_selected_aperture = (
+        AperturePosition.MEDIUM if bbox_size[0] < 2 else AperturePosition.LARGE
     )
+    gda_name = aperture_device._loaded_positions[new_selected_aperture].GDA_name
```
dperl-dls, line 373: should: why make this non-public if you still want to access it here? imo better either leave it public, or provide a method to get the gda name

dperl-dls (changes requested):
Mostly looks good, just one thing

src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
DominicOram: Yh, you're right

dperl-dls (approved):

PR #1544, opened 2024-08-20
Update hyperion to move just the aperture_scatterguard axes required

DominicOram:
Fixes https://github.com/DiamondLightSource/hyperion/issues/1266

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/757

### To test:
1. Do thing x
2. Confirm thing y happens

### DIFF
--- a/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
+++ b/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
@@ -12,6 +12,7 @@ import numpy as np
 from attr import dataclass
 from blueapi.core import BlueskyContext, MsgGenerator
 from dodal.devices.aperturescatterguard import (
+    ApertureInOut,
     AperturePosition,
     ApertureScatterguard,
 )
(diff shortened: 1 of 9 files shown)


src/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -130,7 +130,7 @@ def take_robot_snapshots(oav: OAV, webcam: Webcam, directory: Path):
 def prepare_for_robot_load(composite: RobotLoadThenCentreComposite):
     yield from bps.abs_set(
         composite.aperture_scatterguard,
-        AperturePosition.ROBOT_LOAD,
+        (ApertureInOut.OUT, None),
         group="prepare_robot_load",
     )
```
DominicOram, line 135: We should do a set to `(OUT, SMALL)` small after this but will need to think about when to wait for it

src/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -383,7 +384,9 @@ def set_aperture_for_bbox_size(
         }
     )
     def set_aperture():
-        yield from bps.abs_set(aperture_device, new_selected_aperture)
+        yield from bps.abs_set(
+            aperture_device, (ApertureInOut.IN, new_selected_aperture)
```
DominicOram, line 388: Maybe should be out as we're going to do OAV after this

PR #1546, opened 2024-08-22
Rotation scan moves aperture to LARGE by default

### LINKED ISSUE #203: Aperture not moving in for rotation
During testing today we found that an aperture was not being selected for the rotation scan and the aperture/scatterguard was staying in `ROBOT_LOAD`. 

This may have been because we had no beam and so the XRC was not returning a sensible aperture to go to. Even so we should probably pick some default aperture (which we may have to do in the GDA side). 

## Acceptance Criteria
* We move to a aperture/scatterguard position before rotation

### LINKED ISSUE #1534: Aperture not moving in for rotation
During testing today we found that an aperture was not being selected for the rotation scan and the aperture/scatterguard was staying in `ROBOT_LOAD`. 

This may have been because we had no beam and so the XRC was not returning a sensible aperture to go to. Even so we should probably pick some default aperture (which we may have to do in the GDA side). 

## Acceptance Criteria
* We move to a aperture/scatterguard position before rotation

olliesilvester:
Fixes DiamondLightSource/mx-bluesky#203 

### To test:
1. Check changes and new tests

### DIFF
--- a/src/hyperion/device_setup_plans/manipulate_sample.py
+++ b/src/hyperion/device_setup_plans/manipulate_sample.py
@@ -54,7 +54,7 @@ def move_aperture_if_required(
             aperture_position_gda_name
         )
     else:
-        aperture_position = AperturePosition.LARGE
+        aperture_position = DEFAULT_APERTURE_POSITION
         LOGGER.info(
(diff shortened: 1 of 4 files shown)


olliesilvester:
Moved to https://github.com/DiamondLightSource/mx-bluesky/pull/429

